PR #5, opened 2025-01-14
Build system

Narsil:


pyproject.toml
```diff
@@ -1,17 +1,26 @@
...
 tomli = { version = "^2.0.1", python = "<3.11" }
 
+[tool.kernels.dependencies]
+"kernels-community/activation" = "^0.0.1"
+
 
 [build-system]
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+requires = ["torch", "huggingface_hub", "numpy"]
```
Narsil, line 20: That's quite a bit, but this is how it was written so far.

src/hf_kernels/build.py
```diff
@@ -0,0 +1,143 @@
...
+    # Unlike `find_uv_bin`, this mechanism must work according to PEP 517
+    import os
+    import tomllib
+
+    cwd = os.getcwd()
+    filename = os.path.join(cwd, "pyproject.toml")
+    with open(filename, "rb") as f:
+        data = tomllib.load(f)
+
+    for kernel, _ in (
```
danieldk, line 50: Don't we want to get revisions from here (and then pass them to `install_kernel`)?

danieldk: Getting the locked versions is not supported at the client-side yet, so let's do this later.

PR #6, opened 2025-01-14
CI build and test multiple cuda and torch versions

drbh:
This PR adds a configurable Dockerfile and a small github actions that builds various combinations of cuda and torch versions.

building locally
```bash
docker build \
--platform linux/amd64 \
--build-arg CUDA_VERSION=12.1.0 \
--build-arg TORCH_VERSION=2.4.0 \
-t kernel-ci \
-f docker/Dockerfile .
```

run example
```bash
docker run --gpus all -it --rm -e HF_TOKEN=$HF_TOKEN kernel-ci 
# Activation kernel fetched
# Input tensor created
# Kernel successfully executed
# Calculated values are exact
```

## TODO

> [!NOTE] 
> This PR has been opened early in order to debug the workflow

- [x] ensure CI is building images as expected
- [x] run small test suite after building

### DIFF
--- a/.github/workflows/docker-build-matrix.yml
+++ b/.github/workflows/docker-build-matrix.yml
@@ -60,11 +60,12 @@ jobs:
             CUDA_VERSION=${{ matrix.cuda }}
             TORCH_VERSION=${{ matrix.torch }}
           push: false
-          load: true
+          load: false # should push instead of load
           tags: kernels:${{ matrix.cuda }}-${{ matrix.torch }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
-      - name: Run Tests
-        run: |
-          docker run --gpus all kernels:${{ matrix.cuda }}-${{ matrix.torch }}
+      # TODO: re enable after pushing the images to a registry
+      # - name: Run Tests
+      #   run: |
+      #     docker run --gpus all kernels:${{ matrix.cuda }}-${{ matrix.torch }}


docker/Dockerfile
```diff
@@ -0,0 +1,76 @@
+# syntax=docker/dockerfile:1.4
+ARG CUDA_VERSION=12.4.0
+ARG UBUNTU_VERSION=22.04
```
danieldk: We should probably do the oldest possible to check backwards compat. It seems 18.04 images are not updated for newer CUDA versions, so 20.04?

Or maybe 20.04 by default and use 18.04 for CUDA 11.8 (passing through the Ubuntu version in the GitHub action).

.github/workflows/docker-build-matrix.yml
```diff
@@ -0,0 +1,56 @@
...
+  build:
+    name: Build and Test Docker Image
+    runs-on:
+      group: aws-g6-12xlarge-plus
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        cuda: ['12.1.0', '12.2.0', '12.4.0', '12.6.0']
```
danieldk: We also support 11.8 for the time being, so we should probably also test it.

docker/Dockerfile
drbh: good points! I've updated the Dockerfile to default to `20.04` and am now passing cuda version `11.8` and ubuntu version `18.04` in the build matrix in the latest commit.

PR #8, opened 2025-01-17
PoC: allow users to lock the kernel revisions

danieldk:
This change allows Python projects that use kernels to lock the kernel revisions on a project-basis. For this to work, the user only has to include `hf-kernels` as a build dependency. During the build, a lock file is written to the package's pkg-info. During runtime we can read it out and use the corresponding revision. When the kernel is not locked, the revision that is provided as an argument is used.

### DIFF
--- a/src/hf_kernels/lockfile.py
+++ b/src/hf_kernels/lockfile.py
@@ -1,0 +1,28 @@
+from typing import Any, List
+
+
+def _get_nested_attr(d, attr: List[str]) -> Any:
+    for a in attr:
+        d = d.get(a)
+        if d is None:
+            break
+    return d
+
+
+def write_egg_lockfile(cmd, basename, filename):
+    import json
+    import os
+    import tomllib
+
+    cwd = os.getcwd()
+    pyproject = os.path.join(cwd, "pyproject.toml")
+    with open(pyproject, "rb") as f:
+        data = tomllib.load(f)
+
+    kernel_versions = _get_nested_attr(data, ["tool", "kernels", "dependencies"])
+    if kernel_versions is None:
+        return
+
+    kernel_versions_json = json.dumps(kernel_versions, indent=2)
+
+    cmd.write_or_delete_file(basename, filename, kernel_versions_json)
--- a/src/hf_kernels/utils.py
+++ b/src/hf_kernels/utils.py
@@ -1,4 +1,10 @@
+from types import ModuleType
+from typing import List, Optional
 import importlib
+import importlib.metadata
+from importlib.metadata import Distribution
+import inspect
+import json
 import platform
 import sys
 import os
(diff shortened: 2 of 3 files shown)


src/hf_kernels/lockfile.py
```diff
@@ -0,0 +1,28 @@
...
+    import json
+    import os
+    import tomllib
+
+    cwd = os.getcwd()
+    pyproject = os.path.join(cwd, "pyproject.toml")
+    with open(pyproject, "rb") as f:
+        data = tomllib.load(f)
+
+    kernel_versions = _get_nested_attr(data, ["tool", "kernels", "dependencies"])
```
Narsil, line 22: NIT: I don't like those helper functions (they are super easy to screw up).
```python
data.get("tool", {}).get("kernels", {}).get("dependencies", None)
```
Is more readable to me than delegating to another function.

src/hf_kernels/utils.py
```diff
@@ -50,6 +56,10 @@ def get_kernel(repo_id: str, revision: str = "main"):
 
 
 def load_kernel(repo_id: str, revision: str = "main"):
+    locked_revision = _get_caller_locked_kernel(repo_id)
+    if locked_revision is not None:
+        revision = locked_revision
```
Narsil, line 61: This clashes with the argument.

IMHO:
We should have EITHER revision be used, but never both, meaning:
if we can have a proper lockfile, then we should ALWAYS use the lockfile.
If there are side effects that may get no lockfile, we should ALWAYS ask for the version.

Making where the information comes from is much easier to debug afterwards imho.

src/hf_kernels/utils.py
```diff
@@ -50,6 +56,10 @@ def get_kernel(repo_id: str, revision: str = "main"):
...
+
     filename = hf_hub_download(
         repo_id, "build.toml", local_files_only=True, revision=revision
```
Narsil, line 64: Maybe we should just use `"main"` all the time since we're not going to use `revision` as a versionning system.

src/hf_kernels/utils.py
```diff
@@ -59,3 +69,36 @@ def load_kernel(repo_id: str, revision: str = "main"):
...
+    # Look up all possible distributions that this module could be from.
+    package = module.__name__.split(".")[0]
+    dist_names = importlib.metadata.packages_distributions().get(package)
+    if dist_names is None:
+        return []
+
+    return [importlib.metadata.distribution(dist_name) for dist_name in dist_names]
+
+
+def _get_caller_module() -> Optional[ModuleType]:
```
Narsil, line 96: Pretty tricky. 

I can see some potential screw ups in the resoluion there (like dependency injection, callback feeding etc..).
But in general I think it's good enough to get started.

Worst case we simply inspect the entire callstack, we're bound to find our caller somewhere.

src/hf_kernels/utils.py
danieldk: Yeah, agreed. How should be go about this? Something like `revision=None` means lock file, otherwise get the specified revision? (I only Python had sum types 😁 )

Narsil: I suggest we just support one.

I'm fine looking for the lockfile always if it works (and if people don´t  use setuptools we crash first, figure it out later)

PR #10, opened 2025-01-20
Add support for locking kernels

danieldk:
* `hf-kernels lock .` locks the kernels specified in a project's `pyproject.toml`. Building a package with the `setuptools` build backend, which is the default for `pyproject.toml`, will add the lock file to the package's metadata. The a kernel can then be downloaded and loaded at the locked version using `get_locked_kernel`.
* `hf-kernels download .` downloads the locked kernels to the HF cache directory. The kernel can then be loaded using `load_kernel` (which is a small wrapper for `get_locked_kernel`).

PR uploaded as `hf-kernels` 0.1.1 to PyPI for testing.

In the future, we want to be able to specify versions in `pyproject.toml` (which then get locked), but that's for a later PR.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -75,7 +75,7 @@ COPY examples ./examples
 COPY tests ./tests
 
 # Install the kernel library
-RUN uv pip install .
+RUN uv pip install ./hf_kernels
 
 # Run tests and benchmarks
 CMD [".venv/bin/pytest", "tests", "-v"] 


src/hf_kernels/cli.py
```diff
@@ -0,0 +1,79 @@
+import argparse
+import dataclasses
+import json
+import sys
+from pathlib import Path
+
+if sys.version_info >= (3, 11):
```
Narsil: We should perhaps put this in it's own file since everywhere we import tomllib we need the logic. (Same in the deps).

src/hf_kernels/utils.py
```diff
@@ -49,13 +61,56 @@ def get_kernel(repo_id: str, revision: str = "main"):
...
+
+
+def get_locked_kernel(repo_id: str, local_files_only: bool = False):
+    """Get a kernel using a lock file."""
+    locked_sha = _get_caller_locked_kernel(repo_id)
+
+    if locked_sha is None:
+        raise ValueError(f"Kernel `{repo_id}` is not locked")
+
+    package_name, package_path = install_kernel(
```
Narsil, line 92: `load_kernel` is supposed to ignore entirely the download. This reintroduces it.

`hf_hub_downlod(..., local_files_only=True)` is the only public API I found to IGNORE downloading.

src/hf_kernels/cli.py
danieldk: Fixed.

src/hf_kernels/utils.py
danieldk: If I don't have the model cached and use `load_kernel`, I get the following error:

```
 % python -c 'import kernel_test; kernel_test.run()'
[...]
FileNotFoundError: [Errno 2] No such file or directory: '/scratch/daniel/.cache/huggingface/hub/models--kernels-community--activation/snapshots/a71853ecbdd899526f9810cc558ee24081a6302e/build/torch25-cxx98-cu124-x8
6_64-linux/activation/__init__.py'
```

The error could be better, but it doesn't seem to download?

I did forget to pass through `local_files_only` to `get_metadata`. Pushing a fix for that now. Then it fails even earlier:

```
 % python -c 'import kernel_test; kernel_test.run()'
[...]
huggingface_hub.errors.LocalEntryNotFoundError: Cannot find the requested files in the disk cache and outgoing traffic has been disabled. To enable hf.co look-ups and downloads online, set 'local_files_only' to False.
```

Narsil: What I mean is let's keep the actual code `local_files_only` separate. Otherwise it's super easy to screw up and reintroduce the internet connection. (Better yet if we could sidestep that bad API altogether).

src/hf_kernels/lockfile.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+    if r.siblings is None:
+        raise ValueError(
+            f"Cannot get sibling information for {repo_id} at revision {revision}"
+        )
+
+    file_locks = []
+    for sibling in r.siblings:
+        if sibling.rfilename.startswith("build/torch"):
```
Narsil, line 42: Shouldn't we filter by version too here ? (Maybe subsequent PR ?)

src/hf_kernels/lockfile.py
```diff
@@ -0,0 +1,77 @@
...
+    cwd = Path.cwd()
+    with open(cwd / "pyproject.toml", "rb") as f:
+        data = tomllib.load(f)
+
+    kernel_versions = data.get("tool", {}).get("kernels", {}).get("dependencies", None)
+    if kernel_versions is None:
+        return
+
+    lock_path = cwd / "hf-kernels.lock"
+    if not lock_path.exists():
```
Narsil, line 65: Isn't that always true ?

How can the lockfile exists before it's created ?

danieldk: I think it could happen in an editable install:

- Create the lockfile.
- Do an editable install. (the lockfile gets written into the package info)
- Remove the lockfile.
- Do an editable install.

Though I still need to check whether this is the case.

danieldk: Forgot to add, this checks the existence of the lock file in the project's source directory, which should exist prior to running the install if you want to lock the versions.

src/hf_kernels/utils.py
danieldk: Ok, made separate again.

src/hf_kernels/lockfile.py
danieldk: The lockfile should contain all build variants, because we don't know what Torch/CUDA version a downstream user will have.

Or did you mean kernel version? If that, one particular commit is only supposed to have one version.

Narsil: Yes I meant kernel version/version range.

PR #11, opened 2025-01-21
Accept version specifications for kernels in `pyproject.toml`

danieldk:
We lock the highest compatible version. Until we have a better mechanism, tags of the form `vX.Y.Z` are recognized as versions. The versions are locked by their git commit SHA.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,6 +36,9 @@ ARG PYTHON_VERSION
 
 RUN echo "Building with CUDA_VERSION=${CUDA_VERSION}, TORCH_VERSION=${TORCH_VERSION}, PYTHON_VERSION=${PYTHON_VERSION}"
 
+# Install requested Python version
+RUN uv python install "${PYTHON_VERSION}"
+
 # Initialize uv and create virtual env
 RUN uv init --app kernel-test --python "${PYTHON_VERSION}"
 


src/hf_kernels/lockfile.py
```diff
@@ -25,16 +27,49 @@ def from_json(cls, o: Dict):
...
+
+def get_kernel_locks(repo_id: str, version_spec: str):
+    """
+    Get the locks for a kernel with the given version spec.
+
+    The version specifier can be any valid Python version specifier:
+    https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers
+    """
+    versions = _get_available_versions(repo_id)
+    requirement = SpecifierSet(version_spec)
```
Narsil, line 52: This seems to be accepting a lot more that pure version specifiers.

https://packaging.python.org/en/latest/specifications/version-specifiers/#direct-references

But seems to convenient to pass on for now  :).
Do you happen to know if it also supports the "extensions" (not sure about the proper terminology) `"0.1.0; python_version> '3.9'"`

danieldk: Seems like they are called markers, but they are only parsable when using `Requirement`, but that also includes the package name, etc.:

https://packaging.pypa.io/en/latest/requirements.html

I think we could easily support this in the future if we want to (by stubbing the package name and then parsing everything as a requirement).

Narsil: I was thinking reducing what we accept for now (just about controlling the actual surface of what people will try to do and the errors they receive).

But again no need to reinvent the wheel here.

PR #20, opened 2025-02-05
Make module names unique by path

danieldk:
So far we have been using the extension name in `build.toml` as the module name. However, this can cause naming conflicts. For instance, if a kernel named `moe` is loaded through `hf_kernels`, it would be registered as the `moe` module. This would cause subsequent imports of a module named `moe` from the Python path to be resolved as the module loaded through `hf_kernels`.

Solve this issue by adding some unique material to the module name (hex-encoded hash of the kernel path).

### DIFF
--- a/src/hf_kernels/utils.py
+++ b/src/hf_kernels/utils.py
@@ -1,3 +1,4 @@
+import ctypes
 import importlib
 import importlib.metadata
 import inspect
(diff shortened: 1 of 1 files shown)


Narsil (approved):
Not sure if we need ctypes for this, but GTG

PR #29, opened 2025-02-24
Use per-build variant hashes in the lockfile

danieldk:
This makes the lock file a fair bit shorter than per-file hashes. The hash is computed from filenames + SHA-1 hash for git objects/SHA-256 hash for LFS files.

In contrast to #26, this approach does not require pre-downloading all build variants, since we can retrieve the hashes from the Hub.

### DIFF
--- a/src/hf_kernels/lockfile.py
+++ b/src/hf_kernels/lockfile.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
+import hashlib
 from pathlib import Path
-from typing import Dict, List
+from typing import Dict
 
 from huggingface_hub import HfApi
 from packaging.specifiers import SpecifierSet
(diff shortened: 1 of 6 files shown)


src/hf_kernels/lockfile.py
```diff
@@ -72,17 +75,36 @@ def get_kernel_locks(repo_id: str, version_spec: str):
...
+            files.append((filename.as_posix().encode("utf-8"), hash))
+
+    variant_locks = {}
+    for variant, files in variant_files.items():
+        m = hashlib.sha256()
+        for filename, hash in sorted(files):
+            # Filename as bytes.
+            m.update(filename)
+            # Git blob or LFS file hash as bytes.
+            m.update(bytes.fromhex(hash))
```
LysandreJik, line 103: very clean way to have the whole directory contribute to the sha

LysandreJik:
cc @Wauplin, not directly related to your work in `hfh` but you might be interested in how hashes/sha256 are used here.

PR #32, opened 2025-02-25
hf-kernels -> kernels

LysandreJik:
Renames the `hf-kernels` library to `kernels`.


LysandreJik:
Validated that it works correctly with the `examples/basic.py`.

Once merged I'll release on the regular pypi index.

PR #36, opened 2025-03-06
Add a bunch of cleanups

danieldk:
* Remove old, unused build backend.
* Add types to functions and check them with mypy.
* Use `Path` where possible.
* Remove unused `get_metadata` function.


.github/workflows/test.yml
```diff
@@ -47,5 +47,8 @@ jobs:
       - name: Install setuptools for Triton-based test
         run: uv pip install setuptools
 
+      - name: Check typing
+        run: uv run mypy src/kernels
```
LysandreJik, line 51: Could be cool to add `ruff` to format stuff eventually

LysandreJik (approved):
Clean, thanks for the cleanup!

.github/workflows/test.yml
danieldk: 100%, let's do it!

PR #45, opened 2025-03-11
Add `use_hub_kernel` decorator

danieldk:
This decorator replaces a layer with a layer from the Kernel Hub.

### DIFF
--- a/docs/kernel-requirements.md
+++ b/docs/kernel-requirements.md
@@ -76,6 +76,26 @@ might use two different commits that happen to have the same version
 number. Git tags are not stable, so they do not provide a good way
 of guaranteeing uniqueness of the namespace.
 
+## Layers
+
+Kernels can provide a `layers` attribute containing a Python module with
+layers. Such layers can be used directly by downstream users or through
+the `use_hub_kernel` decorator. For Torch, layers must be a subclass of
+[`nn.Module`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html).
+
+To accommodate portable loading, `layers` must be defined in the main
+`__init__.py` file. For example:
+
+```python
+from . import layers
+
+__all__ = [
+  # ...
+  "layers"
+  # ...
+]
+```
+
 ## Python requirements
 
 - Python code must be compatible with Python 3.9 and later.
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ the Hub.
 
 ## 📚 Documentation
 
+- [Using layers](docs/layers.md)
 - [Locking kernel versions](docs/locking.md)
 - [Using kernels in a Docker container](docs/docker.md)
 - [Kernel requirements](docs/kernel-requirements.md)
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1,0 +1,39 @@
+# Layers
+
+A kernel can provide layers in addition to kernel functions. For Torch
+kernels, layers are subclasses of [`torch.nn.Module`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html).
+
+## Getting a layer
+
+Layers are exposed through the `layers` attribute of a kernel,
+so they can be used directly after loading a kernel with `get_kernel`.
+For example:
+
+```python
+import kernels
+import torch
+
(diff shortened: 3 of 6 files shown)


drbh (approved):
nice dev UX improvement!

docs/kernel-requirements.md
```diff
@@ -76,6 +76,26 @@ might use two different commits that happen to have the same version
...
+```python
+from . import layers
+
+__all__ = [
+  # ...
+  "layers"
+  # ...
+]
+```
+
```
LysandreJik, line 98: Does the kernel builder automatically add layers in the build such as here? https://huggingface.co/kernels-community/activation/blob/layers/build/torch25-cxx11-cu118-x86_64-linux/activation/layers.py

LysandreJik: I would maybe put an example of a `layers.py` file (such as this one https://huggingface.co/kernels-community/activation/blob/layers/build/torch25-cxx11-cu118-x86_64-linux/activation/layers.py) in the docs to show how the `layers` module maps to method calls

LysandreJik (commented):
Looks good, I really like the API! I think the `@use_hub_kernel` is especially powerful. The API itself looks clean; I have a few questions and a few examples of where it failed in unexpected ways that I mention below.

Just for my understanding: is the goal with providing such a `layers` attribute to enable bundling multiple kernels in a single kernel repo, or is it to offer access to sub-parts of a given kernel through the different layers that make it?

Down the road it could be nice to enable better intellisense by having the returned type of `get_kernel`/ `load_kernel` return a type that has a `layers` attribute. We could eventually retrieve the layers themselves if the kernel is already downloaded, but that's probably seeing a bit too far for now.

---

Some issues I ran into:

```py
from kernels import use_hub_kernel
import torch
from torch import nn

@use_hub_kernel(
    "kernels-community/activation",
    layer_name="SiluAndMul",
    revision="layers",
)
class SiluAndMul(nn.Module):
        d = input.shape[-1] // 2
        return nn.functional.silu(input[..., :d]) * input[..., d:]


tensor = torch.Tensor([1, 2, 3])
layer = SiluAndMul()

result = layer(tensor)
```

This doesn't work right now, even though I think it should: it fails as it tries to run through the Hub kernel even though I didn't place my tensor on a device that is supported by the kernel (in this case, the CPU).

I have a GPU-enabled machine so the kernel could be loaded without issue; but I think it should be possible to still run on CPU. I guess a fix here could be to keep a registry of supported devices and a fallback to the original definition in case the incoming tensor is not satisfied by the supported devices.

--- 

Playing around with layer definition, a potential issue I'm foreseeing is the signature of the kernel being different to the method signature. For example, the obviously incorrect following snippet does compile/load correctly; but it does not run.

```py
from kernels import use_hub_kernel
import torch
from torch import nn

@use_hub_kernel(
    "kernels-community/activation",
    layer_name="SiluAndMul",
    revision="layers",
)
class MatMul(nn.Module):
    def forward(self, input1: torch.Tensor, input2: torch.Tensor) -> torch.Tensor:
        return torch.matmul(input1, input2)


tensor = torch.Tensor([1, 2, 3]).to('cuda')
layer = MatMul()

result = layer(tensor, tensor)
print(result) 
```

In this situation, the layer call was correct. The remote kernel does not fit the local method signature however. I think the correct behavior here would be to identify the Hub kernel as being incompatible with the signature here, and defaulting to the non-kernel implementation rather than trying to load the kernel and failing at runtime.

The fix here might be a bit more involved: it would require checking the input/output signatures and ensuring that they match.

---

As another, final example that could be fixed with the above, the `SiluAndMul` method behaves slightly differently in the kernel implementation vs local implementation: running with 
```py
tensor = torch.Tensor([1, 2, 3])
```
produces outputs of different shapes.

Using the kernel:
```
tensor([1.4621], device='cuda:0')
```

Without using the kernel:
```
tensor([1.4621, 2.1932], device='cuda:0')
```

This seems like a hard issue to identify because in most cases it seems to stay true to the expected shape, but maybe having tests on the input/output signatures would help identify these potential issues ahead of time.

---

Thanks a lot for your work! This is super cool and definitely going in the right direction.

danieldk:
Alternative PR: #46.

PR #46, opened 2025-03-13
Add `use_kernel_forward_from_hub` decorator

danieldk:
This decorator replaces a layer's `forward` with the `forward` of a layer on the hub.

---

This change combines #45 and https://github.com/huggingface/transformers/pull/36680:

* Like the transformers PR and unlike #45, we only replace `forward` functions. We only know upon the `forward` call what the device is, so we can only dispatch it to the correct layer at that point.
* Like the transformers PR, the decorator does not directly assign a layer from the hub, but a symbolic name. Hub layers can then be registered by this name through `register_kernel_mapping`.
* `register_kernel_mapping` is changed following @LysandreJik 's suggestion in https://github.com/huggingface/transformers/pull/36680.
* Compared to https://github.com/huggingface/transformers/pull/36680, we validate that the 'donor' layers:
  * They must be pure (no constructor, no member variables, no additional methods beyond `forward`).
  * The `forward` method must have the same number of arguments and the arguments must have the same kind. We can loosen up this requirement a little in the future.

### DIFF
--- a/docs/kernel-requirements.md
+++ b/docs/kernel-requirements.md
@@ -76,6 +76,26 @@ might use two different commits that happen to have the same version
 number. Git tags are not stable, so they do not provide a good way
 of guaranteeing uniqueness of the namespace.
 
+## Layers
+
+Kernels can provide a `layers` attribute containing a Python module with
+layers. Such layers can be used directly by downstream users or through
+the `use_hub_kernel` decorator. For Torch, layers must be a subclass of
+[`nn.Module`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html).
+
+To accommodate portable loading, `layers` must be defined in the main
+`__init__.py` file. For example:
+
+```python
+from . import layers
+
+__all__ = [
+  # ...
+  "layers"
+  # ...
+]
+```
+
 ## Python requirements
 
 - Python code must be compatible with Python 3.9 and later.
--- a/docs/kernel-requirements.md
+++ b/docs/kernel-requirements.md
@@ -78,10 +78,64 @@ of guaranteeing uniqueness of the namespace.
 
 ## Layers
 
-Kernels can provide a `layers` attribute containing a Python module with
-layers. Such layers can be used directly by downstream users or through
-the `use_hub_kernel` decorator. For Torch, layers must be a subclass of
-[`nn.Module`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html).
+A kernel can provide layers in addition to kernel functions. A layer from
+the Hub can replace the `forward` method of an existing layer for a certain
+device type. This makes it possible to provide more performant kernels for
+existing layers. See the [layers documentation](layers.md) for more information
+on how to use layers.
+
+### Writing layers
+
+To make the extension of layers safe, the layers must fulfill the following
+requirements:
+
+- The layers are subclasses of `torch.nn.Module`.
+- The layers are pure, meaning that they do not have their own state. This
+  means that:
+  - The layer must not define its own constructor.
+  - The layer must not use class variables.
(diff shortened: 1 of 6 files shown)


docs/layers.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Registering a hub kernel for a layer
+
+Once a layer is made extensible, users can register hub kernels for it
+by name using the `register_kernel_mapping` function. For example:
+
+```python
+kernel_layer_mapping = {
+    "SiluAndMul": {
+        Device(type="cuda"): LayerRepository(
```
ArthurZucker:
```suggestion
        "cuda": LayerRepository(
```
IMO showing that it can be simple if you device is simple is better (especially for cuda), but you could add an exemple of when it it relevant to use the `Device` class directly?

ArthurZucker: (eg compute capabilities)

src/kernels/layer.py
```diff
@@ -0,0 +1,212 @@
...
+
+    return ContextManager()
+
+
+def register_kernel_mapping(mapping: Dict[str, Dict[Device, LayerRepository]]):
+    """
+    Register a layer mapping.
+
+    This function registers a mapping from a layer identifier and device type
+    to a layer in a kernel repository.
```
ArthurZucker:
```suggestion
    """
    Allows one to register a mapping between a layer name the corresponding kernel to use, depending on the device. 
    This should be use in conjunction with `use_kernel_hub_forward` decorator on the classname. 
    Exemple usage:
    
    ```python
    from kernels import LayerRepository, register_kernel_mapping

    kernel_layer_mapping = {
      "LlamaRMSNorm": {
          "cuda": LayerRepository(
              layer_name="RmsNorm",
              revision="layers",
          ),
      },
    }
    register_kernel_mapping(kernel_layer_mapping)
    ```
```

docs/layers.md
```diff
@@ -0,0 +1,74 @@
...
+
+### External layers
+
+An existing layer that does not (yet) have the `use_kernel_forward_from_hub`
+decorator can be made extensible by by monkeypatching it using the `replace_kernel_forward_from_hub` function.
+
+```python
+from somelibrary import SiluAndMul
+
+replace_kernel_forward_from_hub(SiluAndMul, "SiluAndMul")
```
ArthurZucker, line 37: very nice! I would also add the mapping somewhere to "reming" that both are required

ArthurZucker (commented):
Like this a lot!

docs/layers.md
danieldk: Added

```
register_kernel_mapping(kernel_layer_mapping)
```

to the example and added a forward reference to the section about registering kernel mappings.

docs/layers.md
danieldk: I don't think we have any good examples yet, since we don't support capabilities yet. I changed the registration function to take `Union[Device|str]`. Maybe we should only use `str` in examples until we support capabilities or other `Device` features.

docs/kernel-requirements.md
```diff
@@ -76,6 +76,80 @@ might use two different commits that happen to have the same version
...
+            hidden_states,
+            self.weight,
+            bias=None,
+            residual=None,
+            eps=self.variance_epsilon,
+            dropout_p=0.0,
+            prenorm=False,
+            residual_in_fp32=False,
+        )
+```
```
LysandreJik, line 133: That's nice!

ArthurZucker (approved):
Looks very much ready to me! Thanks a lot!

LysandreJik (approved):
Tried it locally, no red flags! Let's merge it and integrate it in transformers?

PR #60, opened 2025-03-27
build-output instead of build

MekkCyber:
When experimenting with building kernels, i noticed we the output folder is no longer called `build` instead it's `build-output`, so I changed all the occurences of `build`.

### DIFF
--- a/src/kernels/utils.py
+++ b/src/kernels/utils.py
@@ -80,15 +80,15 @@ def install_kernel(
     repo_path = Path(
         snapshot_download(
             repo_id,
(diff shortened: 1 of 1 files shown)


src/kernels/utils.py
```diff
@@ -80,15 +80,15 @@ def install_kernel(
     repo_path = Path(
         snapshot_download(
             repo_id,
-            allow_patterns=[f"build/{variant}/*", f"build/{universal_variant}/*"],
+            allow_patterns=[f"build-output/{variant}/*", f"build-output/{universal_variant}/*"],
```
danieldk, line 83: It should still be `build`. We should modify the Docker image to output to `build` instead.

PR #68, opened 2025-04-08
Do not use kernels without backward when training

danieldk:
Kernel layers can be marked to be inference-only by setting the `has_backward` class variable to `False`. If a layer is in training mode, we will use the fallback `forward`.


ArthurZucker (approved):
🚀

PR #69, opened 2025-04-09
Add `has_kernel` function

danieldk:
This function checks whether a kernel build exists for the current environment (Torch version and compute framework).

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -41,6 +41,9 @@ def test_gelu_fast(kernel, device):
     [
         ("kernels-community/activation", "main", True),
         ("kernels-community/triton-layer-norm", "main", True),
+        # Repo only contains Torch 2.4 kernels (and we don't
+        # support/test against this version).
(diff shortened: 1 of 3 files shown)


tests/test_basic.py
```diff
@@ -36,6 +36,19 @@ def test_gelu_fast(kernel, device):
...
 
+@pytest.mark.parametrize(
+    "kernel_exists",
+    [
+        ("kernels-community/activation", "main", True),
+        ("kernels-community/triton-layer-norm", "main", True),
+        ("google-bert/bert-base-uncased", "87565a309", False),
+    ],
+)
+def test_has_kernel(kernel_exists):
```
Narsil, line 50: Shouldn't you have a test against something like torch 2.0 for instance, since it seems to be the main use of this function ?

danieldk: Added testing against a repo that only contains a Torch 2.4 kernel (and `kernels` only supports >= 2.5).

Narsil (approved):
LGTM

PR #70, opened 2025-04-09
Support `DISABLE_KERNEL_MAPPING` env var for completely disabling kernel mappings

danieldk:
Also:

* Rename `HF_KERNELS_CACHE` to `KERNELS_CACHE`. The old name is from when this package was still called `hf-kernels`. The old name is still recognized, but emits a warning.
* Add documentation for environment variables.


ArthurZucker (approved):
Much needed thanks!

PR #78, opened 2025-04-22
Add tests to validate that `kernels` is `torch.compile` compatible

danieldk:


Narsil (commented):
Will this fail with improper stream on kernel ? or simply skip the cuda graph part ?

It should be easy to try with a simply graphing no (which should fail if it's using stream 0 which is the default which is not torch default stream)

danieldk:
Closing this for now since it's incomplete (doesn't test full graph compiles).

PR #79, opened 2025-04-23
Allow layers to opt in to `torch.compile`

danieldk:
This change allows a layer to set the `can_torch_compile` class variable to indicate that the layer is compatible with `torch.compile`. When enabled, the layer does not fall back to the original implementation when `torch.compile` is used.

### DIFF
--- a/src/kernels/layer.py
+++ b/src/kernels/layer.py
@@ -138,6 +138,8 @@ def replace_kernel_forward_from_hub(cls, layer_name: str, *, use_fallback: bool 
             return fallback_forward(self, x, *args, **kwargs)
 
         needs_backward = self.training
+        is_compiling = _is_torchdynamo_compiling()
+
         kernel = _KERNEL_MAPPING.get().get(layer_name)
         if kernel is None:
             warnings.warn(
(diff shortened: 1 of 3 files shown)


src/kernels/layer.py
```diff
@@ -165,7 +167,13 @@ def forward(self, x, *args, **kwargs):
         # Short-circuit if we already loaded the layer.
         layer = cached_layer.get(repo, None)
         if layer is not None:
-            if needs_backward and not getattr(layer, "has_backward", True):
+            # Switch to fallback when the layer does not support
+            # compilation/compile when needed.
```
MekkCyber:
```suggestion
            # Switch to fallback when the layer does not support:
            # compilation/compile when needed.
            # backward when needed
```

src/kernels/layer.py
```diff
@@ -245,7 +260,7 @@ def _validate_layer(*, check_cls, cls):
     torch_module_members = {name for name, _ in inspect.getmembers(nn.Module)}
     cls_members = {name for name, _ in inspect.getmembers(cls)}
     difference = cls_members - torch_module_members
-    if difference != set() and difference != {"has_backward"}:
+    if not difference <= {"can_torch_compile", "has_backward"}:
```
MekkCyber, line 265:
```suggestion
    # verify if : difference ⊄ {"can_torch_compile", "has_backward"}
    if not difference <= {"can_torch_compile", "has_backward"}:
```

MekkCyber (approved):
Thanks for this PR ! LGTM

PR #80, opened 2025-05-05
Updates _DISABLE_KERNEL_MAPPING to be dynamically set

MekkCyber:
Updates `_DISABLE_KERNEL_MAPPING` to be dynamic rather than fixed at import,  
so `DISABLE_KERNEL_MAPPING` can now be set using `os.environ` within a Python script to disable the kernel.

### DIFF
--- a/src/kernels/layer.py
+++ b/src/kernels/layer.py
@@ -135,7 +135,6 @@ def replace_kernel_forward_from_hub(cls, layer_name: str, *, use_fallback: bool 
     cached_layer: Dict[LayerRepository, nn.Module] = {}
 
     def forward(self, x, *args, **kwargs):
-        print("disable kernel mapping", _DISABLE_KERNEL_MAPPING())
         if _DISABLE_KERNEL_MAPPING():
             return fallback_forward(self, x, *args, **kwargs)
 
--- a/src/kernels/layer.py
+++ b/src/kernels/layer.py
@@ -11,7 +11,8 @@ from .utils import get_kernel
 if TYPE_CHECKING:
     from torch import nn
 
-_DISABLE_KERNEL_MAPPING: bool = bool(int(os.environ.get("DISABLE_KERNEL_MAPPING", "0")))
+def _DISABLE_KERNEL_MAPPING():
+    return bool(int(os.environ.get("DISABLE_KERNEL_MAPPING", "0")))
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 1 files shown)


src/kernels/layer.py
```diff
@@ -11,7 +11,8 @@
 if TYPE_CHECKING:
     from torch import nn
 
-_DISABLE_KERNEL_MAPPING: bool = bool(int(os.environ.get("DISABLE_KERNEL_MAPPING", "0")))
+def _DISABLE_KERNEL_MAPPING():
+    return bool(int(os.environ.get("DISABLE_KERNEL_MAPPING", "0")))
```
drbh, line 15: while the impact is likely minimal I think it may be best to avoid directly reading env vars in each forward pass. 

maybe we can cache the values after the first invocation, or another pattern to keep the hot path as simple as possible?

maybe something like?

```python
def _get_disable_kernel_mapping() -> bool:
    if not hasattr(_get_disable_kernel_mapping, "cached_value"):
        _get_disable_kernel_mapping.cached_value = bool(int(os.environ.get("DISABLE_KERNEL_MAPPING", "0")))
    return _get_disable_kernel_mapping.cached_value
```

drbh (commented):
nice improvement, just left a small comment! Thanks!

src/kernels/layer.py
danieldk: Then it would it would not be picked up when the env var is set, but then changed in a Python scrip later. But I think setting environment variables in a Python script is a bit of an anti-pattern.

I think in general this kind of global state should be immutable. Otherwise you can end up in a situation where module A wants kernels to be enabled, but it also happens to call module B and module B disables it by tweaking the env var, breaking the invariant in module A (if B doesn't reset the environment variable). It gets even worse with threading added to the mix. Context vars work much better in handling that kind of global state.

I am curious about the use case: under what circumstances should a script set it and not the user? I'd rather fix those cases where `kernels` need to be disabled because they are not compatible.

---

If we do want to enable control over this from code, maybe we need a context manager that exposes this in a nicer way?

```
with disable_kernels():
    ...
```

or more generic

```
with enable_kernels(False):
    ...
```

Where the default of the context var is set to `DISABLE_KERNEL_MAPPING` when the module is imported?

(Though even in this case, the semantics are not completely clear to me, if some inner context enables kernels and an outer does not, do we follow the wishes of each context, or does the outer context overrule the inner?)

MekkCyber: Thanks for the reviews @drbh @danieldk. I understand the concerns this change raises and why immutability is important.

My main use case is benchmarking the two implementations—with and without kernels—directly from Python. For that, it’s much simpler to set the environment variable in Python code.

Even looking ahead, I believe there may be cases where we’ll want to disable the kernels programmatically from Python.

The idea of using context is definitely worth considering. However, perhaps we can find a way to ensure the setting remains immutable during each forward pass.

Would that be a reasonable compromise if it’s technically feasible?

danieldk: Maybe we can revisit it after reimplementing the mapping using thread-local, this might give us some insights about this case as well?

danieldk: I think we can close this now that `kernelize` has landed?

MekkCyber: yup I agree

PR #84, opened 2025-05-08
kernels: add the `to-wheel` subcommand

danieldk:
This subcommand accepts a kernel repo and version as arguments:

    kernels to-wheel kernels-community/activation 0.0.3

Wheels will then be generated for every build variant.


drbh (approved):
nice! lgtm

PR #87, opened 2025-05-13
Make the forward pass `torch.compile` compatible

MekkCyber:
This PR updates the flow of replace_kernel_forward_from_hub to ensure the appropriate forward pass is selected before model compilation. It introduces a new function `kernelize` that can be called to replace the forward before the model is compiled. It can also be called after registering new kernels for more flexibility

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -23,33 +23,98 @@ class SiluAndMul(nn.Module):
         return F.silu(input[..., :d]) * input[..., d:]
 ```
 
-The decorator changes the layer, so that other implementations of the `forward`
-method can be registered using the name `SiluAndMul`.
+The decorator does not change the behavior of the class -- it annotates
+the class with the given name (here `SiluAndMul`). The `kernelize` function
+described below uses this name to look up kernels for the layer.
 
 ### External layers
 
 An existing layer that does not (yet) have the `use_kernel_forward_from_hub`
-decorator can be made extensible by by monkeypatching it using the `replace_kernel_forward_from_hub` function.
+decorator can be made extensible using the `replace_kernel_forward_from_hub`
+function:
 
 ```python
 from somelibrary import SiluAndMul
 
 replace_kernel_forward_from_hub(SiluAndMul, "SiluAndMul")
-register_kernel_mapping(kernel_layer_mapping)
 ```
-
-The `register_kernel_mapping` call maps the name `SiluAndMul` to actual
-hub kernels. See the [Registering a hub kernel for a layer](#registering-a-hub-kernel-for-a-layer)
-section for more information.
 
 **Warning:** we strongly recommend using layers with a decorator, since
 it signifies that the maintainer intends to keep the `forward` signature
 compatible with layers from the hub.
 
+## Kernelizing a model
+
+A model will not use Hub kernels by default, even if it contains extensible
+layers. To enable the use of Hub kernels in the model, it needs to be
+'kernelized' using the `kernelize` function. This function traverses the
+model graph and replaces the `forward` methods of extensible layers for which
+Hub kernels are registered. Kernelize can be used as follows:
+
+```python
+model = MyModel(...)
+model = kernelize(model)
+```
+
+**Note:** the `kernelize` function modifies the model in-place, the model
+itself is returned as a convenience.
+
+### Kernel device
+
+Kernels can be registered per device type. For instance, separate `cuda` and
+`metal` kernels could be registered for the name `SiluAndMul`. By default,
+`kernelize` will try to infer the device type from the model's parameters.
+You can pass the device type to `kernelize` if the device type cannot be
+inferred (e.g. because the model has no parameters):
+
+```python
+model = MyModel(...)
+model = kernelize(model, device="cuda")
+```
+
+### Backprop
+
+Not all Hub kernels support backprop. If a model is going to be used with
+backprop for training, then we want to avoid that `kernelize` uses such
+kernels. By default, `kernelize` will exclude inference-only kernels when
+the kernelized model is set to training (`Model.train()`). You can override
+this behavior using the `needs_backwards`:
+
+```python
+model = MyModel(...)
+model = kernelize(model, needs_backward=False)
+```
+
+### `torch.compile`
+
+Not all Hub kernels support `torch.compile`. If you want to compile a model
+after kernelizing it, pass the `needs_torch_compile` argument to ensure that
+only kernels that support `torch.compile` will be loaded:
+
+```python
+model = MyModel(...)
+model = kernelize(model, needs_torch_compile=True)
+```
+
+### Fallback forward
+
+The `needs_backwards`/`needs_torch_compile` arguments will fall back to
+the layer's original `forward` if the registered kernels do not support
+backprop/`torch.compile`. You can let `kernelize` raise an exception
+instead by using `use_fallback=False`:
+
+```python
+model = MyModel(...)
+model = kernelize(model, needs_torch_compile=True, use_fallback=False)
+```
+
+This can be useful if you want to guarantee that Hub kernels are used.
+
 ## Registering a hub kernel for a layer
 
-Once a layer is made extensible, users can register hub kernels for it
-by name using the `register_kernel_mapping` function. For example:
+`kernelize`` relies on kernel mappings to find Hub kernels for layers.
+Kernel mappings map a kernel name such as `SiluAndMul` to a kernel on
+the Hub. For example:
 
 ```python
 kernel_layer_mapping = {
@@ -61,7 +126,11 @@ kernel_layer_mapping = {
         )
     }
 }
+```
 
+You can register such a mapping using `register_kernel_mapping`:
+
+```python
 register_kernel_mapping(kernel_layer_mapping)
 ```
 
@@ -72,7 +141,7 @@ used with the `use_kernel_mapping` context manager:
 ```python
 with use_kernel_mapping(kernel_layer_mapping):
     # Use the layer for which the mapping is applied.
-    ...
+    model = kernelize(model)
 ```
 
 This ensures that the mapping is not active anymore outside the
--- a/src/kernels/layer.py
+++ b/src/kernels/layer.py
@@ -253,9 +253,7 @@ def kernelize(
     return model
 
 
-def use_kernel_forward_from_hub(
-    layer_name: str, *, device: Device = Device(type="cuda"), use_fallback: bool = True
-):
+def use_kernel_forward_from_hub(layer_name: str):
     """
     Replace the forward function of a layer using a layer from the kernel hub.
     This decorator can be applied to a layer and replaces the forward method
@@ -328,22 +326,6 @@ def _validate_layer(*, check_cls, cls):
             raise TypeError(
                 f"Forward signature does not match: different kind of arguments ({param} ({param.kind}) and {ref_param} ({ref_param.kind})"
             )
-
-
-def _is_torchdynamo_compiling():
-    # Importing torch._dynamo causes issues with PyTorch profiler (https://github.com/pytorch/pytorch/issues/130622)
-    # hence rather relying on `torch.compiler.is_compiling()` when possible (torch>=2.3)
-    try:
-        import torch
-
-        return torch.compiler.is_compiling()
-    except Exception:
-        try:
-            import torch._dynamo as dynamo  # noqa: F401
-
-            return dynamo.is_compiling()
-        except Exception:
-            return False
 
 
 def _find_device(model: "nn.Module") -> Device:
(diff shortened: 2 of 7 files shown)


src/kernels/layer.py
```diff
@@ -119,96 +111,118 @@ def register_kernel_mapping(
...
 
 
-def replace_kernel_forward_from_hub(cls, layer_name: str, *, use_fallback: bool = True):
-    """
-    Replace the forward function of a layer using a layer from the kernel hub.
-    This function monkeypatches a layer, replacing the `forward` method
-    of the layer with that of a layer from the hub. The replacement is done
-    when a layer matching `layer_name` and device type is registered through
-    `register_layer_mapping`. The device type is inferred from the first
-    argument to `forward`.
```
danieldk, line 129: The docstring got removed.

src/kernels/layer.py
```diff
@@ -119,96 +111,118 @@ def register_kernel_mapping(
...
+def replace_kernel_forward_from_hub(
+    cls,
+    layer_name: str,
+    *,
+    device: Device = Device(type="cuda"),
+    needs_backward: bool = False,
+    use_fallback: bool = True,
+):
+    cls.kernel_layer_name = layer_name
+    kernelize(cls, traverse_graph=False, device=device, needs_backward=needs_backward, use_fallback=use_fallback)
```
danieldk: I think we should remove `kernelize` here and remove the `needs_backward` and `device` parameters from `replace_kernel_forward_from_hub`. This function and the decorator should just set up a class <-> name mapping and not mutate the classes (we would also lose the fallbacks).

src/kernels/layer.py
```diff
@@ -119,96 +111,118 @@ def register_kernel_mapping(
...
 
-    fallback_forward = cls.forward
 
-    cached_layer: Dict[LayerRepository, nn.Module] = {}
+def kernelize(
+    module_,
+    traverse_graph: bool = True,
+    device: Device = Device(type="cuda"),
+    needs_backward: bool = False,
```
danieldk: We can get this from `module_` itself by checking whether it's in eval or training mode.

src/kernels/layer.py
```diff
@@ -119,96 +111,118 @@ def register_kernel_mapping(
...
+    use_fallback: bool = True,
+):
+    cls.kernel_layer_name = layer_name
+    kernelize(cls, traverse_graph=False, device=device, needs_backward=needs_backward, use_fallback=use_fallback)
 
-    fallback_forward = cls.forward
 
-    cached_layer: Dict[LayerRepository, nn.Module] = {}
+def kernelize(
+    module_,
```
danieldk:
```suggestion
    module: nn.Module,
```

src/kernels/layer.py
```diff
@@ -119,96 +111,118 @@ def register_kernel_mapping(
...
-                raise ValueError(
-                    f"No layer mapping for `{layer_name}` with device type `{device.type}`"
+    # We assume only one Module per repository.
+    cached_layer: Dict[LayerRepository, nn.Module] = {}
+    is_compiling = _is_torchdynamo_compiling()
+    # If we don't traverse the graph, we stop after the first module
+    modules_list = [("", module_)] if not traverse_graph else module_.named_modules()
+
+    for _, module in modules_list:
+        if hasattr(module, "kernel_layer_name") or not traverse_graph:
```
danieldk: What is the ` or not traverse_graph` condition for? I the layer has no kernel layer name, there is nothing to do, right?

src/kernels/layer.py
```diff
@@ -119,96 +111,118 @@ def register_kernel_mapping(
...
-    """
+def replace_kernel_forward_from_hub(
+    cls,
+    layer_name: str,
+    *,
+    device: Device = Device(type="cuda"),
+    needs_backward: bool = False,
+    use_fallback: bool = True,
+):
+    cls.kernel_layer_name = layer_name
```
danieldk, line 145: Maybe we should also store the fallback forward? Then we would really have the proper fallback even if someone runs `kernelize` multiple times.

src/kernels/layer.py
```diff
@@ -119,96 +111,118 @@ def register_kernel_mapping(
...
 
-    cls.forward = forward
+            # Switch to fallback when the layer does not support
+            # compilation/compile when needed.
+            needs_fallback = needs_backward and not getattr(layer, "has_backward", True)
+            needs_fallback |= is_compiling and not getattr(layer, "can_torch_compile", False)
+            if needs_fallback:
+                module.forward = fallback_forward
+                continue
+            module.forward = layer.forward
```
danieldk:
```suggestion
            if needs_fallback:
                module.forward = fallback_forward
            else:
                module.forward = layer.forward
```

src/kernels/layer.py
```diff
@@ -119,96 +111,118 @@ def register_kernel_mapping(
...
 
 
-def use_kernel_forward_from_hub(layer_name: str, *, use_fallback: bool = True):
+def use_kernel_forward_from_hub(
+    layer_name: str, *, device: Device = Device(type="cuda"), needs_backward: bool = False, use_fallback: bool = True
```
danieldk: See above.

src/kernels/layer.py
MekkCyber: My idea was that we can make `kernelize` handle two cases, swap the forwards during the import, and give the possibility to swap even after the import, if some new kernels are added if `kernel_layer_name` is set, in the first case, we don't traverse the model graph, we only do that in the second case where we loop over the modules

src/kernels/layer.py
MekkCyber: Nice catch

src/kernels/layer.py
MekkCyber: but where would we call `kernelize` then ? And it's not really efficient to loop after the model is already imported I think

danieldk: One issue that I see with this is that previously, you could simply disable kernels at no cost by e.g. using an empty mapping in a context manager. Now disabling it (if not done using the env var) entails rewriting the whole graph, whereas I think for the time being _not_ using kernels should be the simple/cheap case.

There was also a discussion on Slack that using kernels should probably be opt-in for the time being. One reason being that kernels tend to introduce more non-determinism, which is not nice for tests, etc. Explicitly calling `kernelize` facilitates that. It's pretty similar to `torch.compile` being opt-in. If we return the model itself in `kernelize`, users could even do

```
compiled = torch.compile(kernelize(model, ...), ...)
```

and have an optimized model.

MekkCyber: To make avoiding kernels the simplest option, we could add a new argument to `kernelize` defaulting to `True` that explicitly disables the kernels. What do you think?

As for rewriting the whole graph, I don’t see it as an issue. I’d prefer to keep the forward pass fast by avoiding any kernel-switching logic there. Disabling kernels once via a call to `kernelize` seems much better than checking conditions repeatedly during each forward pass, added to that the compile compatibility concerns.

src/kernels/layer.py
```diff
@@ -119,96 +110,125 @@ def register_kernel_mapping(
...
+                # Switch to fallback when the layer does not support:
+                # compilation/compile when needed.
+                # backward when needed
+                needs_fallback = needs_backward and not getattr(layer, "has_backward", True)
+                needs_fallback |= is_compiling and not getattr(layer, "can_torch_compile", False)
+                if needs_fallback:
+                    module_class.forward = fallback_forward
+                    continue
+                module_class.forward = layer.forward
+                continue
```
danieldk:
```suggestion
                if needs_fallback:
                    module_class.forward = fallback_forward
                else:
                    module_class.forward = layer.forward
                continue
```

src/kernels/layer.py
```diff
@@ -119,96 +110,125 @@ def register_kernel_mapping(
...
+        cls: The layer class to decorate
+        layer_name: The name of the layer to use for kernel lookup
     """
+    cls.kernel_layer_name = layer_name
+    cls.fallback_forward = cls.forward
 
-    fallback_forward = cls.forward
 
-    cached_layer: Dict[LayerRepository, nn.Module] = {}
+cached_layer: ContextVar[Dict[LayerRepository, nn.Module]] = ContextVar("cached_layer", default={})
```
danieldk: Since this is now a global, I'd put it at the top of the file and name it `_CACHED_LAYER`.

We don't really use this as a context var, so I think this could just be a regular `Dict`?

src/kernels/layer.py
```diff
@@ -119,96 +110,125 @@ def register_kernel_mapping(
...
 
-    def forward(self, x, *args, **kwargs):
-        if _DISABLE_KERNEL_MAPPING:
-            return fallback_forward(self, x, *args, **kwargs)
+def kernelize(
+    model: nn.Module,
+    device: Device = Device(type="cuda"),
```
danieldk: I think it would be nicer if this was:

```suggestion
    device: Optional[Device] = None,
```

Behavior: if the user passes a device type, use it. If the user does not pass a device type, get it from a (probably the first, since individual layers may not have parameters) parameter of the model. Similarly, I would add an extra argument:

```
    needs_backward: Optional[bool] = None,
```

If the user sets a bool, then do as they wish. Otherwise, use the logic that is already there below. This has the benefit that if a model is in `eval()`, the user could still kernelize it with kernels that support backward in case they want to move to `train()` later.

Then, with these new options, I would make everything after `model` kwarg-only. Positional arguments lead to a lot of bugs/failures.

src/kernels/layer.py
```diff
@@ -8,8 +8,7 @@
 
 from .utils import get_kernel
 
-if TYPE_CHECKING:
-    from torch import nn
+from torch import nn
```
danieldk: Needs to be conditional, `torch` is an optional dependency.

danieldk (commented):
Looks great! Added some comments to polish it up.

src/kernels/layer.py
```diff
@@ -119,96 +110,125 @@ def register_kernel_mapping(
...
-    of the layer with that of a layer from the hub. The replacement is done
-    when a layer matching `layer_name` and device type is registered through
-    `register_layer_mapping`. The device type is inferred from the first
-    argument to `forward`.
+    Decorator that prepares a layer class to use a kernel from the Hugging Face Hub.
+    
+    This decorator stores the layer name and original forward method, which will be used
+    by the kernelize function to replace the forward implementation with the appropriate
+    kernel from the hub.
+
```
ArthurZucker: let's explain why we don't replace on the fly (cf compile)

src/kernels/layer.py
```diff
@@ -119,96 +110,125 @@ def register_kernel_mapping(
...
-        repo = kernel.get(Device(type=device.type))
-        if repo is None:
-            if not use_fallback:
-                raise ValueError(
-                    f"No layer mapping for `{layer_name}` with device type `{device.type}`"
+    Returns:
+        The kernelized model
+    """
+    is_compiling = _is_torchdynamo_compiling()
+    needs_backward = model.training
```
ArthurZucker: does this mean we cannot adapt to the user changing `model.training` . 
Would be nice if we can register two forwards, and switch between them on the fly!

ArthurZucker (commented):
Nice!

src/kernels/layer.py
danieldk: I added a FAQ with this question in it.

src/kernels/layer.py
danieldk: Yeah, that would be nice! I think to support this we would have to override `Module.train` and switch forwards based on the value that is passed. I think it makes sense that I do this in a PR after this, since this one is already getting pretty big.

Then we'll remove the `needs_backward` argument again in the next PR (which should be fine as long we don't make a release).

docs/layers.md
```diff
@@ -23,33 +23,98 @@ class SiluAndMul(nn.Module):
...
+A model will not use Hub kernels by default, even if it contains extensible
+layers. To enable the use of Hub kernels in the model, it needs to be
+'kernelized' using the `kernelize` function. This function traverses the
+model graph and replaces the `forward` methods of extensible layers for which
+Hub kernels are registered. Kernelize can be used as follows:
+
+```python
+model = MyModel(...)
+model = kernelize(model)
+```
```
LysandreJik, line 57: I like that opt-in, that's clean

docs/layers.md
```diff
@@ -23,33 +23,98 @@ class SiluAndMul(nn.Module):
...
 
-The `register_kernel_mapping` call maps the name `SiluAndMul` to actual
-hub kernels. See the [Registering a hub kernel for a layer](#registering-a-hub-kernel-for-a-layer)
-section for more information.
-
 **Warning:** we strongly recommend using layers with a decorator, since
 it signifies that the maintainer intends to keep the `forward` signature
 compatible with layers from the hub.
 
+## Kernelizing a model
```
LysandreJik, line 46: Very clear doc

LysandreJik (approved):
Very nice! Love the new additions and the clear doc, thanks for spending time on this

PR #89, opened 2025-05-24
Add support for Metal builds

danieldk:
This change add initial support for loading Metal kernels on macOS. CI will be hooked up in a later PR, once we have sorted out how to get runners with Metal/MPS.

### DIFF
--- a/src/kernels/utils.py
+++ b/src/kernels/utils.py
@@ -43,13 +43,21 @@ def build_variant() -> str:
     elif torch.version.hip is not None:
         rocm_version = parse(torch.version.hip.split("-")[0])
         compute_framework = f"rocm{rocm_version.major}{rocm_version.minor}"
+    elif torch.backends.mps.is_available():
+        compute_framework = "metal"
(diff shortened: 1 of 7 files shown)


src/kernels/utils.py
```diff
@@ -43,14 +43,22 @@ def build_variant() -> str:
     elif torch.version.hip is not None:
         rocm_version = parse(torch.version.hip.split("-")[0])
         compute_framework = f"rocm{rocm_version.major}{rocm_version.minor}"
+    elif torch.backends.mps.is_available():
+        compute_framework = "metal"
```
LysandreJik, line 47: There's no versioning for metal yet?

LysandreJik (approved):
Looks good to me! Thanks @danieldk

src/kernels/utils.py
danieldk: It's tied to the macOS version and similar to Linux I think we'll just put the minimum version in the specs.

PR #90, opened 2025-06-03
Automatic fallback for kernels that don't support training

danieldk:
For kernels that do not support backward, fall back to the original implementation if `model.train(True)` is called. This removes the need for the `needs_backward` argument of `kernelize`.


drbh (approved):
LGTM!

PR #102, opened 2025-07-01
Add `get_local_kernel` function

danieldk:
This function loads a kernel from a local repository (e.g. the output of kernel-builder), which can be handy for testing.


EricLBuehler (approved):
Looks good!

PR #103, opened 2025-07-03
Support registering inference/training-specific layers

danieldk:
This change makes it possible to register kernels specialized for inference, training, and/or `torch.compile`. To do so, the mapping notation is extended to support registering specialized kernels for a specific 'mode'. For instance, the following mapping,

```python
kernel_layer_mapping = {
    "SiluAndMul": {
        "cuda": {
          Mode.DEFAULT: LayerRepository(
              layer_name="SiluAndMul",
          ),
          Mode.TRAINING | Mode.TORCH_COMPILE: LayerRepository(
              repo_id="kernels-community/activation-training-optimized",
              layer_name="SiluAndMul",
          ),
      }
    }
}
```

uses `kernels-community/activation` by default, but will switch to using `kernels-community/activation-training-optimized` if a model is kernelized for training and `torch.compile`. The 'old' way of registering kernels is still supported and will register a kernel as `Mode.DEFAULT`.

To make it easier to add more modes in the future and to unify the `register_kernel_mapping` and `kernelize` signatures, the `training` and `needs_torch_compile` arguments of `kernelize` are replaced by a single `mode` argument:

```python
model = MyModel(...)
model = kernelize(model, mode=Mode.TRAINING | Mode.TORCH_COMPILE)
```

This change also removes the automatic `forward` updates when switching a model from eval to training or vice versa (we prefer explicitness).

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -164,9 +164,9 @@ kernel_layer_mapping = {
 }
 ```
 
-The kernels will matched exactly on the mode. So, for instance, no kernel
-layer is used when the `mode` passed to `kernelized` is
-`Mode.INFERENCE | Mode.TORCH_COMPILE` or `Mode.TRAINING`. If you want to
+The kernels will match exactly on the mode. So, for instance in the example above, no kernel
+layer is used when the `mode` passed to `kernelize` is
+`Mode.INFERENCE | Mode.TORCH_COMPILE` or `Mode.TRAINING`. However, if you want to
 register a kernel to be used when the mode does not match any of the
 modes in the mapping, you can use the special `Mode.DEFAULT` mode to do
 so. For example:
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -53,7 +53,15 @@ Hub kernels are registered. Kernelize can be used as follows:
 
 ```python
 model = MyModel(...)
-model = kernelize(model)
+model = kernelize(model, mode=Mode.INFERENCE)
+```
+
+The `mode` specifies that the model will be used in inference. Similarly,
+you can ask `kernelize` to prepare the model for training:
+
+```python
+model = MyModel(...)
+model = kernelize(model, mode=Mode.TRAINING)
 ```
 
 **Note:** the `kernelize` function modifies the model in-place, the model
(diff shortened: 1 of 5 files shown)


docs/layers.md
```diff
@@ -132,3 +140,58 @@ with use_kernel_mapping(kernel_layer_mapping):
...
+      }
+    }
+}
+```
+
+The kernels will matched exactly on the mode. So, for instance, no kernel
+layer is used when the `mode` passed to `kernelized` is
+`Mode.INFERENCE | Mode.TORCH_COMPILE` or `Mode.TRAINING`. If you want to
+register a kernel to be used when the mode does not match any of the
+modes in the mapping, you can use the special `Mode.DEFAULT` mode to do
```
MekkCyber:
```suggestion
The kernels will match exactly on the mode. So, for instance in the example above, no kernel
layer is used when the `mode` passed to `kernelize` is
`Mode.INFERENCE | Mode.TORCH_COMPILE` or `Mode.TRAINING`. However, if you want to
register a kernel to be used when the mode does not match any of the
modes in the mapping, you can use the special `Mode.DEFAULT` mode to do
```

src/kernels/layer.py
```diff
@@ -17,6 +27,38 @@
...
+    Kernelize mode
+
+    The `Mode` flag is used by `kernelize` to select kernels for the given
+    mode. Mappings can be registered for specific modes.
+
+    * `INFERENCE`: The kernel is used for inference.
+    * `TRAINING`: The kernel is used for training.
+    * `TORCH_COMPILE`: The kernel is used with `torch.compile`.
+    * `DEFAULT`: In a kernel mapping, this kernel that is used when no other mode
+       matches.
```
MekkCyber:
```suggestion
    * `DEFAULT`: In a kernel mapping, this kernel is used when no other mode
       matches.
```

src/kernels/layer.py
```diff
@@ -171,6 +240,9 @@ def kernelize(
     """
     import torch
 
+    if mode == Mode.DEFAULT:
+        raise ValueError("Mode.DEFAULT can only be used to register kernel mappings.")
+
```
MekkCyber, line 248: why raising an error here ?

src/kernels/layer.py
```diff
@@ -331,45 +406,72 @@ def _conditionally_replace_forward(
...
-    # compilation/compile when needed.
-    # backward when needed
-    needs_fallback = needs_torch_compile and not getattr(
+    # Switch to fallback if the mode is not supported
+    # by the given layer.
+    needs_fallback = Mode.TORCH_COMPILE in mode and not getattr(
         layer, "can_torch_compile", False
     )
+    needs_fallback |= Mode.TRAINING in mode and not getattr(layer, "has_backward", True)
+
```
MekkCyber: aren't we calling `_validate_layer_has_mode` before to raise an error if the repo don't support the mode ?

drbh (dismissed):
LGTM!

src/kernels/layer.py
drbh: +1

src/kernels/layer.py
danieldk: It's a mode that can only be used to register a default kernel. We want people to pick inference or training. Though I realize now that we should check that `TRAINING` or `INFERENCE` is in `mode`.

src/kernels/layer.py
danieldk: Yeah, but that does not apply to the `DEFAULT` layer (since it is not registered as a specific mode), which is what this fallback is for. I'll update the comment to clarify this.

src/kernels/layer.py
danieldk: Added some checks.

src/kernels/layer.py
danieldk: (Hopefully) clarified by extending the comment.

src/kernels/layer.py
```diff
@@ -331,45 +415,75 @@ def _conditionally_replace_forward(
...
 
-    # Switch to fallback when the layer does not support:
-    # compilation/compile when needed.
-    # backward when needed
-    needs_fallback = needs_torch_compile and not getattr(
+    # Switch to fallback if the mode is not supported by the layer.
+    # Note that this is useful even after _validate_layer_has_mode because
+    # layers registered with the DEFAULT mode never get rejected by
+    # _validate_layer_has_mode. For such layers, we want to fall back in
+    # case the layer does not support the given mode.
```
MekkCyber, line 427: Cool ! thanks for the explanation

MekkCyber (approved):
Sounds good ! This is a very cool and useful feature 🔥

PR #107, opened 2025-07-15
Set `kernelize` default mode to `Mode.TRAINING | Mode.TORCH_COMPILE`

danieldk:
Also update docs and tests.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -49,12 +49,15 @@ A model will not use Hub kernels by default, even if it contains extensible
 layers. To enable the use of Hub kernels in the model, it needs to be
 'kernelized' using the `kernelize` function. This function traverses the
 model graph and replaces the `forward` methods of extensible layers for which
-Hub kernels are registered. Kernelize can be used as follows:
+Hub kernels are registered. `kernelize` can be used as follows:
 
 ```python
 model = MyModel(...)
 model = kernelize(model, mode=Mode.INFERENCE)
 ```
+
+The `kernelize` function modifies the model in-place, the model
+itself is returned as a convenience.
 
 The `mode` specifies that the model will be used in inference. Similarly,
 you can ask `kernelize` to prepare the model for training:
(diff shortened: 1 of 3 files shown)


docs/layers.md
```diff
@@ -196,6 +202,31 @@ In this case, modes other than `Mode.INFERENCE` and
...
+
+1. Kernel `A` is registered for `Mode.DEFAULT`. This kernel supports training
+   (backward), but not `torch.compile`.
+2. Kernel `B` is registered for `Mode.INFERENCE | Mode.COMPILE` and supports
+   `torch.compile`.
+
+`kernelize` modes will then behave as follows:
+
+- `Mode.INFERENCE | Mode.COMPILE`` uses kernel `B`: exact match.
+- `Mode.INFERENCE` uses kernel `A`: no exact match, so fall back to
```
ArthurZucker, line 222:
```suggestion
- `Mode.INFERENCE` uses kernel `B`: no exact match, so fall back to
```

docs/layers.md
```diff
@@ -196,6 +202,31 @@ In this case, modes other than `Mode.INFERENCE` and
...
+  `Mode.DEFAULT`.
```
ArthurZucker, line 223:
```suggestion
  `Mode.DEFAULT`, thus running with kernel `B`.
```
or did I get it wrong?

docs/layers.md
```diff
@@ -196,6 +202,31 @@ In this case, modes other than `Mode.INFERENCE` and
...
+original layer's `forward` method will be used instead.
+
+As an example, suppose that two kernels were registered for a layer:
+
+1. Kernel `A` is registered for `Mode.DEFAULT`. This kernel supports training
+   (backward), but not `torch.compile`.
+2. Kernel `B` is registered for `Mode.INFERENCE | Mode.COMPILE` and supports
+   `torch.compile`.
+
+`kernelize` modes will then behave as follows:
```
ArthurZucker, line 219:
```suggestion
Using `kernelize(model)` will behave as follow:
```

src/kernels/layer.py
```diff
@@ -325,7 +325,7 @@ def _select_repository(
 def kernelize(
     model: "nn.Module",
     *,
-    mode: Mode,
+    mode: Mode = Mode.TRAINING | Mode.TORCH_COMPILE,
```
ArthurZucker, line 328: thanks!

PR #111, opened 2025-07-21
`get_kernel`: allow Python-style version specifiers

danieldk:
Use Python-style version specifiers to resolve to tags. E.g., given the presence of the tags `v0.1.0`, `v0.1.1`, and `v0.2.0`,

```python
get_kernel("my/kernel", version=">=0.1.0,<0.2.0")
```

would resolve to `v0.1.1`.

### DIFF
--- a/src/kernels/utils.py
+++ b/src/kernels/utils.py
@@ -16,6 +16,7 @@ from typing import Dict, List, Optional, Tuple
 from huggingface_hub import file_exists, snapshot_download
 from packaging.version import parse
 
+from kernels._versions import resolve_version_spec_as_ref
 from kernels.lockfile import KernelLock, VariantLock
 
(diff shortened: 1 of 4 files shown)


src/kernels/utils.py
```diff
@@ -182,13 +183,31 @@ def install_kernel_all_variants(
     return repo_path / "build"
 
 
-def get_kernel(repo_id: str, revision: str = "main") -> ModuleType:
+def get_kernel(
+    repo_id: str, revision: Optional[str] = None, version: Optional[str] = None
```
danieldk, line 187: `version` is probably nicer than `version_spec`?

drbh: yea strong agree

PR #120, opened 2025-07-29
Fix some spelling errors to check docs CI is working

danieldk:


MekkCyber (approved):
I think these can be merged now

PR #122, opened 2025-07-31
Add ROCm device discovery

ahadnagy:
PR for adding device discovery for the ROCm platform so we can differentiate if from CUDA.

### DIFF
--- a/tests/test_layer.py
+++ b/tests/test_layer.py
@@ -173,7 +173,9 @@ def test_rocm_kernel_mapping():
         # Verify the repository is correctly stored
         rocm_repos = mapping["SiluAndMul"]["rocm"]
         assert rocm_repos is not None
-        assert rocm_repos.repos[Mode.FALLBACK]._repo_id == "kernels-community/activation"
+        assert (
+            rocm_repos.repos[Mode.FALLBACK]._repo_id == "kernels-community/activation"
+        )
         assert rocm_repos.repos[Mode.FALLBACK].layer_name == "SiluAndMul"
 
 


ivarflakstad (approved):
My review has no power but this lgtm! ☺️

src/kernels/layer.py
```diff
@@ -690,6 +767,31 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+        dev_type = param.device.type
+        if dev_type == "cuda":
+            # Refine based on actual platform
+            if _is_rocm_platform():
+                return Device(type="rocm")
+            elif _is_cuda_platform():
+                return Device(type="cuda")
+
+        return Device(type=dev_type)
```
drbh: is there a specific reason that `_find_device` is nested within `kernelize`?

drbh (approved):
lgtm

src/kernels/layer.py
ahadnagy: The logic depends on `torch` now, and I didn't want to make the torch import global, that's just it.

ahadnagy: Moving it outside :)

danieldk (commented):
I think it's useful to be able to register using a shorthand, e.g.:

```
kernel_layer_mapping = {
    "SiluAndMul": {
        "rocm": LayerRepository(
            layer_name="SiluAndMul",
        )
    }
}
```

I think it already works with the PR as-is, but this needs a test. It doesn't actually have to run a kernel, it's enough to check that the mapping created with the shorthand is correct.

The device type needs to be documented in `layer.md` as well.

Maybe it's also nice to add a test for ROCm (with a mark to ensure that it only runs on machines with a ROCm GPU available).

docs/source/layers.md
```diff
@@ -270,3 +274,9 @@ Capabilities behave as follows:
   than 75..89. The motivation is that kernels with smaller ranges
   tend to be more optimized for a specific set of GPUs. **This behavior
   might still change in the future.**
+
+### Registering kernels for specific ROCm capabilities
+
+Registering kernels for the ROCm arhitecture follows the exact same
```
danieldk:
```suggestion
Registering kernels for the ROCm architecture follows the exact same
```

docs/source/layers.md
```diff
@@ -270,3 +274,9 @@ Capabilities behave as follows:
...
+pattern as CUDA kernels, including the selective compute capability
+with the respective ROCm version numbers.
```
danieldk:
```suggestion
pattern as CUDA kernels, using `min_capability` and `max_capability` to restrict
a kernel to a range of ROCm capabilities.
```

src/kernels/layer.py
```diff
@@ -693,7 +795,7 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
     if device is None:
         device_type = _find_device(model)
     elif isinstance(device, str):
-        device_type = Device(type=torch.device(device).type)
+        device_type = Device(type=device)
```
danieldk, line 849: This change looks incorrect. E.g., if someone erroneously passes `cuda:0` as the device type, then in the old code:

```
>>> torch.device("cuda:0").type
'cuda'
```

in the new code `cuda:0` would get stored in device.

src/kernels/layer.py
```diff
@@ -615,6 +691,33 @@ def _select_repository(
...
+    dev_type = param.device.type
+    if dev_type == "cuda":
+        # Refine based on actual platform
+        if _is_rocm_platform():
+            return Device(type="rocm")
+        elif _is_cuda_platform():
+            return Device(type="cuda")
+
+    return Device(type=dev_type)
+
```
danieldk: Could this me moved to the old location of `_find_devices`? Makes the diff cleaner/easier to bisect if needed.

tests/test_layer.py
```diff
@@ -124,6 +128,51 @@ def test_hub_forward(cls, device):
         assert silu_and_mul_with_kernel.n_calls == 1
 
 
+@pytest.mark.rocm_only
+def test_hub_forward_rocm():
+    torch.manual_seed(0)
+
+    silu_and_mul = SiluAndMul()
+    X = torch.randn((32, 64), device="cpu")  # Use CPU tensor since ROCm might not be available for tensor creation
```
danieldk: Doesn't `mark.rocm_only` already check that ROCm + GPUs are available?

tests/test_layer.py
```diff
@@ -124,6 +128,51 @@ def test_hub_forward(cls, device):
...
+    Y = silu_and_mul(X)
+
+    silu_and_mul_with_kernel = kernelize(SiluAndMulNoCompileKernel(), device="rocm", mode=Mode.INFERENCE)
+    Y_kernel = silu_and_mul_with_kernel(X)
```
danieldk, line 144: How can it work if the tensor is on the CPU?

tests/test_layer.py
```diff
@@ -954,3 +1003,6 @@ def forward(self) -> str:
                 }
             }
         )
+
+
+
```
danieldk: I think `black` needs to be run on the file. Seems like it's missing from the lints. I'll add it in a PR.

src/kernels/layer.py
ahadnagy: The issue with the old code is that PyTorch won't recognize `rocm` as a device and will throw an error. ROCm falls under CUDA in PyTorch land. Should we just check against the supported devices here and throw an error?

tests/test_layer.py
ahadnagy: It does, forgot to remove the check after I added the decorator.

src/kernels/layer.py
danieldk: Yeah, sounds good, so just rejecting things like `cuda:0` or `mps:0`, etc.

src/kernels/layer.py
```diff
@@ -887,6 +963,26 @@ def _validate_layer(*, check_cls, cls):
             )
 
 
+@lru_cache
+def _find_capability() -> int:
+    import torch
+
+    major, minor = torch.cuda.get_device_capability(device=None)
+    return major * 10 + minor
+
```
danieldk: Could you also move this one? The diff is still very botched. Probably a good idea to move `_is_{cuda,rocm}_platform` down as well to have them together.

Sorry for nitpicking, but I've found that clean diffs are really useful when bisecting.

PR #128, opened 2025-08-07
feat: add custom type for kernel modules

drbh:
This PR includes a subclass of the `ModuleType` type to mark dynamically loaded kernels as `_KernelModuleType`

```bash
uv run typed-kernels.py
```

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "accelerate",
#     "torch==2.7.0",
#     "transformers",
#     "kernels",
#     "ipdb",
# ]
#
# [tool.uv.sources]
# transformers = { git = "https://github.com/drbh/transformers.git", branch = "allow-kernel-rev" }
# kernels = { path = "../kernels", editable = true }
# ///
import torch

# Download optimized kernels from the Hugging Face hub
activation = get_kernel("kernels-community/activation")

print(torch)
print(activation)
print(activation.is_kernel)
print(type(activation))
```

output

```txt
Fetching 7 files: 100%|████████████████████████████████████████████████████| 7/7 [00:00<00:00, 119350.11it/s]
<module 'torch' from '/home/ubuntu/.cache/uv/environments-v2/kern-type-b885b03600690c7b/lib/python3.13/site-packages/torch/__init__.py'>
<class 'module'>
<kernel_module 'activation_68e410966316795f' from '/home/ubuntu/.cache/huggingface/hub/models--kernels-community--activation/snapshots/2fafa6a3a38ccb57a1a98419047cf7816ecbc071/build/torch27-cxx11-cu126-x86_64-linux/activation/__init__.py'>
True
<class 'kernel_module'>
```

### DIFF
--- a/src/kernels/utils.py
+++ b/src/kernels/utils.py
@@ -71,6 +71,17 @@ def universal_build_variant() -> str:
     return "torch-universal"
 
 
+# Custom module type to identify dynamically loaded kernel modules.
+# Using a subclass lets us distinguish these from regular imports.
+class _KernelModule(ModuleType):
+    """Marker class for modules loaded dynamically from a path."""
+    module_name: str
+    is_kernel: bool = True
+
+    def __repr__(self):
+        return f"<kernel_module '{self.module_name}' from '{self.__file__}'>"
+
+
 def import_from_path(module_name: str, file_path: Path) -> ModuleType:
     # We cannot use the module name as-is, after adding it to `sys.modules`,
     # it would also be used for other imports. So, we make a module name that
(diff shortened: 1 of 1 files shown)


src/kernels/utils.py
```diff
@@ -84,6 +102,9 @@ def import_from_path(module_name: str, file_path: Path) -> ModuleType:
     module = importlib.util.module_from_spec(spec)
     if module is None:
         raise ImportError(f"Cannot load module {module_name} from spec")
+    module.__class__ = _KernelModuleType
```
danieldk, line 105: Nice! I am a bit worried that there is some module graph processing code that directly matches on `ModuleType` and will break. Maybe it's safer to directly inject the attributes/methods for now without changing the class (we can always do it in the future without additional breakage)?

Maybe it's also worth storing the `LayerRepository`/`LocalRepository` besides the module name , since that provides full information about where the module comes from.

We also need to add to the spec that any `method_name` and `is_kernel` attributes are overridden and probably document this attributes.

PR #132, opened 2025-08-14
`cpu` is not (yet) a supported device type

### LINKED ISSUE #131: Bug: "cpu" kernels are not supported
Documentation [lists](https://github.com/huggingface/kernels/blob/1caa4c1393a184a735a361c57e8d22c7d90b854d/src/kernels/layer.py#L90) "cpu" as a kind of device for which kernels are supported
> type (`str`):
>     The device type (e.g., "cuda", "mps", "cpu").

However only three device types are [implemented](https://github.com/huggingface/kernels/blob/1caa4c1393a184a735a361c57e8d22c7d90b854d/src/kernels/layer.py#L120): "cuda", "rocm", and "mps".  All other device types raise a `ValueError`.

## To reproduce

```python
from kernels import kernelize, use_kernel_mapping
from my_private_repo import MyRmsNorm

with use_kernel_mapping({"RmsNorm": {"cpu": MyRmsNorm}}):
        model = kernelize(model)
```

danieldk:
Fixes #131.

### DIFF
--- a/src/kernels/layer.py
+++ b/src/kernels/layer.py
@@ -87,7 +87,7 @@ class Device:
 
     Args:
         type (`str`):
-            The device type (e.g., "cuda", "mps", "cpu").
+            The device type (e.g., "cuda", "mps", "rocm").
         properties ([`CUDAProperties`], *optional*):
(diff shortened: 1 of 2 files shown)


drbh (approved):
LGTM!

PR #138, opened 2025-09-08
[feat] add an uploading utility.

sayakpaul:
## TODOs

- [x] Add an example?
- [x] Add a test (staging test perhaps)?

One can easily run `kernels upload <path_to_build_dir> --repo_id=my-hub-username/my-kernel-name` to upload their cool kernels to the Hub.

### DIFF
--- a/src/kernels/cli.py
+++ b/src/kernels/cli.py
@@ -7,7 +7,7 @@ from pathlib import Path
 from kernels.compat import tomllib
 from kernels.lockfile import KernelLock, get_kernel_locks
 from kernels.utils import install_kernel, install_kernel_all_variants
-
+from huggingface_hub import create_repo, upload_folder
 from .doc import generate_readme_for_kernel
 from .wheel import build_variant_to_wheel
 
@@ -30,6 +30,24 @@ def main():
         help="Download all build variants of the kernel",
     )
     download_parser.set_defaults(func=download_kernels)
+
+    upload_parser = subparsers.add_parser("upload", help="Upload kernels to the Hub")
+    upload_parser.add_argument(
+        "kernel-dir",
+        type=Path,
+        help="Directory of the kernel build",
+    )
+    upload_parser.add_argument(
+        "repo-id",
+        type=Path,
+        help="Repository ID to use to upload to the Hugging Face Hub",
+    )
+    upload_parser.add_argument(
+        "private",
+        action="store_true",
+        help="If the repository should be private.",
+    )
+    upload_parser.set_defaults(func=upload_kernels)
 
     lock_parser = subparsers.add_parser("lock", help="Lock kernel revisions")
     lock_parser.add_argument(
@@ -151,6 +169,17 @@ def lock_kernels(args):
 
     with open(args.project_dir / "kernels.lock", "w") as f:
         json.dump(all_locks, f, cls=_JSONEncoder, indent=2)
+
+def upload_kernels(args):
+    repo_id = create_repo(
+        repo_id=args.repo_type, private=args.private, exist_ok=True
+    ).repo_id
+    upload_folder(
+        repo_id=repo_id, 
+        folder_path=args.kernel_dir,
+        commit_message="Uploaded from `kernels`."
+    )
+    print(f"✅ Kernel upload successful. Find the kernel in https://hf.co/{repo_id}.")
 
 
 class _JSONEncoder(json.JSONEncoder):
--- a/src/kernels/cli.py
+++ b/src/kernels/cli.py
@@ -188,7 +188,7 @@ def upload_kernels(args):
     delete_patterns: set[str] = set()
     for build_variant in build_dir.iterdir():
         if build_variant.is_dir():
-            delete_patterns.add(f"{build_variant}/**")
+            delete_patterns.add(f"{build_variant.name}/**")
 
     upload_folder(
         repo_id=repo_id,
--- a/src/kernels/cli.py
+++ b/src/kernels/cli.py
@@ -1,7 +1,6 @@
 import argparse
 import dataclasses
 import json
-import os
 import sys
 from pathlib import Path
 
@@ -12,7 +11,6 @@ from kernels.lockfile import KernelLock, get_kernel_locks
 from kernels.utils import install_kernel, install_kernel_all_variants
 
 from .doc import generate_readme_for_kernel
-from .utils import _get_filenames_from_a_repo
 from .wheel import build_variant_to_wheel
 
 
@@ -177,33 +175,29 @@ def lock_kernels(args):
 
 def upload_kernels(args):
     kernel_dir = Path(args.kernel_dir).resolve()
+    build_dir = kernel_dir / "build"
     if not kernel_dir.is_dir():
         raise ValueError(f"{kernel_dir} is not a directory")
+    if not build_dir.is_dir():
+        raise ValueError(f"Couldn't find `build` directory inside `kernel_dir`")
 
-    base_in_repo = kernel_dir.name
     repo_id = create_repo(
         repo_id=args.repo_id, private=args.private, exist_ok=True
     ).repo_id
 
-    repo_filenames = _get_filenames_from_a_repo(repo_id)
-    repo_base_filenames = [f for f in repo_filenames if f.startswith(f"{base_in_repo}/")]
-
-    build_dir = kernel_dir / "build"
-
     delete_patterns: set[str] = set()
     for build_variant in build_dir.iterdir():
         if build_variant.is_dir():
-            delete_patterns.add(f"{"build" / build_variant}/**")
+            delete_patterns.add(f"{build_variant}/**")
 
     upload_folder(
         repo_id=repo_id,
-        folder_path=str(kernel_dir),
-        path_in_repo=base_in_repo,
+        folder_path=build_dir,
+        path_in_repo="build",
         delete_patterns=list(delete_patterns),
         commit_message="Build uploaded using `kernels`.",
     )
     print(f"✅ Kernel upload successful. Find the kernel in https://hf.co/{repo_id}.")
-
 
 
 class _JSONEncoder(json.JSONEncoder):
--- a/src/kernels/cli.py
+++ b/src/kernels/cli.py
@@ -3,14 +3,14 @@ import dataclasses
 import json
 import sys
 from pathlib import Path
-
+import os
 from kernels.compat import tomllib
 from kernels.lockfile import KernelLock, get_kernel_locks
 from kernels.utils import install_kernel, install_kernel_all_variants
 from huggingface_hub import create_repo, upload_folder
 from .doc import generate_readme_for_kernel
 from .wheel import build_variant_to_wheel
-
+from .utils import _get_filenames_from_a_repo
 
 def main():
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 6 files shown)


src/kernels/cli.py
```diff
@@ -152,6 +170,17 @@ def lock_kernels(args):
     with open(args.project_dir / "kernels.lock", "w") as f:
         json.dump(all_locks, f, cls=_JSONEncoder, indent=2)
 
+def upload_kernels(args):
+    repo_id = create_repo(
+        repo_id=args.repo_type, private=args.private, exist_ok=True
+    ).repo_id
+    upload_folder(
+        repo_id=repo_id, 
+        folder_path=args.kernel_dir,
```
sayakpaul: Currently, this is assuming that the user knows which folder to use to upload to the Hub so that it works nicely when `get_kernel()` is called. But we could make it a bit more relaxed by trying to validate if the `kernel_dir` contains the expected files.

I think uttimately this will be combined with the `build-and-copy` command so that, users can start a build and automatically upload the build to the Hub. 

But LMK.

sayakpaul: We could also use `upload_large_folder()` in place of `upload_folder()`. This is beneficial especially if the folder contents are large and if we want to leverage parallel uploads.

danieldk: Yeah, uploads are typically large, a lot of kernels are hundreds of MiB. By default `.so` files (which are usually the large files in kernels) are not tracked by LFS. Do we need to make sure that this is the case, can we add `*.so` to be tracked through the `huggingface_hub` API?

For uploads, I think there are two scenarios:

1. The sources are hosted somewhere else like GitHub. The user only wants to upload the build artifacts to HF Hub. I think the subcommand added in this PR works really well for that usecase.
2. The sources are also hosted on HF Hub (like the `kernels-community` kernels). In this case we want to make a local Git commit and push that, so that the local and remote are in sync. (Or perhaps upload and then git pull.)

Though maybe the second case should be handled by `build-and-copy`, which should then check if the local repo has the hub as its main remote?

danieldk: Another thing:

* When e.g. `build/torch28-cxx11-cu128-x86_64-linux` already exists, we do not only want to update existing files, but we also want remove local files that do not exist locally anymore. E.g. suppose that before the upload, `build/torch28-cxx11-cu128-x86_64-linux` contains `_causal_conv1d_db03e28.abi3.so`, but the new build contains `_causal_conv1d_ecb27af.abi3.so`. If we do not remove `_causal_conv1d_db03e28.abi3.so`, then the client will unnecessarily download an unused old dynamic library (in this case 102MiB). So, I think we need to use `delete_patterns` to remove stale files.
* We cannot unconditionally remove everything from `build/`, because builds for different architectures (e.g. x86_64-linux and aarch64-darwin) are usually separate uploads. So we have to iterate over what build variants are uploaded now and delete old files from those directories.

Since this is easy to break we should probably have a test that generates a build directory with some randomly-named file and verify that after the upload the previous randomly-named file is gone. We have a test org, we could make a repo there for this test: https://huggingface.co/kernels-test

I think we should probably also upload only the `build` directory.

sayakpaul: > Yeah, uploads are typically large, a lot of kernels are hundreds of MiB. By default .so files (which are usually the large files in kernels) are not tracked by LFS. Do we need to make sure that this is the case, can we add *.so to be tracked through the huggingface_hub API?

Both `upload_folder()` and `upload_large_folder()` would work without any explicit specification to include `.so` files. 

> Though maybe the second case should be handled by build-and-copy, which should then check if the local repo has the hub as its main remote?

Ideally, `upload_folder()` or `upload_large_folder()` should consider these on behalf of the users. The users are supposed to issue the `upload` command once they feel okay with the local development, akin to doing `model.push_to_hub()` (which serializes the state dict and takes care of uploading the state dict and config to the Hub).

> When e.g. build/torch28-cxx11-cu128-x86_64-linux already exists, we do not only want to update existing files, but we also want remove local files that do not exist locally anymore. E.g. suppose that before the upload, build/torch28-cxx11-cu128-x86_64-linux contains _causal_conv1d_db03e28.abi3.so, but the new build contains _causal_conv1d_ecb27af.abi3.so. If we do not remove _causal_conv1d_db03e28.abi3.so, then the client will unnecessarily download an unused old dynamic library (in this case 102MiB). So, I think we need to use delete_patterns to remove stale files.

I am not sure if I fully understand. Is it a scenario where the local files are outdated w.r.t the remote files?

How do we determine `ignore_patterns` in this case? 👀 

> We cannot unconditionally remove everything from build/, because builds for different architectures (e.g. x86_64-linux and aarch64-darwin) are usually separate uploads. So we have to iterate over what build variants are uploaded now and delete old files from those directories.

I am not sure I understand this fully, either. Could you elaborate with an example?

> Since this is easy to break we should probably have a test that generates a build directory with some randomly-named file and verify that after the upload the previous randomly-named file is gone.

You mean the remote repo would have a set of files. With the test, 

* we would first generate a bunch randomly named files (maybe with similar structure as the ones existing in the remote repo)
* upload them with the `upload` command
* verify in the repo that the previous files are gone?

danieldk: > Both upload_folder() and upload_large_folder() would work without any explicit specification to include .so files.

I mean that we have to mark `.so` files to be tracked by git-lfs. That's not needed for regular model files, because the default `.gitattributes` file on the Hub already set `*.safetensors`, etc. als LFS-tracked. But not `*.so`.

> Ideally, `upload_folder()` or `upload_large_folder()` should consider these on behalf of the users. The users are supposed to issue the upload command once they feel okay with the local development, akin to doing `model.push_to_hub()` (which serializes the state dict and takes care of uploading the state dict and config to the Hub).

Ah, I see, so these functions also make the commit to the locally checked out repo?

> I am not sure if I fully understand. Is it a scenario where the local files are outdated w.r.t the remote files?

So, if we have on the server a build with `build/torch28-cxx11-cu128-x86_64-linux/causal_conv1d/_causal_conv1d_db03e28.abi3.so` and we are uploading a new, the name of the shared library will be different (e.g. `build/torch28-cxx11-cu128-x86_64-linux/causal_conv1d/_causal_conv1d_ecb27af.abi3.so`). We have to remove the old shared library `build/torch28-cxx11-cu128-x86_64-linux/causal_conv1d/_causal_conv1d_db03e28.abi3.so` (or any other stale files that are not used in the current build).

> I am not sure I understand this fully, either. Could you elaborate with an example?

If I have a kernel for x86_64-linux and aarch64-darwin, I would be uploading:

* `build/{torch27-cxx11-cu118-x86_64-linux,torch27-cxx11-cu126-x86_64-linux,torch27-cxx11-cu128-x86_64-linux,torch28-cxx11-cu126-x86_64-linux,torch28-cxx11-cu128-x86_64-linux,torch28-cxx11-cu129-x86_64-linux}` from a Linux machine.
* `build/{torch27-metal-aarch64-darwin,torch28-metal-aarch64-darwin}` from a Mac.

So we cannot fully clear out the `build` directory when uploading. We have to remove stale files from only the build variants that we are uploading.

> How do we determine ignore_patterns in this case? 👀

I think we need `delete_patterns`. I don't really know how they work. But maybe it would be enough to have a delete pattern for every build variant that we are uploading? Something like `build/<variant>/**`?

sayakpaul: > I mean that we have to mark .so files to be tracked by git-lfs. That's not needed for regular model files, because the default .gitattributes file on the Hub already set *.safetensors, etc. als LFS-tracked. But not *.so.

How are the so files currently uploaded on your end?

> Ah, I see, so these functions also make the commit to the locally checked out repo?

I don't think users should first `git clone` a repo and then run `upload`. `upload` is supposed to create a repo if it doesn't exist and then push the files as requested with an awareness of files that are not to be pushed. This is how it's done at least in other libraries such as `transformers`, `diffusers`, etc. Did you have anything else in mind? 

> So, if we have on the server a build with build/torch28-cxx11-cu128-x86_64-linux/causal_conv1d/_causal_conv1d_db03e28.abi3.so and we are uploading a new, the name of the shared library will be different (e.g. build/torch28-cxx11-cu128-x86_64-linux/causal_conv1d/_causal_conv1d_ecb27af.abi3.so). We have to remove the old shared library build/torch28-cxx11-cu128-x86_64-linux/causal_conv1d/_causal_conv1d_db03e28.abi3.so (or any other stale files that are not used in the current build).

Ah nice, thanks for explaining. Yes, I think we can do that. We will need a utility to determine the stale files given a build and the remote repository. Is there something (or something similar) available along those lines? 

> So we cannot fully clear out the build directory when uploading. We have to remove stale files from only the build variants that we are uploading.

Ah this makes sense, too. Thanks!

> I think we need delete_patterns. I don't really know how they work. But maybe it would be enough to have a delete pattern for every build variant that we are uploading? Something like build/<variant>/**?

I think we would need this IFF, the `variant` is also existing on the server. Otherwise, there's no need, right?

I think it might be helpful to have two hosts with identical kernels, but one would have updated build files while the other would have stale build files. This way, we can quickly iterate through the utility. WDYT?

If you have something handy, feel free to host them on the Hub. Otherwise, I can a look into creating those myself.

src/kernels/utils.py
```diff
@@ -487,3 +487,16 @@ def git_hash_object(data: bytes, object_type: str = "blob"):
...
+    try:
+        repo_info = model_info(repo_id=repo_id, files_metadata=True)
+        repo_siblings = repo_info.siblings
+        if repo_siblings is not None:
+            filenames = [f.rfilename for f in repo_siblings]
+            return filenames
+        else:
+            raise ValueError("No repo siblings found.")
+    except Exception as e:
+        logging.error(f"Error connecting to the Hub: {e}.")
```
danieldk: Shouldn't we let the exception propagate? The caller expects an iterable over strings.

src/kernels/utils.py
```diff
@@ -487,3 +487,16 @@ def git_hash_object(data: bytes, object_type: str = "blob"):
 
 def package_name_from_repo_id(repo_id: str) -> str:
     return repo_id.split("/")[-1].replace("-", "_")
+
+
+def _get_filenames_from_a_repo(repo_id: str):
```
danieldk:
```suggestion
def _get_filenames_from_a_repo(repo_id: str) -> List[str]:
```

src/kernels/utils.py
```diff
@@ -487,3 +487,16 @@ def git_hash_object(data: bytes, object_type: str = "blob"):
...
+    try:
+        repo_info = model_info(repo_id=repo_id, files_metadata=True)
+        repo_siblings = repo_info.siblings
+        if repo_siblings is not None:
+            filenames = [f.rfilename for f in repo_siblings]
+            return filenames
```
danieldk:
```suggestion
            return [f.rfilename for f in repo_siblings]
```

src/kernels/cli.py
```diff
@@ -153,6 +175,43 @@ def lock_kernels(args):
...
+    repo_base_filenames = [f for f in repo_filenames if f.startswith(f"{base_in_repo}/")]
+
+    delete_patterns: set[str] = set()
+    for folder in os.listdir(kernel_dir):
+        folder_path = kernel_dir / folder
+        if not folder_path.is_dir():
+            continue
+
+        if any(f.startswith(f"{base_in_repo}/{folder}/") for f in repo_base_filenames):
+            delete_patterns.add(f"{folder}/**")
```
danieldk: I was thinking more along these lines (untested):

```suggestion
    build_dir = kernel_dir / "build"

    delete_patterns: set[str] = set()
    for build_variant in build_dir.iterdir():
        if build_variant.is_dir():
            delete_patterns.add(f"{"build" / build_variant}/**")
```

I don't think we have to check what exists remotely, if the remote path `build/<build variant>` is in `delete_patterns` but doesn't exist, I think it simply has no effect?

src/kernels/cli.py
```diff
@@ -153,6 +175,43 @@ def lock_kernels(args):
...
+
+    upload_folder(
+        repo_id=repo_id,
+        folder_path=str(kernel_dir),
```
danieldk: I think we only want to upload the `build` directory?

src/kernels/cli.py
```diff
@@ -153,6 +175,43 @@ def lock_kernels(args):
...
+        path_in_repo=base_in_repo,
+        delete_patterns=list(delete_patterns),
+        commit_message="Uploaded from `kernels`.",
```
danieldk:
```suggestion
        commit_message="Build uploaded using `kernels`.",
```

src/kernels/utils.py
```diff
@@ -487,3 +487,16 @@ def git_hash_object(data: bytes, object_type: str = "blob"):
...
+    try:
+        repo_info = model_info(repo_id=repo_id, files_metadata=True)
+        repo_siblings = repo_info.siblings
+        if repo_siblings is not None:
+            return [f.rfilename for f in repo_siblings]
+        else:
+            raise ValueError("No repo siblings found.")
+    except Exception as e:
+        logging.error(f"Error connecting to the Hub: {e}.")
+        raise
```
danieldk: I think this can go to the test now, since it's only used there.

sayakpaul: Done!

danieldk (approved):
Looks great! Let's merge after we get the CI bits sorted out.

PR #157, opened 2025-09-24
Add a `Makefile` to run formatting one-shot

sayakpaul:
Run `make style` before pushing code. That's it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,9 @@ kernels = "kernels.cli:main"
 [project.entry-points."egg_info.writers"]
 "kernels.lock" = "kernels.lockfile:write_egg_lockfile"
 
+[tool.isort]
+profile = "black"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -71,4 +74,4 @@ line-length = 119
 # Ignored rules:
 # "E501" -> line length violation
 lint.ignore = ["E501"]
-lint.select = ["E", "F", "I", "W"]
```
sayakpaul, line 74: As we're relying on `isort` for formatting imports. We can also just use `ruff` for all of these.

sayakpaul: Example: https://github.com/huggingface/diffusers/blob/main/pyproject.toml

danieldk (approved):

danieldk:
Nice 🎉!

PR #158, opened 2025-09-25
Add the `kernels check` subcommand

danieldk:
This subcommand checks a given kernel. Currently it applies the same ABI checks as `kernel-abi-check` in `kernel-builder`.

### DIFF
--- a/src/kernels/check.py
+++ b/src/kernels/check.py
@@ -54,9 +54,9 @@ def check_kernel(
 
             # TODO: also check operating system
             if o.format() == BinaryFormat.ELF:
-                check_manylinux(o, manylinux, indent + 2)
+                has_issues |= check_manylinux(o, manylinux, indent + 2)
             elif o.format() == BinaryFormat.MACH_O:
-                check_macos(o, macos, indent + 2)
+                has_issues |= check_macos(o, macos, indent + 2)
 
     if has_issues:
         sys.exit(1)


nix/kernel-abi-check.nix
```diff
@@ -0,0 +1,27 @@
...
+    inherit pname version src sourceRoot;
+    hash = "sha256-+1jdbKsDKmG+bf0NEVYMv8t7Meuge1z2cgYfbdB9q8A=";
+  };
+
+  sourceRoot = "kernel_abi_check-${version}/bindings/python";
+
+  pyproject = true;
+
+  nativeBuildInputs = with rustPlatform; [ cargoSetupHook maturinBuildHook ];
+}
```
MekkCyber, line 27: Does this mean we can't install `kernels` without installing `kernel-abi-check` ?

src/kernels/check.py
```diff
@@ -0,0 +1,136 @@
...
+                revision=revision,
+            )
+        )
+        / "build"
+    )
+
+    has_issues = False
+    for variant_path in variants_path.iterdir():
+        if not variant_path.is_dir():
+            continue
```
MekkCyber, line 40: Shouldn't this be an anomaly ?

src/kernels/check.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+
+            o = ObjectFile(dylib_path)
+            has_issues |= check_abi3(o, python_abi, indent + 2)
+
+            # TODO: also check operating system
+            if o.format() == BinaryFormat.ELF:
+                check_manylinux(o, manylinux, indent + 2)
+            elif o.format() == BinaryFormat.MACH_O:
+                check_macos(o, macos, indent + 2)
```
MekkCyber: shouldn't we have : 
```suggestion
            if o.format() == BinaryFormat.ELF:
                has_issues |= check_manylinux(o, manylinux, indent + 2)
            elif o.format() == BinaryFormat.MACH_O:
                has_issues |= check_macos(o, macos, indent + 2)
```

MekkCyber (dismissed):
Woow very cool feature !

nix/kernel-abi-check.nix
danieldk: It's an optional dependency in `pyproject.toml`. The Nix is for the project's devshell, where we always want to have the package available.

src/kernels/check.py
danieldk: Great catch, fixed!

src/kernels/check.py
danieldk: Indeed, updated this to print an error.

nix/kernel-abi-check.nix
MekkCyber: perfect !

MekkCyber (approved):
LGTM

PR #161, opened 2025-10-08
feat: allow kernels to be uploaded to a revision

sayakpaul:


sayakpaul:
The failing tests seem to be unrelated?

src/kernels/check.py
```diff
@@ -1,18 +1,19 @@
-from pathlib import Path
 import sys
+from pathlib import Path
```
sayakpaul, line 2: Result of `make style`.

src/kernels/cli.py
```diff
@@ -69,6 +69,11 @@ def main():
         type=str,
         help="Repository ID to use to upload to the Hugging Face Hub",
     )
+    upload_parser.add_argument(
+        "--revision",
+        type=None,
+        help="If set, the upload will be made to a particular branch of the provided `repo_id`.",
+    )
```
danieldk, line 76: I wonder if we should be more explicit and call this `branch`. I find revision a bit confusing, since people use it to refer to commit SHAs, tags, etc. as well.

It's also not super-clear from the `huggingface_hub` docs what this means in `upload_folder`:

> The git revision to commit from. Defaults to the head of the "main" branch.

What happens if I'd put an actual git SHA here?

sayakpaul: Indeed. I changed it to `branch`. Maybe that's a better cognitive distinction but in theory, we can still pass all the acceptable revision variants (SHA, tag, etc.), though. But I don't see any clean way to differentiate amongst them, either.

danieldk: No me neither, but at least it's clear what should be passed.

PR #167, opened 2025-10-14
feat: allow get_kernel to log telemetry.

sayakpaul:
The values in `user_agent` are pretty non-identifiable. But LMK if you think otherwise. 

I have also made it so that the logging is configurable through an env var.

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -39,3 +39,13 @@ The approach of `forward`-replacement is the least invasive, because
 it preserves the original model graph. It is also reversible, since
 even though the `forward` of a layer _instance_ might be replaced,
 the corresponding class still has the original `forward`.
+
+## Misc
+
+### How can I disable kernel reporting in the user-agent?
+
+By default, we collect telemetry when a call to `get_kernel()` is made.
+This only includes the `kernels` version, `torch` version, and the build
+information for the kernel being requested.
+
+You can disable this by setting `export DISABLE_TELEMETRY=yes`.


src/kernels/utils.py
```diff
@@ -501,3 +515,28 @@ def git_hash_object(data: bytes, object_type: str = "blob"):
...
+    if user_agent is None:
+        user_agent = {
+            "kernels": __version__,
+            "torch": torch.__version__,
+            "build_variant": build_variant(),
+            "file_type": "kernel",
+        }
+
+    if disable_telemetry:
+        user_agent = None
```
danieldk: I'd swap around:

```suggestion
    if os.getenv("DISABLE_TELEMETRY", "false").upper() in ENV_VARS_TRUE_VALUES:
        return None

    if user_agent is None:
        user_agent = {
            "kernels": __version__,
            "torch": torch.__version__,
            "build_variant": build_variant(),
            "file_type": "kernel",
        }
```

src/kernels/utils.py
```diff
@@ -501,3 +515,28 @@ def git_hash_object(data: bytes, object_type: str = "blob"):
...
+    disable_telemetry = (
+        os.getenv("DISABLE_TELEMETRY", "false").upper() in ENV_VARS_TRUE_VALUES
+    )
+
+    if user_agent is None:
+        user_agent = {
+            "kernels": __version__,
+            "torch": torch.__version__,
+            "build_variant": build_variant(),
+            "file_type": "kernel",
```
danieldk, line 535: Looks non-revealing 👍, though I would be in favor of mentioning in the docs that we are collecting this information + how to switch it off.

sayakpaul: Sure! Which part of the docs should we put this in?

danieldk: Not sure. Maybe an item here https://huggingface.co/docs/kernels/faq with an additional question like _How can I disable kernel reporting in the user-agent?_

sayakpaul: Done.

PR #170, opened 2025-10-17
Add attributes to user_agent

MekkCyber:
Instead of expecting the user_agent from transformers/diffusers to have all the data, we just add it in `kernels` side.


sayakpaul (approved):
Nice. From a library, one could still maybe pass {"framework": "transformers", "framework_version": ...}.

danieldk (approved):

PR #180, opened 2025-11-07
misc(upload): specifically define what folder to allow uploading

mfuntowicz:
Especially as Windows build have a lot of other folders around `torch**`

### DIFF
--- a/src/kernels/cli.py
+++ b/src/kernels/cli.py
@@ -244,7 +244,7 @@ def upload_kernels(args):
         path_in_repo="build",
         delete_patterns=list(delete_patterns),
         commit_message="Build uploaded using `kernels`.",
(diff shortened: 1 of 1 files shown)


src/kernels/cli.py
```diff
@@ -244,6 +244,7 @@ def upload_kernels(args):
         path_in_repo="build",
         delete_patterns=list(delete_patterns),
         commit_message="Build uploaded using `kernels`.",
+        allow_patterns=["torch*"],
```
danieldk: Ah, yes, this was never an issue with nix builds.

Nitpick to make it as specific as possible (without spelling out all patterns):

```suggestion
        allow_patterns=["torch-*"],
```

MekkCyber (approved):
Sounds good!

PR #182, opened 2025-11-18
Validate kernel Python dependencies

danieldk:


src/kernels/utils.py
```diff
@@ -89,6 +90,14 @@ def universal_build_variant() -> str:
 
 
 def _import_from_path(module_name: str, variant_path: Path) -> ModuleType:
+    metadata_path = variant_path / "metadata.json"
+    if metadata_path.exists():
+        with open(metadata_path, "r") as f:
+            metadata = json.load(f)
+            deps = metadata.get("python-depends", [])
+            if deps is not None:
+                validate_dependencies(deps)
```
MekkCyber: small nit: `deps` will never be None here

tests/test_deps.py
```diff
@@ -0,0 +1,21 @@
...
+from kernels import get_kernel
+
+
+def test_python_deps():
+    must_raise = find_spec("nvidia_cutlass_dsl") is None
+    if must_raise:
+        with pytest.raises(
+            ImportError, match=r"Kernel requires dependency `nvidia-cutlass-dsl`"
+        ):
+            get_kernel("kernels-test/python-dep")
```
MekkCyber, line 14: an other nit : this path will not be tested since we installed `nvidia-cutlass-dsl` in the runner before running tests

danieldk: We run tests twice, once without `nvidia-cutlass-dsl` and once with.

src/kernels/utils.py
danieldk: Great catch! Fixed.

tests/test_deps.py
MekkCyber: ah sorry yes didn't see that

PR #187, opened 2025-11-21
Split up `kernels.layer` into several modules

danieldk:
The `layer` module became too large and there was some coupling between data types.


MekkCyber (approved):
I love the refactor ! Thanks for taking the time to do it

PR #188, opened 2025-11-25
Support functions as layers

danieldk:
This change adds two types of new functionality. First of all, it introduces the `(Locked|Local)?FuncRepo` classes these can be used to extend a layer with a kernel function. For instance, a layer like

```
@use_kernel_forward_from_hub("SiluAndMul")
class SiluAndMul(nn.Module):
        d = input.shape[-1] // 2
        return F.silu(input[..., :d]) * input[..., d:]
```

can now also be kernelized using a function `silu_and_mul` from the Hub:

```
with use_kernel_mapping({
    "SiluAndMul": {
        "cuda": FuncRepository(
            func_name="silu_and_mul",
        ),
    }
}):
    kernelize(...)
```

This makes it easier to kernelize pure layers (layers that do not use module state), since the Hub kernel does not have to provide a `layers` Python module with wrappers.

Secondly, we introduce a decorator `use_kernel_func_from_hub` that turns functions into layers that can be kernelized. For example:

```
@use_kernel_forward_from_hub("silu_and_mul")
def silu_and_mul(x: torch.Tensor) -> torch.Tensor:
    d = x.shape[-1] // 2
    return F.silu(x[..., :d]) * x[..., d:]
```

will implicitly create an instance of the following class:

```
class Func(nn.Module):
    # We add some magic to preserve the function's signature.
    def forward(self, *args, **kwargs):
        return silu_and_mul(*args, **kwargs)
```

Due to the `__call__` implementation of `nn.Module`, the instance still behaves as a function:

```
out = silu_and_mul(x)
```

However, when the function is used as a member of an `nn.Module`, it will be kernelized:

```
class FeedForward(nn.Module):
  def __init__(self, in_features: int, out_features: int):
      self.linear = nn.Linear(in_features, out_features)
      # Note: silu_and_mul is a Torch module.
      self.silu_and_mul = silu_and_mul

  def forward(self, x: torch.Tensor) -> torch.Tensor:
      return self.silu_and_mul(self.linear(x))
```

### DIFF
--- a/docs/source/api/layers.md
+++ b/docs/source/api/layers.md
@@ -56,7 +56,7 @@
 
 [[autodoc]] kernels.LocalLayerRepository
 
-### LocalFuncRepository
+### LockedFuncRepository
 
 [[autodoc]] kernels.LockedFuncRepository
 
--- a/docs/source/layers.md
+++ b/docs/source/layers.md
@@ -46,8 +46,8 @@ compatible with layers from the hub.
 ### Using a function as a layer
 
 Sometimes it can be useful to make a function extensible, for example
-because the function cannot be replaced by a layer. In such a case, you
-can annotate the function with the `use_kernel_func_from_hub` function:
+because the function cannot be replaced by a layer. In such cases, you
+can annotate the function with the `use_kernel_func_from_hub` decorator:
 
 ```python
 @use_kernel_func_from_hub("silu_and_mul")
@@ -57,11 +57,10 @@ def silu_and_mul(x: torch.Tensor) -> torch.Tensor:
 ```
 
 This will replace the function by an instantiated `torch.nn.Module`
-(singleton) that calls the function itself in its forward method. So, the
-'function module'
+(singleton) that calls the function itself in its forward method.
 
 **Note:** for kernelization to see the function, it must be a member of
-another `torch.nn.Module` that is past of the model. For example:
+another `torch.nn.Module` that is part of the model. For example:
 
 ```python
 class FeedForward(nn.Module):
@@ -188,7 +187,7 @@ with use_kernel_mapping(kernel_layer_mapping):
 This ensures that the mapping is not active anymore outside the
 `with`-scope.
 
-If the layer is stateless (it does not use member variables _or_ it was
+If the layer is stateless (it does not use member variables in its forward _or_ it was
 originally a function that was converted into a kernel layer with
 `use_kernel_func_from_hub`), it can also be mapped to a kernel function:
 


docs/source/api/layers.md
```diff
@@ -36,14 +40,26 @@
...
+[[autodoc]] kernels.LocalFuncRepository
+
 ### LocalLayerRepository
 
 [[autodoc]] kernels.LocalLayerRepository
 
+### LocalFuncRepository
+
+[[autodoc]] kernels.LockedFuncRepository
+
```
MekkCyber:
```suggestion
### LockedFuncRepository

[[autodoc]] kernels.LockedFuncRepository

```

docs/source/layers.md
```diff
@@ -43,6 +43,37 @@ replace_kernel_forward_from_hub(SiluAndMul, "SiluAndMul")
 it signifies that the maintainer intends to keep the `forward` signature
 compatible with layers from the hub.
 
+### Using a function as a layer
+
+Sometimes it can be useful to make a function extensible, for example
+because the function cannot be replaced by a layer. In such a case, you
+can annotate the function with the `use_kernel_func_from_hub` function:
```
MekkCyber:
```suggestion
because the function cannot be replaced by a layer. In such cases, you
can annotate the function with the `use_kernel_func_from_hub` decorator:
```

docs/source/layers.md
```diff
@@ -43,6 +43,37 @@ replace_kernel_forward_from_hub(SiluAndMul, "SiluAndMul")
...
+```python
+@use_kernel_func_from_hub("silu_and_mul")
+def silu_and_mul(x: torch.Tensor) -> torch.Tensor:
+    d = x.shape[-1] // 2
+    return F.silu(x[..., :d]) * x[..., d:]
+```
+
+This will replace the function by an instantiated `torch.nn.Module`
+(singleton) that calls the function itself in its forward method. So, the
+'function module'
```
MekkCyber: Is it truncated ?
```suggestion
This will replace the function by an instantiated `torch.nn.Module`
(singleton) that calls the function itself in its forward method.
```

docs/source/layers.md
```diff
@@ -43,6 +43,37 @@ replace_kernel_forward_from_hub(SiluAndMul, "SiluAndMul")
...
+
+**Note:** for kernelization to see the function, it must be a member of
+another `torch.nn.Module` that is past of the model. For example:
```
MekkCyber:
```suggestion
**Note:** for kernelization to see the function, it must be a member of
another `torch.nn.Module` that is part of the model. For example:
```

docs/source/layers.md
```diff
@@ -157,6 +188,21 @@ with use_kernel_mapping(kernel_layer_mapping):
 This ensures that the mapping is not active anymore outside the
 `with`-scope.
 
+If the layer is stateless (it does not use member variables _or_ it was
+originally a function that was converted into a kernel layer with
+`use_kernel_func_from_hub`), it can also be mapped to a kernel function:
```
MekkCyber: IIUC it's only the forward that should be stateless not other methods inside the layer

```suggestion
If the layer is stateless (it does not use member variables in its forward _or_ it was
originally a function that was converted into a kernel layer with
`use_kernel_func_from_hub`), it can also be mapped to a kernel function:
```

docs/source/layers.md
```diff
@@ -157,6 +188,21 @@ with use_kernel_mapping(kernel_layer_mapping):
...
+
+```python
+kernel_layer_mapping = {
+    "SiluAndMul": {
+        "cuda": FuncRepository(
+            repo_id="kernels-community/activation",
+            func_name="silu_and_mul",
+        ),
+    }
+}
```
MekkCyber, line 202: Awesome!!!

src/kernels/layer/func.py
```diff
@@ -0,0 +1,306 @@
...
+
+    # Use function signature with args prepended by self to support
+    # module validation.
+    func_sig = inspect.signature(func)
+    new_args = [Parameter("self", Parameter.POSITIONAL_OR_KEYWORD)]
+    new_args.extend(func_sig.parameters.values())
+    Func.forward.__signature__ = Signature(  # type: ignore[attr-defined]
+        parameters=new_args,
+        return_annotation=func_sig.return_annotation,
+    )
```
MekkCyber, line 304: So calling kernelize won’t cause any issues from replacing a method with a function, right?

src/kernels/layer/func.py
```diff
@@ -0,0 +1,306 @@
...
+
+    return _create_func_module(func)
+
+
+def _create_func_module(func: Callable) -> Type["nn.Module"]:
+    from torch import nn
+
+    class Func(nn.Module):
+        def forward(self, *args, **kwargs):
+            return func(*args, **kwargs)
```
MekkCyber, line 294: Great idea!

MekkCyber (commented):
Amazing work 🫡 !

I'm still a confused why we didn't need to change how kernelize works for the function mapping for example in here: https://github.com/huggingface/kernels/blob/main/src/kernels/layer/layer.py#L463.

src/kernels/layer/func.py
danieldk: Indeed, we need to fix the signature so that `validate_layer` works.

danieldk:
> I'm still a confused why we didn't need to change how kernelize works for the function mapping for example in here: https://github.com/huggingface/kernels/blob/main/src/kernels/layer/layer.py#L463.

Since the `FuncReposity` classes and `use_kernel_func_from_hub` wrap the function in an `nn.Module`, to `kernelize` there is no difference between a layer or a layer that started out as a function.

PR #190, opened 2025-12-03
Support noarch build variants

danieldk:
This change adds support for noarch build variants. So far we have used the universal variant for kernels that do not have any AoT-compiled code. However, the universal variant has two important issues:

1. A kernel without AoT-compiled might still be backend-specific. E.g. NVIDIA CuTe-based kernels are not universal in the sense that they don't work on non-NVIDIA GPUs.
2. We cannot specify dependencies per backend.

To solve these issues, we introduce the noarch variants to replace universal kernels. Noarch kernels have variants of the shape `torch-<backend>` (e.g. `torch-xpu`). This resolves the issues outlined.

This change introduces support for loading noarch kernels. In the future, we will start emitting deprecation warnings for universal kernels (to eventually remove support).

---

This change also fixes some small issues with CPU kernels that surfaced while making a test for noarch.

### DIFF
--- a/src/kernels/utils.py
+++ b/src/kernels/utils.py
@@ -317,7 +317,6 @@ def get_local_kernel(repo_path: Path, package_name: str) -> ModuleType:
     """
     # Presume we were given the top level path of the kernel repository.
     for base_path in [repo_path, repo_path / "build"]:
-        # Prefer the universal variant if it exists.
         for v in build_variants():
             variant_path = base_path / v
             if variant_path.exists():


src/kernels/utils.py
```diff
@@ -173,23 +194,22 @@ def _find_kernel_in_repo_path(
...
+            break
+
+    if variant is None:
         raise FileNotFoundError(
-            f"Kernel at path `{repo_path}` does not have one of build variants: {specific_variant}, {universal_variant}"
+            f"Kernel at path `{repo_path}` does not have one of build variants: {', '.join(variants)}"
         )
 
+    assert variant_path is not None
+
```
MekkCyber, line 212: not really needed but okay to keep it

src/kernels/utils.py
```diff
@@ -295,13 +315,10 @@ def get_local_kernel(repo_path: Path, package_name: str) -> ModuleType:
...
-    universal_variant = universal_build_variant()
-
     # Presume we were given the top level path of the kernel repository.
     for base_path in [repo_path, repo_path / "build"]:
         # Prefer the universal variant if it exists.
-        for v in [universal_variant, variant]:
+        for v in build_variants():
             variant_path = base_path / v
             if variant_path.exists():
                 return _import_from_path(package_name, variant_path)
```
MekkCyber: we don't really prefer the universal variant here, we need to update the comment

src/kernels/utils.py
danieldk: It is necessary to communicate to the type checker that at that point `variant_path` cannot be `None` anymore (mypy will fail without the assert).

src/kernels/utils.py
danieldk: Fixed. I am not even sure why we preferred the universal variant before 😅.

src/kernels/utils.py
MekkCyber: ah okay makes sense 👍

PR #194, opened 2025-12-15
Add support for backend dependencies

danieldk:
This change adds support for backend-specific dependencies. We use the dependencies from `kernel-builder` and validate upon loading that all dependencies are permitted dependencies.

### DIFF
--- a/update_python_depends.py
+++ b/update_python_depends.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+"""
+Download python_depends.json from the kernel-builder repository.
+"""
+
+import argparse
+import json
+from pathlib import Path
+from typing import Dict
+from urllib.request import Request, urlopen
+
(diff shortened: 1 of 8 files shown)


update_python_depends.py
```diff
@@ -0,0 +1,68 @@
...
+Download python_depends.json from the kernel-builder repository.
+"""
+
+import argparse
+import json
+from pathlib import Path
+from typing import Dict
+from urllib.request import Request, urlopen
+
+URL = "https://raw.githubusercontent.com/huggingface/kernel-builder/refs/heads/backend-deps/build2cmake/src/python_dependencies.json"
```
danieldk, line 12: This is a bit of a chicken and the egg-situation. But since this PR is simpler, I think it's a better entry point than the `kernel-builder` PR. Will update `s/backend-deps/main/` once the PR is merged on the `kernel-builder` side.

MekkCyber: This should be removed right ?

danieldk: Once the `kernel-builder` PR is merged, I'll update the URL.

PR #207, opened 2026-01-09
Initial benchmark command

drbh:
This is a work in progress PR to add a new benchmark command to the kernels cli tool. The idea is to enable a standard way to benchmark kernels

```bash
uvx \
--from git+https://github.com/huggingface/kernels.git@initial-benchmark-command \
--with torch \
--with numpy \
kernels benchmark kernels-community/activation # <- the expected command once merged
```

output

```bash
    Updated https://github.com/huggingface/kernels.git (daa75e4edfaccca487b9de9fb2b85b4cd052fd42)
      Built kernels @ git+https://github.com/huggingface/kernels.git@daa75e4edfaccca487b9de9fb2b85b4cd052fd42
Installed 45 packages in 81ms
Downloading kernels-community/activation@main...
Running benchmark.py...

┌────────────────┬──────────┬────────────┬────────────┬────────────┬────────────┬────────────┬──────────┬───────────┐
│ Benchmark      │ Workload │   Mean(ms) │    Std(ms) │    Min(ms) │    Max(ms) │    IQR(ms) │ Outliers │ Ref Match │
├────────────────┼──────────┼────────────┼────────────┼────────────┼────────────┼────────────┼──────────┼───────────┤
│ SiluWorkloads  │ medium   │     0.0109 │     0.0003 │     0.0105 │     0.0130 │     0.0002 │        3 │     ✓     │
│ SiluWorkloads  │ small    │     0.0051 │     0.0005 │     0.0049 │     0.0095 │     0.0002 │        3 │     ✓     │
│ SiluWorkloads2 │ medium   │     0.0298 │     0.0017 │     0.0282 │     0.0390 │     0.0010 │        5 │     ·     │
│ SiluWorkloads2 │ small    │     0.0061 │     0.0025 │     0.0055 │     0.0305 │     0.0001 │        7 │     ✓     │
└────────────────┴──────────┴────────────┴────────────┴────────────┴────────────┴────────────┴──────────┴───────────┘

Dry run - use --upload to submit results

```

[benchmark file](https://huggingface.co/kernels-community/activation/blob/main/benchmark.py)

### DIFF
--- a/src/kernels/check.py
+++ b/src/kernels/check.py
@@ -2,7 +2,7 @@ import sys
 from pathlib import Path
 
 from huggingface_hub import snapshot_download
-from kernel_abi_check import (
+from kernel_abi_check import (  # type: ignore[import-not-found]
     BinaryFormat,
     IncompatibleAbi3Symbol,
     IncompatibleMacOSVersion,


drbh:
### updated output

```bash
uvx \
--from git+https://github.com/huggingface/kernels.git@initial-benchmark-command \
--with torch \
--with numpy \
kernels benchmark kernels-community/activation # <- the expected command once merged
```

output

```bash
kernels benchmark kernels-community/activation # <- the expected command once merged
    Updated https://github.com/huggingface/kernels.git (a78fc45da9c22ac06411fb282570a8d7123fb412)
      Built kernels @ git+https://github.com/huggingface/kernels.git@a78fc45da9c22ac06411fb282570a8d7123fb412
Installed 45 packages in 73ms
Downloading kernels-community/activation@main...
Running benchmark.py...

┌────────────────┬──────────┬───────┬─────────┬────────────┬────────────┬────────────┬────────────┬────────────┬──────────┬────────────┬───────┐
│ Benchmark      │ Workload │     N │ Speedup │   Mean(ms) │    Std(ms) │    Min(ms) │    Max(ms) │    IQR(ms) │ Outliers │    Ref(ms) │ Match │
├────────────────┼──────────┼───────┼─────────┼────────────┼────────────┼────────────┼────────────┼────────────┼──────────┼────────────┼───────┤
│ SiluWorkloads  │ medium   │   100 │   6.93x │     0.0109 │     0.0002 │     0.0107 │     0.0123 │     0.0002 │        4 │     0.0755 │   ✓   │
│ SiluWorkloads  │ small    │   100 │  10.62x │     0.0053 │     0.0011 │     0.0050 │     0.0159 │     0.0002 │        6 │     0.0563 │   ✓   │
│ SiluWorkloads2 │ medium   │   100 │         │     0.0152 │     0.0018 │     0.0138 │     0.0257 │     0.0004 │       14 │            │   ·   │
│ SiluWorkloads2 │ small    │   100 │  10.60x │     0.0103 │     0.0006 │     0.0095 │     0.0136 │     0.0004 │        5 │     0.1092 │   ✓   │
└────────────────┴──────────┴───────┴─────────┴────────────┴────────────┴────────────┴────────────┴────────────┴──────────┴────────────┴───────┘

  medium: 6.93x faster (95% CI: 0.0109-0.0109ms vs ref 0.0755ms) ✓ significant
  small: 10.62x faster (95% CI: 0.0051-0.0055ms vs ref 0.0563ms) ✓ significant
  small: 10.60x faster (95% CI: 0.0102-0.0104ms vs ref 0.1092ms) ✓ significant

Kernel: 6ff8e1a  Benchmark: 9b68fca
Dry run - use --upload to submit results

```

drbh:
the following benches now run and use pre defined benchmarks that contain logic in this branch and pointer files in the kernel repos

```bash
uv run kernels benchmark kernels-community/activation
uv run kernels benchmark kernels-community/flash-attn2
uv run kernels benchmark kernels-community/flash-attn3
uv run kernels benchmark kernels-community/vllm-flash-attn3
```

https://huggingface.co/kernels-community/activation/blob/main/benchmark.py
https://huggingface.co/kernels-community/flash-attn2/blob/main/benchmark.py
https://huggingface.co/kernels-community/flash-attn3/blob/main/benchmark.py
https://huggingface.co/kernels-community/vllm-flash-attn3/blob/main/benchmark.py


## Future related work

This PR is the first step to enable more benchmarking features to kernels. Following PRs will continue to add
- [ ] documentation for writing custom benchmarks
- [ ] documentation for improving benches contained in this library
- [ ] explore pushing benchmark results to a shared location to aggregate community runs
- [ ] improve serialization format and benchmark identifiers for previous bullet point

pyproject.toml
```diff
@@ -13,9 +13,12 @@ readme = "README.md"
 requires-python = ">= 3.9"
 dependencies = [
   "huggingface_hub>=0.26.0,<2.0",
+  "numpy>=2.0.2",
   "packaging>=20.0",
   "pyyaml>=6",
+  "requests>=2.32.5",
   "tomli>=2.0; python_version<'3.11'",
+  "torch",
```
danieldk: Torch should definitely stay an optional dependency, since we might support other frameworks in the future. I think it's best if numpy and requests are also optional, since numpy is fairly large and requests has a lot of dependencies.

We could make a `benchmark` dependency group. Then the `benchmark` subcommand could conditionally import these dependencies and request users to `pip install kernels[benchmark]` if one of the modules cannot be imported.

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+
+try:
+    import torch
+
+    TORCH_AVAILABLE = True
+except ImportError:
+    torch = None  # type: ignore[assignment]
+    TORCH_AVAILABLE = False
+
+BENCHMARK_PATHS = ["benchmarks/bench.py", "benchmark.py"]
```
danieldk: I'd be in favor of a single path (probably `benchmarks/bench.py` to allow multiple files, or maybe even the `benchmark` directory and discover benchmarks pytest-style?).

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+                self.out = torch.empty(128, 512, device="cuda", dtype=torch.float16)
+
+            def benchmark_silu(self):
+                self.kernel.silu_and_mul(self.out, self.x)
+
+            def verify_silu(self) -> torch.Tensor:
+                # Return reference tensor; runner compares with self.out
+                return torch.nn.functional.silu(self.x[..., :512]) * self.x[..., 512:]
+
+    Run with: kernels benchmark <kernel-id> --script my_benchmark.py
```
danieldk:
```suggestion
    Run with: kernels benchmark <repo_id> --script my_benchmark.py
```

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+
+    return q1, q3, iqr, outliers
+
+
+class Benchmark:
+    """Base class for kernel benchmarks.
+
+    Subclass this to create a benchmark script with automatic timing,
+    verification, and reproducibility support. The kernel is loaded
+    automatically from the kernel_id specified in the CLI command.
```
danieldk:
```suggestion
    automatically from the repo_id specified in the CLI command.
```

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+    mean_ms: float
+    std_ms: float
+    min_ms: float
+    max_ms: float
+    iterations: int
+    q1_ms: float = 0.0  # 25th percentile
+    q3_ms: float = 0.0  # 75th percentile
+    iqr_ms: float = 0.0  # Interquartile range (Q3 - Q1)
+    outliers: int = 0  # Count of outliers (outside Q1-1.5*IQR to Q3+1.5*IQR)
+    verified: Optional[bool] = (
```
danieldk:
```suggestion
    verified: bool | None = (
```

Also for other `Optionals`, since #212 was merged.

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+        None  # None = no verify fn, True = passed, False = failed
+    )
+    ref_mean_ms: Optional[float] = None  # Reference implementation mean time
+
+
+@dataclass
+class MachineInfo:
+    gpu: str
+    cuda_version: str
```
danieldk: Maybe

```suggestion
    backend: str
    backend_version: Version
```

? Since we also want to support Metal, XPU, etc.

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+        )
+        if result.returncode == 0:
+            data = json.loads(result.stdout)
+            displays = data.get("SPDisplaysDataType", [])
+            if displays:
+                gpu_name = displays[0].get("sppci_model", "")
+                if gpu_name:
+                    return gpu_name
+    except (subprocess.TimeoutExpired, FileNotFoundError, json.JSONDecodeError):
+        pass
```
danieldk: `system_profiler` is quite heavy. I also wonder if it is worth it, since isn't this the CPU on Apple Silicon Macs? (at least on mine, this is also `Apple M5`).

danieldk: `sppci_cores` seems more interesting, but I wonder if there is a nicer way to get it than through `system_profiler`.

(Maybe we can get the number of GPU cores using ctypes?)

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+    # Suppress progress bars for cleaner output (files are often cached)
+    disable_progress_bars()
+
+    if local_dir:
+        repo_path = Path(local_dir)
+        if not repo_path.exists():
+            print(f"Error: Local directory not found: {local_dir}", file=sys.stderr)
+            sys.exit(1)
+        print(f"Using local directory: {repo_path}", file=sys.stderr)
+        kernel_sha = _get_local_commit_sha(repo_path)
```
danieldk: We should get the sha from the kernel's op name, since the commit SHA may not be correct (e.g. if you are benchmarking x86_64-linux, but you uploaded a aarch64-linux build since then).

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+    else:
+        print(f"Downloading {repo_id}@{revision}...", file=sys.stderr)
+        repo_path = Path(snapshot_download(repo_id=repo_id, revision=revision))
+        kernel_sha = get_kernel_commit_sha(repo_id, revision)
```
danieldk: Same here.

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+
+    if system == "Darwin":
+        cpu = _get_macos_chip() or cpu
+        gpu = _get_macos_gpu() or gpu
+
+    if TORCH_AVAILABLE:
+        pytorch_version = torch.__version__
+        if torch.cuda.is_available():
+            gpu = torch.cuda.get_device_name(0)
+            cuda_version = torch.version.cuda or "N/A"
```
danieldk: It's worth adding XPU/ROCm as well.

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+    if sha is None:
+        raise ValueError(f"Could not get commit SHA for {repo_id}@{revision}")
+    return sha
+
+
+def run_benchmark_class(
+    benchmark_cls: type[Benchmark],
+    iterations: int,
+    warmup: int,
+    kernel_id: str,
```
danieldk: repo_id

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+        if name.startswith("benchmark_") and callable(getattr(benchmark_cls, name))
+    ]
+
+    if not benchmark_methods:
+        raise RuntimeError(f"No benchmark_* methods found in {benchmark_cls.__name__}")
+
+    # Load kernel once for all workloads
+    from kernels import get_kernel
+
+    kernel = get_kernel(kernel_id)
```
danieldk: Shouldn't the revision be passed?

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+        verified: Optional[bool] = None
+        ref_mean_ms: Optional[float] = None
+        if verify_fn is not None:
+            benchmark_fn()  # Populate output
+            torch.cuda.synchronize()
+
+            # Warmup the verify/reference computation
+            for _ in range(warmup):
+                verify_fn()
+                torch.cuda.synchronize()
```
danieldk: Does this work on non-CUDA?

src/kernels/benchmark.py
```diff
@@ -0,0 +1,810 @@
...
+        "Add benchmarks/bench.py or use --central for well-known benchmarks",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+
+
+def _run_benchmark_script_subprocess(
+    script_path: Path, iterations: int, warmup: int, cwd: Path
+) -> dict[str, TimingResults]:
+    """Run a legacy benchmark script as a subprocess."""
```
danieldk: We have legacy benchmark scripts? If so, maybe we should just remove support since it was never an advertised feature?

src/kernels/check.py
```diff
@@ -2,7 +2,7 @@
 from pathlib import Path
 
 from huggingface_hub import snapshot_download
-from kernel_abi_check import (
+from kernel_abi_check import (  # type: ignore[import-not-found]
```
danieldk: Do we need this? mypy isn't failing currently

src/kernels/cli.py
```diff
@@ -113,6 +113,63 @@ def main():
...
+        "repo_id",
+        type=str,
+        help="Kernel repo ID (e.g., kernels-community/activation)",
+    )
+    benchmark_parser.add_argument(
+        "--script",
+        type=str,
+        default=None,
+        help="Custom benchmark script path (overrides auto-discovery)",
+    )
```
danieldk: I think we shouldn't allow this, just let people put it in a standard location.

src/kernels/cli.py
```diff
@@ -113,6 +113,63 @@ def main():
...
+    benchmark_parser.add_argument(
+        "--json",
+        action="store_true",
+        help="Print full JSON results to stdout (in addition to table)",
+    )
+    benchmark_parser.add_argument("--iterations", type=int, default=100)
+    benchmark_parser.add_argument("--warmup", type=int, default=10)
+    benchmark_parser.add_argument("--api-url", type=str, default=None)
+    benchmark_parser.add_argument("--token", type=str, default=None)
```
danieldk: Shouldn't people use `HF_TOKEN`?

pyproject.toml
drbh: really good point, updated to prefer a benchmark dependency group as suggested in latest. thanks for the recommendation!

src/kernels/benchmark.py
drbh: this makes sense, I've reduced to a single directory and look for files with the pattern `"benchmark*.py"`

src/kernels/benchmark.py
drbh: ah yes, agreed this should be repo not kernel. updated in latest

src/kernels/benchmark.py
drbh: updated in latest

src/kernels/benchmark.py
drbh: great point, updated to prefer native types in latest

src/kernels/benchmark.py
drbh: yep good catch, updated in latest

src/kernels/benchmark.py
drbh: good point, this slipped after our team conversation. fixed in the latest!

src/kernels/benchmark.py
drbh: *note the fix was applied here since the `_get_local_commit_sha` was removed recent commits when the local dir was removed

src/kernels/benchmark.py
drbh: good catch, updated in latest

src/kernels/benchmark.py
drbh: updated in latest

src/kernels/benchmark.py
drbh: ah yea totally, fixed in latest

src/kernels/benchmark.py
drbh: good catch updated to call the appropriate synchronize based on the system

src/kernels/benchmark.py
drbh: ah this should have been removed earlier, removed and cleaned up in latest

src/kernels/check.py
drbh: oh whoops that's an artifact from adding the ty extension in zed. removed in latest changes

src/kernels/cli.py
drbh: agreed removed in latest, this is and `--local-dir`, `--force`, `--api-url`, `--token` and `--upload` are all artifacts from debugging locally and having an easy way to test with local scripts and push to a local endpoint. 

I've removed them all for now and will revisit when the upload/endpoint path is better defined

src/kernels/cli.py
drbh: removed in latest

src/kernels/benchmark.py
drbh: very good point, the system_profiler is very slow and addd a much faster and better implementation via `iokit` and `ctypes`. 

example output for `collect_machine_info` is now

```py
MachineInfo(gpu='Apple M3 Max', backend='MPS', pytorch_version='2.9.1', os='Darwin 25.2.0', cpu='Apple M3 Max', gpu_cores=30)
```

danieldk: Nice!!! 🚀

danieldk (commented):
So nice to have this!

PR #208, opened 2026-01-12
Fix too-aggressive layer caching

danieldk:
We did not store the kernel revision in the `LockedLayerRepository` and `LockedFuncRepository`. This results in cache hits when loading a kernel twice with different revisions. Fix this by storing the revision in instances as well and using them in `__eq__` and `__hash__`.


drbh (approved):
nice lgtm!

PR #209, opened 2026-01-13
Add support for kernel versions

danieldk:
This change adds support for kernel versions branches as implemented by `kernel-builder`:

* `kernels upload` will upload the branch corresponding to the kernel's version.
* `get_kernel`, `LayerRepository`, etc. accept version branches through the `version` argument.
* Kernel locking (e.g. `kernels lock`) support reading version branches from `pyproject.yaml` and correctly locking them.

All the kernel functions and layers already supported a `version` argument for version specifiers (based on tags). These are now overloaded -- if `version` is a `str`, it is parsed as a version specifier. If it is an `int`, it is used as a version branch.

This overloading is used to gradually move to version branches, deprecating version specifiers. For now, using version specifiers will emit a warning.

### DIFF
--- a/kernels/src/kernels/versions_cli.py
+++ b/kernels/src/kernels/versions_cli.py
@@ -16,7 +16,7 @@ def print_kernel_versions(repo_id: str):
         print(f"Repository does not support kernel versions: {repo_id}")
         return
 
-    for version, ref in versions:
+    for version, ref in sorted(versions, key=lambda x: x[0]):
         print(f"Version {version}: ", end="")
         variants = [
             f"{variant} ✅" if variant in compatible_variants else f"{variant}"


docs/source/kernel-requirements.md
```diff
@@ -40,21 +40,33 @@ must be available for that combination.
 ## Kernel metadata
 
 The build variant directory can optionally contain a `metadata.json` file.
-Currently the only purpose of the metadata is to specify the kernel python dependencies, for example:
+Currently the metadata specifies the kernel's versin and Python dependencies,
```
MekkCyber:
```suggestion
Currently the metadata specifies the kernel's version and Python dependencies,
```

docs/source/kernel-requirements.md
```diff
@@ -40,21 +40,33 @@ must be available for that combination.
...
+version are stored in a branch of the form `v<version>`. Each build
+variant will also have the kernel version in `metadata.json`.
 
-We recommend using [semver](https://semver.org/) to version kernels.
+The version **must** be bumped in the following cases:
+
+- The kernel API is changed in an incompatible way.
+- The API is extended in a compatible way, but not all build variants
+  receive the extension (e.g. because they are for older Torch versions
+  that are not supported by `kernel-builder` anymore).
```
MekkCyber, line 116: niiice!

docs/source/migration.md
```diff
@@ -0,0 +1,68 @@
...
+without specifying a version. This is deprecated in kernels 0.12
+and will become an error in kernels 0.14. Instead, use of a kernel
+should always specify a version (except for local kernels).
+
+Kernels only use a major version. The kernel maintainer is responsible
+for never breaking a kernel within a major version and should bump up
+the major version if the kernel API changes and/or when support for
+older Torch versions is removed.
+
+You can find the versions that are supported by a kernel using the
```
MekkCyber, line 17: awesome!

kernels/src/kernels/_versions.py
```diff
@@ -18,33 +38,55 @@ def _get_available_versions(repo_id: str) -> dict[Version, GitRefInfo]:
...
     elif version is not None:
-        revision = resolve_version_spec_as_ref(repo_id, version).target_commit
-    assert revision is not None
-    return revision
+        return resolve_version_spec_as_ref(repo_id, version).target_commit
+
+    warnings.warn(
+        "Future versions of `kernels` will require specifying a kernel version or revision."
+        "See: https://huggingface.co/docs/kernels/migration"
+    )
```
MekkCyber, line 90:
```suggestion
    warnings.warn(
        "Future versions of `kernels` (>=0.14) will require specifying a kernel version or revision."
        "See: https://huggingface.co/docs/kernels/migration"
    )
```

MekkCyber (approved):

kernels/src/kernels/variants.py
```diff
@@ -0,0 +1,3 @@
+import re
+
+BUILD_VARIANT_REGEX = re.compile(r"^(torch\d\d+.*|torch-(cpu|cuda|metal|rocm|xpu))")
```
drbh: the regex is a bit more permissive now, previously we limited the torch version to only have numbers like `torch123` but now it looks like we'd match `torch123abc` and etc. do we need to update the regex?

previous regex
```
BUILD_VARIANT_REGEX = re.compile(r"^(torch\d+\d+|torch-(cpu|cuda|metal|rocm|xpu))")
```

**update
thinking about this more I actually think this may introduce an issue when calling `match.group(1)` to get the torch version in `_get_build_variants` where the `+.*` may cause the whole string to be matched (not just the torch section)

```python
import re

old = re.compile(r"^(torch\d+\d+|torch-(cpu|cuda|metal|rocm|xpu))")
new = re.compile(r"^(torch\d\d+.*|torch-(cpu|cuda|metal|rocm|xpu))")

s = "torch27-extra-stuff"

print("old:", old.match(s).group(1))
print("new:", new.match(s).group(1))
# old: torch27
# new: torch27-extra-stuff
```

kernels/src/kernels/versions_cli.py
```diff
@@ -0,0 +1,38 @@
...
+def print_kernel_versions(repo_id: str):
+    api = HfApi()
+    compatible_variants = set(build_variants())
+
+    versions = _get_available_versions(repo_id).items()
+    if not versions:
+        print(f"Repository does not support kernel versions: {repo_id}")
+        return
+
+    for version, ref in _get_available_versions(repo_id).items():
```
drbh: maybe we can avoid calling `_get_available_versions` twice since it may make apis calls? I think we can just reuse the `versions` variable above?

```suggestion
    versions = _get_available_versions(repo_id).items()
    if not versions:
        print(f"Repository does not support kernel versions: {repo_id}")
        return

    for version, ref in versions:
```

kernels/src/kernels/_versions.py
```diff
@@ -18,33 +38,55 @@ def _get_available_versions(repo_id: str) -> dict[Version, GitRefInfo]:
...
+        ref = versions.get(version_spec, None)
+        if ref is None:
+            raise ValueError(
+                f"Version {version_spec} not found, available versions: {', '.join(sorted(str(v) for v in versions.keys()))}"
+            )
+        return ref
+    else:
+        warnings.warn(
+            """Version specifiers are deprecated, support will be removed in a future `kernels` version.
+            "For more information on migrating to versions, see: https://huggingface.co/docs/kernels/migration"""
```
drbh: just an extra `"`

```suggestion
            For more information on migrating to versions, see: https://huggingface.co/docs/kernels/migration"""
```

kernels/src/kernels/versions_cli.py
drbh: maybe we can avoid calling `_get_available_versions` twice since it may make apis calls? I think we can just reuse the `versions` variable above?

```suggestion
    versions = _get_available_versions(repo_id).items()
    if not versions:
        print(f"Repository does not support kernel versions: {repo_id}")
        return

    for version, ref in versions:
```

**update

also not sure if we should sort here so the versions are in a asc/desc order?

drbh (approved):
looks good just small nits!

kernels/src/kernels/_versions.py
danieldk: Thanks! Fixed!

kernels/src/kernels/versions_cli.py
danieldk: Fixed!

kernels/src/kernels/variants.py
danieldk: I don't think it's more permissive, since we only use the regexp for `match`. The group is a great catch, seems like that might cause issues. I have reverted to the old regex and used the path to get the full variant name in `kernels versions`.

In the future I want to build a proper parser for build variants, so that we can also apply sort orders, etc.

kernels/src/kernels/versions_cli.py
danieldk: Great idea, sorting by version now (ascending, since then the latest version is what people will see in the terminal).

PR #211, opened 2026-01-14
Update docs/source/layers.md with LocalFuncRepository example

onel:
I wanted to update the guides with more example of usage for `LocalFuncRepository` and `LockedFuncRepository`

### DIFF
--- a/docs/source/locking.md
+++ b/docs/source/locking.md
@@ -44,15 +44,11 @@ kernel_layer_mapping = {
 register_kernel_mapping(kernel_layer_mapping)
 ```
 
-### Locked kernel functions
-
-Locked functions can also be used with `LockedFuncRepository`. This is useful for stateless
-layers or functions that were converted with `use_kernel_func_from_hub`:
+Similarly, you can use the `LockedFuncRepository` class to lock kernel function
+versions:
 
 ```python
-from kernels import LockedFuncRepository, use_kernel_mapping
-
-kernel_func_mapping = {
+kernel_layer_mapping = {
     "silu_and_mul": {
         "cuda": LockedFuncRepository(
             repo_id="kernels-community/activation",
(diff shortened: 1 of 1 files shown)


docs/source/layers.md
```diff
@@ -348,7 +348,7 @@ a kernel to a range of ROCm capabilities.
 
 ### Loading from a local repository for testing
 
-The `LocalLayerRepository` class is provided to load a repository from
+The `LocalLayerRepository` class is provided to load a layer repository from
```
MekkCyber: we can have repositories that contain layers, and functions at the same time, so there is not such thing as layer repository and function repository

docs/source/layers.md
```diff
@@ -366,3 +366,28 @@ with use_kernel_mapping(
...
+    },
+    inherit_mapping=False,
+):
+    kernelize(model, mode=Mode.INFERENCE)
+```
+
+The parameters are:
+- `repo_path`: Path to the local kernel repository directory
+- `package_name`: The Python package name of the kernel
+- `func_name`: The name of the function within the kernel repository
```
MekkCyber: maybe we don't need to add this, it's already very explicit within the example, wdyt ?

docs/source/locking.md
```diff
@@ -44,6 +44,28 @@ kernel_layer_mapping = {
 register_kernel_mapping(kernel_layer_mapping)
 ```
 
+### Locked kernel functions
+
+Locked functions can also be used with `LockedFuncRepository`. This is useful for stateless
+layers or functions that were converted with `use_kernel_func_from_hub`:
```
MekkCyber: not sure this is accurate since all layers are stateless, maybe we can just say 
```suggestion
Locking is also supported for functions, just like for layers, through the use of `LockedFuncRepository`:
```

MekkCyber (approved):
Yes makes more sense! thanks for working on this

PR #217, opened 2026-01-22
Merge Nix from kernel-builder with kernels/build2cmake

danieldk:
Merge and clean up the Nix from kernel-builder, kernels, and build2cmake. The cleanup mostly consists of moving some builder bits out of `flake.nix`. The devshells also work better, because they use Torch with CUDA (Linux) or Metal (macOS) support.


.gitignore
```diff
@@ -16,7 +16,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
```
danieldk, line 19: This was interacting with the `nix/lib` folder, blocking editor search and `git add`.

PR #220, opened 2026-01-22
Benchmark command fixes

drbh:
This PR 

- adds a small fix to the `get_kernel_sha_from_build_name` as the name parsing was not working with arbitrary kernels.
- hotfixes the `_synchronize` that had a typo from the previous changes
- adds layer norm benchmarking function for [layer_norm](https://github.com/huggingface/kernels-community/tree/main/layer_norm) 
- adds a `--benchmarks` and `--benchmarks-only` flag to the `upload` subcommand to upload the benchmark directory

### DIFF
--- a/kernels/src/kernels/benchmark.py
+++ b/kernels/src/kernels/benchmark.py
@@ -88,11 +88,11 @@ class Benchmark:
 
     seed: int | None = None  # Optional: seed for reproducibility
 
-    def __init__(self):
-        self.kernel = None
+    def __init__(self) -> None:
+        self.kernel: Any = None
         self.out: Any = None  # Output tensor, set by setup methods
 
-    def setup(self):
+    def setup(self) -> None:
         """Override to set up tensors as instance attributes."""
         pass
 
--- a/kernels/src/kernels/metadata.py
+++ b/kernels/src/kernels/metadata.py
@@ -1,7 +1,8 @@
 import json
 from dataclasses import dataclass
 from pathlib import Path
-import tomllib
+
+from kernels.compat import tomllib
 
 
 @dataclass


kernels/src/kernels/cli.py
```diff
@@ -202,6 +214,28 @@ def upload_kernels(args):
...
+    # benchmarks directory upload (doesn't require build variants)
+    if args.benchmarks or args.benchmarks_only:
+        upload_folder(
+            repo_id=repo_id,
+            folder_path=kernel_dir / "benchmarks",
+            revision=args.branch,
+            path_in_repo="benchmarks",
+            delete_patterns=["benchmark*.py"],
+            commit_message="Benchmarks uploaded using `kernels`.",
+            allow_patterns=["benchmark*.py"],
```
MekkCyber: small nit: delete_patterns are the same as allow_patterns, is this expected

drbh: ah yea this is strange looking syntax but I believe its correct, since I think this is how to specify which files to overwrite/remove if they don't exist in the latest local copy of the dir.

kernels/src/kernels/cli.py
```diff
@@ -202,6 +214,28 @@ def upload_kernels(args):
     # Resolve `kernel_dir` to be uploaded.
     kernel_dir = Path(args.kernel_dir).resolve()
 
+    repo_id = create_repo(
+        repo_id=args.repo_id, private=args.private, exist_ok=True
+    ).repo_id
+
+    if args.branch is not None:
+        create_branch(repo_id=repo_id, branch=args.branch, exist_ok=True)
+
```
danieldk: With the versions PR merged, this needs a sync against the version handling.

kernels/src/kernels/cli.py
MekkCyber: makes sense! thanks for clarifying

kernels/src/kernels/upload.py
```diff
@@ -7,10 +7,39 @@
...
 
 def upload_kernels_dir(
-    kernel_dir: Path, *, repo_id: str, branch: str | None, private: bool
+    kernel_dir: Path,
+    *,
+    repo_id: str,
+    branch: str | None,
+    private: bool,
+    benchmarks: bool = False,
+    benchmarks_only: bool = False,
```
danieldk: I think we always want to pass these through explicitly? (So no kwarg default to ensure that they are always passed.)

drbh: good point, updated in latest commits

kernels/src/kernels/upload.py
```diff
@@ -7,10 +7,46 @@
...
+            path_in_repo="benchmarks",
+            delete_patterns=["benchmark*.py"],
+            commit_message="Benchmarks uploaded using `kernels`.",
+            allow_patterns=["benchmark*.py"],
+        )
+
+        print(
+            f"✅ Benchmarks upload successful. Find the kernel in: https://hf.co/{repo_id}"
+        )
+        return  # Exit if only benchmarks are to be uploaded
```
danieldk, line 48: I think we should completely remove the `benchmark_only` case:

* It would only be useful in the case one also has built kernels, but only wants to upload the benchmark, which seems very rare. (And if no kernels were built, there is no additional cost, since existing variants will not be uploaded again).
* It relies on the existence of `build.toml`, but this file may not be present (e.g. if someone has their own build infra).
* I makes the upload logic quite a bit more complex.

In the very rare case where the first point happens, someone could use the huggingface CLI to upload.

PR #222, opened 2026-01-22
Add back README to the Python package

danieldk:


kernels/README.md
```diff
@@ -0,0 +1,50 @@
+# kernels
+
```
danieldk, line 2: I didn't put the logo image here, since it takes up a lot of valuable space on PyPI and we it's slow to load (as long as we don't have an SVG).

PR #240, opened 2026-01-26
update relative paths and readme cleanups

drbh:
This pr simply updates the relative paths and urls in various docs, it also updates the repos landing page readme to use the logo that was previously on the kernel-builder readme


builder/README.md
```diff
@@ -70,14 +66,14 @@ See [dockerfiles/README.md](./dockerfiles/README.md) for more options, including
 
 # 📚 Documentation
 
-- [Writing Hub kernels](./docs/writing-kernels.md)
-- [Building kernels with Nix](./docs/nix.md)
+- [Writing Hub kernels](../docs/source/builder/writing-kernels.md)
```
danieldk, line 69: Can be a separate PR, but I think in the future we could link these out to the hf.co documentation pages, since it renders more nicely?

drbh: ah yea great point, i'll open a small PR with that improvement soon

PR #244, opened 2026-01-27
Improve benchmark command

drbh:
This pr includes two changes

1. adds `device` to the `Benchmark` class so the benchmark runner can pass the detected device into the benchmark scripts. It also allows overriding this value within a benchmark script similar to `seed`
2. adds `tabulate` as a optional dependency for benchmarks to avoid the hacky box drawing logic.

### DIFF
--- a/kernels/src/kernels/benchmark.py
+++ b/kernels/src/kernels/benchmark.py
@@ -14,6 +14,8 @@ from typing import Any
 from huggingface_hub import get_token, snapshot_download
 from huggingface_hub.utils import disable_progress_bars
 
+from kernels.utils import backend
+
 MISSING_DEPS: list[str] = []
 
 try:
@@ -405,21 +407,10 @@ def _get_macos_gpu() -> tuple[str | None, int | None]:
         return None, None
 
 
-def get_device() -> str:
-    if TORCH_AVAILABLE:
-        if torch.cuda.is_available():
-            return "cuda"
-        elif hasattr(torch, "xpu") and torch.xpu.is_available():
-            return "xpu"
-        elif hasattr(torch.backends, "mps") and torch.backends.mps.is_available():
-            return "mps"
-    return "cpu"
-
-
 def collect_machine_info() -> MachineInfo:
     gpu = "N/A"
     gpu_cores = None
-    backend = "N/A"
+    backend_type = "N/A"
     pytorch_version = "N/A"
     system = platform.system()
     os_info = f"{system} {platform.release()}"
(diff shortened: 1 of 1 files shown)


kernels/src/kernels/benchmark.py
```diff
@@ -71,10 +77,11 @@ class Benchmark:
     Example:
         class MyBenchmark(Benchmark):
             seed = 42
+            device = "cpu"
```
danieldk: Why does the benchmark implementer have to put a device here? Don't we always want the benchmarker to set the device based on what is available?

kernels/src/kernels/benchmark.py
```diff
@@ -468,12 +445,24 @@ def _synchronize() -> None:
         torch.mps.synchronize()
 
 
+def _device_from_backend(backend: str) -> str:
+    backend_lower = backend.lower()
```
danieldk: Shouldn't we rely on the caller to provide a proper non-capitalized backend name?

kernels/src/kernels/benchmark.py
```diff
@@ -468,12 +445,24 @@ def _synchronize() -> None:
...
+    if backend_lower.startswith("cuda") or backend_lower.startswith("rocm"):
```
danieldk: Why `startswith`?

kernels/src/kernels/benchmark.py
drbh: originally the idea was to allow the benchmark implementer a way to override the device if they wanted to force a different backend, but since this differs from how the kernel is pulled I realized it doesn't make sense to allow this and I've removed in the latest changes

kernels/src/kernels/benchmark.py
drbh: this was added because I was reusing the `machine_info` that checked the device type and would make an capitalized string of the backend, however this is also removed in the latest

kernels/src/kernels/benchmark.py
drbh: related to the comment above, the backend was a human readable string that would include the cuda or hip version in the string. the latest changes improve by using a `get_device` function that returns a simple lowercase string, and completely avoids the parsing of a human readable/prettified  string

kernels/src/kernels/benchmark.py
```diff
@@ -409,6 +405,17 @@ def _get_macos_gpu() -> tuple[str | None, int | None]:
...
 
+def get_device() -> str:
+    if TORCH_AVAILABLE:
+        if torch.cuda.is_available():
+            return "cuda"
+        elif hasattr(torch, "xpu") and torch.xpu.is_available():
+            return "xpu"
+        elif hasattr(torch.backends, "mps") and torch.backends.mps.is_available():
+            return "mps"
+    return "cpu"
```
danieldk: I think we can reuse this function: https://github.com/huggingface/kernels/blob/3ce9687e33e85f530aa06ed691ad1914e2aacc01/kernels/src/kernels/utils.py#L50

But we can do that in a follow-up PR.

drbh: oh great point, just took a look and updated in this PR!

PR #245, opened 2026-01-27
build2cmake: always generate kernel components for all backends

danieldk:
This change always generates the kernel components for all backends. This is a necessary step for merging the preambles later.

### DIFF
--- a/build2cmake/src/torch/kernel.rs
+++ b/build2cmake/src/torch/kernel.rs
@@ -51,7 +51,7 @@ fn render_kernel_component(
     Ok(())
 }
 
-pub fn render_kernel_component_cpu(
+fn render_kernel_component_cpu(
     env: &Environment,
     kernel_name: &str,
     kernel: &Kernel,
@@ -134,7 +134,7 @@ fn render_kernel_component_cuda(
     Ok(())
 }
 
-pub fn render_kernel_component_metal(
+fn render_kernel_component_metal(
     env: &Environment,
     kernel_name: &str,
     kernel: &Kernel,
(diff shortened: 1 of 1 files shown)


build2cmake/src/torch/kernel.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+        Kernel::Xpu { .. } => {
+            render_kernel_component_xpu(env, kernel_name, kernel, sources, write)?
+        }
+    }
+
+    Ok(())
+}
+
+pub fn render_kernel_component_cpu(
```
drbh: totally a nit, but maybe all the `render_kernel_*` can be private like the `render_kernel_component_cuda`?

```suggestion
fn render_kernel_component_cpu(
```

build2cmake/src/torch/kernel.rs
```diff
@@ -0,0 +1,191 @@
...
+            &mut *write,
+        )
+        .wrap_err("Cannot render kernel template")?;
+
+    write.write_all(b"\n")?;
+
+    Ok(())
+}
+
+pub fn render_kernel_component_metal(
```
drbh:
```suggestion
fn render_kernel_component_metal(
```

build2cmake/src/torch/kernel.rs
```diff
@@ -0,0 +1,191 @@
...
+            &mut *write,
+        )
+        .wrap_err("Cannot render kernel template")?;
+
+    write.write_all(b"\n")?;
+
+    Ok(())
+}
+
+pub fn render_kernel_component_xpu(
```
drbh:
```suggestion
fn render_kernel_component_xpu(
```

drbh (approved):
nice improvement, looks good to me! just added a tiny nit but not needed to merge

PR #247, opened 2026-01-28
add init command that pulls template repo

drbh:
this PR adds a new `init` subcommand that clones a huggingface 

```bash
kernels init drbh/my-kernel
```

outputs
```
Downloading template from drbh/template...
Initialized kernel project: /Users/drbh/Projects/kernels/my-kernel
├── my_kernel_cpu
│   └── my_kernel_cpu.cpp
├── my_kernel_cuda
│   └── my_kernel.cu
├── my_kernel_metal
│   ├── my_kernel.metal
│   └── my_kernel.mm
├── my_kernel_xpu
│   └── my_kernel.cpp
├── tests
│   ├── __init__.py
│   └── test_my_kernel.py
├── torch-ext
│   ├── my_kernel
│   │   └── __init__.py
│   ├── torch_binding.cpp
│   └── torch_binding.h
├── .gitattributes
├── .gitignore
├── README.md
├── build.toml
├── example.py
└── flake.nix

Next steps:
  cd my-kernel
  nix run .#build-and-copy -L
  uv run example.py
```

now build
```bash
cd my-kernel
nix run .#build-and-copy -L

# tree build -L 1
# build
# ├── torch210-cpu-aarch64-darwin
# ├── torch210-metal-aarch64-darwin
# ├── torch29-cpu-aarch64-darwin
# └── torch29-metal-aarch64-darwin

# 5 directories, 0 files
```

then run

```bash
uv run example.py
```

outputs
```
Using device: mps
Input:  tensor([1., 2., 3.], device='mps:0')
Output: tensor([2., 3., 4.], device='mps:0')
Success!
```

and tested on a cuda device

```bash
# ... same commands as above

# tree build -L 1
# build
# ├── torch210-cxx11-cpu-x86_64-linux
# ├── torch210-cxx11-cu126-x86_64-linux
# ├── torch210-cxx11-cu128-x86_64-linux
# ├── torch210-cxx11-cu130-x86_64-linux
# ├── torch210-cxx11-rocm70-x86_64-linux
# ├── torch210-cxx11-rocm71-x86_64-linux
# ├── torch210-cxx11-xpu20253-x86_64-linux
# ├── torch29-cxx11-cpu-x86_64-linux
# ├── torch29-cxx11-cu126-x86_64-linux
# ├── torch29-cxx11-cu128-x86_64-linux
# ├── torch29-cxx11-cu130-x86_64-linux
# ├── torch29-cxx11-rocm63-x86_64-linux
# ├── torch29-cxx11-rocm64-x86_64-linux
# └── torch29-cxx11-xpu20252-x86_64-linux

# 14 directories, 0 files
```

outputs
```
Using device: cuda
Input:  tensor([1., 2., 3.], device='cuda:0')
Output: tensor([2., 3., 4.], device='cuda:0')
Success!

```

### DIFF
--- a/kernels/pyproject.toml
+++ b/kernels/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
   "packaging>=20.0",
   "pyyaml>=6",
   "tomli>=2.0; python_version<'3.11'",
-  "tomlkit==0.13.3",
+  "tomlkit>=0.13.3",
 ]
 
 [build-system]


kernels/src/kernels/init.py
```diff
@@ -8,9 +9,40 @@
...
+from kernels.utils import KNOWN_BACKENDS
+
 
 def run_init(args: Namespace) -> None:
     kernel_name = args.kernel_name
+    if args.backends is None:
+        backends = ["metal"] if sys.platform == "darwin" else ["cuda"]
+    else:
+        backends = [
+            v.strip().lower()
```
danieldk: Is stripping and lowercasing needed? I think all the whitespace should be consumed by the argument parser. I think it's better if people just have to give the arguments in lowercase.

kernels/src/kernels/cli.py
```diff
@@ -162,6 +162,12 @@ def main():
         default="drbh/template",
         help="HuggingFace repo ID for the template",
     )
+    init_parser.add_argument(
+        "--backends",
+        nargs="+",
+        default=None,
```
danieldk: I think we can use `action="extend"` and then the user can use something like `--backends cuda rocm`.

kernels/src/kernels/init.py
```diff
@@ -8,9 +9,40 @@
...
+            for item in args.backends
+            for v in item.split(",")
```
danieldk: See above, I don't think we need extra parsing, we can let argparse handle it.

kernels/src/kernels/cli.py
```diff
@@ -162,6 +162,12 @@ def main():
...
+        help="Backends to include ('all' or list like: cpu cuda metal rocm xpu). Defaults: cuda on Linux/Windows, metal on macOS.",
```
danieldk: Since this is a finite list, it would be nice to use `choices=...`. We can also set the default to `cuda` or `metal` here. I think argparse will then even show it in the usage.

kernels/src/kernels/init.py
```diff
@@ -8,9 +9,40 @@
...
 from huggingface_hub.utils import disable_progress_bars
 
+from kernels.compat import tomllib
+from kernels.utils import KNOWN_BACKENDS
+
 
 def run_init(args: Namespace) -> None:
     kernel_name = args.kernel_name
+    if args.backends is None:
+        backends = ["metal"] if sys.platform == "darwin" else ["cuda"]
```
danieldk: See above.

kernels/src/kernels/init.py
```diff
@@ -8,9 +9,40 @@
...
+        if "all" in backends:
+            if len(backends) > 1:
+                print(
+                    "Error: --backends must be either 'all' or a list of backends.",
+                    file=sys.stderr,
+                )
+                sys.exit(1)
+            backends = []
+        else:
+            valid = set(KNOWN_BACKENDS)
```
danieldk: See above, best to let argparse do the work.

kernels/src/kernels/init.py
```diff
@@ -140,3 +175,58 @@ def _print_tree(directory: Path, prefix: str = "") -> None:
...
+        if isinstance(cfg, dict) and cfg.get("backend") not in set(backends)
+    }
+    backends_list = ", ".join(f'"{b}"' for b in backends)
+    new_line = f"backends = [{backends_list}]"
+    pattern = r"(\[general\][\s\S]*?)^\s*backends\s*=\s*\[[^\]]*\]"
+    new_text, count = re.subn(pattern, r"\1" + new_line, text, count=1, flags=re.M)
+    if remove_kernels:
+        new_text = _remove_kernel_sections(new_text, remove_kernels)
+    if count or remove_kernels:
+        build_toml_path.write_text(new_text)
```
danieldk: I think it's cleaner to parse the TOML, modify the Python data structures and then serialize as TOML again. regexps are fragile.

kernels/src/kernels/utils.py
```diff
@@ -22,6 +22,7 @@
 from kernels.metadata import Metadata
 
 ENV_VARS_TRUE_VALUES = {"1", "ON", "YES", "TRUE"}
+KNOWN_BACKENDS = ("cpu", "cuda", "metal", "rocm", "xpu", "npu")
```
danieldk: Better to make it a set?

kernels/src/kernels/init.py
drbh: totally agree, the main reason I opt'ed for regex over toml is that I was concerned that writing toml back into the file would lose any comments/formatting added to the template. 

I think its a reasonable trade off (more robust toml parsing, instead of comments) however my initial approach attempted to be comment preserving. 

happy to update with the toml route, just wanted to flag the reasoning/tradeoff

danieldk: Maybe we should consider using tomlkit, since it can preserve comments?

kernels/src/kernels/init.py
drbh: great suggestion the code is much cleaner in the latest commit, thanks!

kernels/src/kernels/cli.py
drbh: ended up using `"+"` instead of extend so if a user specifies a backend it overrides the default, otherwise we default to metal on darwin and cuda in all other cases

kernels/src/kernels/init.py
drbh: same as above, updated in latest commit

kernels/src/kernels/cli.py
drbh: same as above, updated in latest commit

kernels/src/kernels/init.py
drbh: same as above, updated in latest commit

kernels/src/kernels/init.py
drbh: fully agree, same as above, updated in latest commit

kernels/src/kernels/init.py
drbh: great suggestion, I've added `tomlkit` as a dep in the latest commits. Im still open to using the builtins if we don't see a benefit from preserving comments/whitespace, but the current changes are working well

kernels/src/kernels/utils.py
drbh: agreed, updated in latest commit

kernels/pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
   "packaging>=20.0",
   "pyyaml>=6",
   "tomli>=2.0; python_version<'3.11'",
+  "tomlkit==0.13.3",
```
danieldk: We should use a a lower bound, not a hard bound. This becomes really hard for downstream projects if they need a newer version.

PR #255, opened 2026-02-02
Remove the last backend-specific writer functions

danieldk:


drbh (approved):
great improvement, its much nicer that everything follows a common path now. 

LGTM!

PR #258, opened 2026-02-03
Fix the XPU compilation issue

YangKai0616:
The currently compiled XPU kernels are unavailable due to the missing `sycl_flags` parameter. This PR:

1. Fixes the XPU compilation issue
2. Resolves the `CUTLASS_SYCL_REVISION` matching problem
3. Moves the XPU SYCL compiler configuration before the `project()` call to prevent repeated reconfiguration and infinite loops.

The nix build results have passed correctness tests successfully on both **XPU** and **CUDA**.


YangKai0616:
Noticed that PR https://github.com/huggingface/kernels/pull/256 has fixed the 3 point, updating this PR accordingly.

Cc @sywangyi

PR #259, opened 2026-02-03
Remove backend-specific generation and also use CMake variant generation in Nix

danieldk:
This combines several changes that are mostly co-dependent:

- Remove the `--backend` option from `build2cmake`, we now generate a `CMakeLists.txt` that works with all backends. The backend is determined by querying PyTorch.
- On Linux and Darwin we move the install phase fully from Nix to CMake for arch kernels. This makes Linux, Darwin, and Windows share code paths for installation.
- In the Nix builder, a single kernel build will be in `result/<variant-name>/` in place of `result/`.
- Add installation of the compatibility module to CMake installation (this fixes Windows to support old and new-style kernel paths).
- Fixes in the CMake installation e.g.: previously kernel names with dashes were not handled correctly and Windows builds were not correctly recursively copying Python files.
- Scrub test shell `PYTHONPATH` for better impure tests.

### DIFF
--- a/builder/lib/build.nix
+++ b/builder/lib/build.nix
@@ -217,7 +217,17 @@ rec {
         bundleOnly = true;
       };
       buildToml = readBuildConfig path;
-      contents = builtins.map (pkg: toString pkg) (builtins.attrValues extensions);
+      benchmarksPath = path + "/benchmarks";
+      hasBenchmarks = builtins.pathExists benchmarksPath;
+      benchmarks =
+        with lib.fileset;
+        toSource {
+          root = path;
+          fileset = maybeMissing benchmarksPath;
+        };
+      contents =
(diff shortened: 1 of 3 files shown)


builder/examples/extra-data/torch-ext/extra_data/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+# non-Python data. This can be used for e.g. Triton tuning data.
+
+
+def _read_json() -> dict:
+    json_path = Path(__file__).parent / "data.json"
+    with open(json_path, "r") as f:
+        return json.load(f)
+
+
+EASTER_EGG = _read_json()
```
drbh, line 22: haha great name

drbh (approved):
LGTM

PR #262, opened 2026-02-04
Fix example use relu kernel as an example

MekkCyber:


docs/source/basic-usage.md
```diff
@@ -2,26 +2,26 @@
...
-Here is how you would use the [activation](https://huggingface.co/kernels-community/activation) kernels from the Hugging Face Hub:
+Here is how you would use the [relu](https://huggingface.co/kernels-community/relu) kernels from the Hugging Face Hub:
 
 ```python
 import torch
 
 # Download optimized kernels from the Hugging Face hub
-activation = get_kernel("kernels-community/activation", version=1)
+relu = get_kernel("kernels-community/relu", version=1)
```
danieldk, line 12: I am not sure if we want to move to `kernels-community/relu` as an example? This kernel is our testing playground.

MekkCyber (commented):
Will close this as `activation` kernels have cuda/metal builds now

PR #263, opened 2026-02-04
Clean-up huggingface_hub integration

Wauplin:
This PR cleans up a bit the integration with `huggingface_hub`. In particular, it defines a `_get_hf_api` helper that instantiate an `HfApi` client with the correct user agent set. The goal is to have consistent user agent across all calls made to the Hub.

List of changes:
- add `kernels.utils._get_hf_api` to return a `HfApi` instance
- use it everywhere in `kernels`. No more `from huggingface_hub import create_repo, snapshot_download, etc.` anymore
- slightly refactored user agent creation to always return a string (=> deduplicates some logic). Removed `_user_agents` helper (not used anymore)
- rely on `huggingface_hub.constants.HF_HUB_DISABLE_TELEMETRY` instead of `os.getenv("DISABLE_TELEMETRY")`. The hfh constant is more robust (e.g. handles `DO_NOT_TRACK`  env variable)
- removed `requests` from benchmark dependencies. Use `get_session`, `build_hf_headers` and `hf_raise_for_status` instead. This should get you better logs in case of errors and relies either on `requests` or `httpx` depending on what's available. It also handles token retrieval by default. 

Overall `kernels` behavior and API remains exactly the same. Changes are meant to be purely internal.

I am sorry in advance for all the style changes. I've run `make style` locally and inadvertently pushed more changes than what I thought. Let me know if you want me to try to revert them.

### DIFF
--- a/kernels/tests/test_layer.py
+++ b/kernels/tests/test_layer.py
@@ -18,8 +18,13 @@ from kernels import (
     use_kernel_forward_from_hub,
     use_kernel_mapping,
 )
-from kernels.layer.layer import _KERNEL_MAPPING, _validate_layer
-from kernels.utils import install_kernel
+from kernels.layer.layer import (
+    _KERNEL_MAPPING,
+    _validate_layer,
+)
+from kernels.utils import (
+    install_kernel,
+)
 
 kernel_layer_mapping = {
     "SiluAndMul": {
@@ -542,8 +547,7 @@ def test_validate_kernel_layer():
         _validate_layer(cls=BadLayer2, check_cls=SiluAndMul, repo=stub_repo(BadLayer2))
 
     class BadLayer3(nn.Module):
-        def forward(self, x: torch.Tensor, foo: int) -> torch.Tensor:
-            ...
+        def forward(self, x: torch.Tensor, foo: int) -> torch.Tensor: ...
 
     with pytest.raises(
         TypeError,
@@ -552,8 +556,7 @@ def test_validate_kernel_layer():
         _validate_layer(cls=BadLayer3, check_cls=SiluAndMul, repo=stub_repo(BadLayer3))
 
     class BadLayer4(nn.Module):
-        def forward(self, *, x: torch.Tensor) -> torch.Tensor:
-            ...
+        def forward(self, *, x: torch.Tensor) -> torch.Tensor: ...
 
     with pytest.raises(
         TypeError,


kernels/src/kernels/utils.py
```diff
@@ -555,38 +553,29 @@ def package_name_from_repo_id(repo_id: str) -> str:
     return repo_id.split("/")[-1].replace("-", "_")
 
 
-def _get_user_agent(
```
sayakpaul, line 559: IMO, it wouldn't hurt to keep this function and use it inside `_get_hf_api()`. I think that's more idiomatic and structured.

kernels/src/kernels/utils.py
```diff
@@ -555,38 +553,29 @@ def package_name_from_repo_id(repo_id: str) -> str:
...
+            user_agent_str += f"; glibc/{glibc}"
 
-    elif isinstance(user_agent, str):
-        user_agent += f"; kernels/{__version__}; python/{python}; torch/{torch.__version__}; build_variant/{build_variant()}; file_type/kernel"
-        if glibc is not None:
-            user_agent += f"; glibc/{glibc}"
-
-    return user_agent
+    return HfApi(
+        library_name="kernels", library_version=__version__, user_agent=user_agent_str
```
sayakpaul, line 587: Very nice refactor!

kernels/src/kernels/utils.py
```diff
@@ -555,38 +553,29 @@ def package_name_from_repo_id(repo_id: str) -> str:
...
-    if isinstance(user_agent, dict):
-        user_agent.update(
-            {
-                "kernels": __version__,
-                "python": python,
-                "torch": torch.__version__,
-                "build_variant": build_variant(),
-                "file_type": "kernel",
-            }
-        )
```
sayakpaul, line 584: Hopefully, it doesn't break anything. Cc: @MekkCyber do you reckon anything to be breaking for this?

kernels/uv.lock
```diff
@@ -135,111 +135,6 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/e6/ad/3cc14f097111b4de0040c83a525973216457bbeeb63739ef1ed275c1c021/certifi-2026.1.4-py3-none-any.whl", hash = "sha256:9943707519e4add1115f44c2bc244f782c0249876bf51b6599fee1ffbedd685c", size = 152900, upload-time = "2026-01-04T02:42:40.15Z" },
 ]
 
-[[package]]
-name = "charset-normalizer"
```
sayakpaul, line 139: Explain this change a bit?

kernels/src/kernels/utils.py
Wauplin: my goal was to deduplicate logic (the user agent was defined both in a dictionary format and a string format, now we only define it once)

kernels/uv.lock
Wauplin: this is a lock file so not something I've created myself manually. I suppose it has to do with the removal of the `requests` package? (but honestly not something I controlled myself)

kernels/src/kernels/utils.py
MekkCyber: good for me since we still convert the dict to str so not breaking

kernels/src/kernels/utils.py
MekkCyber: indeed ! thanks @Wauplin

MekkCyber (approved):
Nice refactor! thanks for pr @Wauplin

sayakpaul (dismissed):

PR #265, opened 2026-02-05
Support local kernels in benchmark

drbh:
This PR updates the benchmark command to allow locally build kernels to be used if the repo_id starts with a `path:` prefix. 

example
```bash
kernels benchmark path:/Users/drbh/Projects/kernels-community/activation
```
*note this requires and absolute path

this feature is helpful in case a user want to bench a locally built kernel that hasn't be pushed to the hub yet

### DIFF
--- a/kernels/src/kernels/benchmark.py
+++ b/kernels/src/kernels/benchmark.py
@@ -7,6 +7,7 @@ import random
 import subprocess
 import sys
 import time
+import warnings
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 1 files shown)


MekkCyber (approved):

kernels/src/kernels/benchmark.py
```diff
@@ -493,9 +493,14 @@ def run_benchmark_class(
         raise RuntimeError(f"No benchmark_* methods found in {benchmark_cls.__name__}")
 
     # Load kernel once for all workloads
-    from kernels import get_kernel
+    from kernels import get_local_kernel, get_kernel
+
+    if repo_id.startswith("path:"):
```
danieldk: Any downside to checking if the argument is a local path first? Then we do not need a special prefix.

drbh: ah yea thats a good point, in the case the local path is absolute (which is required now) it makes sense to simply check if the repo id is a path. 

the only downside I can think of is if we update to support relative paths, than there could be a conflict between a local path like `foo/bar` and a repo on the hub since they both have the same format..

PR #266, opened 2026-02-06
cmake: merge loops for handing Python and data extensions

danieldk:


build2cmake/src/templates/build-variants.cmake
```diff
@@ -153,13 +157,15 @@ function(add_kernels_install_target TARGET_NAME PACKAGE_NAME BUILD_VARIANT_NAME)
...
-          COMPONENT ${TARGET_NAME})
+    # Install data files with specified extensions
+    foreach(ext IN LISTS ALL_EXTENSIONS)
+        file(GLOB_RECURSE DATA_FILES RELATIVE "${CMAKE_SOURCE_DIR}/torch-ext/${PACKAGE_NAME}" "${CMAKE_SOURCE_DIR}/torch-ext/${PACKAGE_NAME}/*.${ext}")
+        foreach(data_file IN LISTS DATA_FILES)
+            get_filename_component(data_file_dir "${data_file}" DIRECTORY)
+            install(FILES "${CMAKE_SOURCE_DIR}/torch-ext/${PACKAGE_NAME}/${data_file}"
+                DESTINATION "${KERNEL_INSTALL_DIR}/${data_file_dir}"
+                COMPONENT ${TARGET_NAME})
+        endforeach()
```
MekkCyber, line 168: qq: is there a reason we add the inner loop ? `install`ing multiple files doesn't preserve the source structure ?

build2cmake/src/templates/build-variants.cmake
```diff
@@ -284,6 +257,20 @@ function(add_local_install_target TARGET_NAME PACKAGE_NAME BUILD_VARIANT_NAME)
...
+            ${CMAKE_SOURCE_DIR}/metadata-${_BACKEND}.json
+            ${VARIANT_DIR}/metadata.json
+
+            # Compatibility with older kernels and direct Python imports.
+            COMMAND ${CMAKE_COMMAND} -E copy_if_different
+            ${CMAKE_SOURCE_DIR}/compat.py
+            ${VARIANT_DIR}/${PACKAGE_NAME}/__init__.py
+
+            COMMAND_EXPAND_LISTS
+    )
```
MekkCyber, line 268: why isn't this handled with other extensions ?

MekkCyber (dismissed):
lgtm! I have some quick questions

build2cmake/src/templates/build-variants.cmake
danieldk: Correct. E.g. if we have `json`, there might be multiple `*.json` that could be in different directories and it seems we can only properly preserve the target directory structure by looping over them.

build2cmake/src/templates/build-variants.cmake
danieldk: Oh, it's above, the code already existed, we just added the handling of `.py` to the same loop. This particular block just moved, though not completely, I have fixed that up - the diff should look cleaner now.

PR #268, opened 2026-02-06
`get_kernel`: support specifying the backend

danieldk:
So far, `get_kernel` has always auto-detected the backend from the Torch build. However, this causes issues in fixed-device settings where we may want to run some ops on the CPU and others on the GPU. This change allows specifying `backend` as an argument to `get_kernel` and friends.

Currently, only the `cpu` backend and the backend that Torch was built for are supported (since we cannot determine the version, etc. of a backend that Torch was not built for). However, we decided to make it a string-based argument as opposed to e.g. a `cpu` bool, in the case Torch might do multi-backend builds in the future.

### DIFF
--- a/.github/workflows/test_kernels.yaml
+++ b/.github/workflows/test_kernels.yaml
@@ -38,13 +38,13 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install Torch version
-        working-directory: ./kernels
-        run: uv pip install "torch==${{ matrix.torch-version }}"
-
       - name: Install the project
         working-directory: ./kernels
         run: uv sync --all-extras --dev
+
+      - name: Install Torch version
+        working-directory: ./kernels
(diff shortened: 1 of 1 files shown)


kernels/src/kernels/utils.py
```diff
@@ -102,36 +104,57 @@ def build_variant() -> str:
...
-def build_variant_noarch() -> str:
-    import torch
+def _supported_backends() -> set[str]:
+    return {"cpu", _backend()}
 
-    if torch.version.cuda is not None:
+
+def _select_backend(backend: str | None) -> str:
+    if backend is None:
+        backend = _backend()
```
MekkCyber: why not simply return `_backend()`here ?

MekkCyber (approved):
Looks good! just a nit

kernels/src/kernels/utils.py
danieldk: Great catch! Fixed.

PR #269, opened 2026-02-09
Make CLI-related modules submodules of `cli`

danieldk:
Also add missing dependencies to the Nix flake.


MekkCyber (approved):
nice refactor

PR #272, opened 2026-02-10
Fix versions torch dependency

### LINKED ISSUE #264: Using `uvx kernels versions` errors out complaining `torch` not installed
Issuing the command
```
uvx kernels versions kernels-community/activation
```

Provides the following error
```
Traceback (most recent call last):
  File "/Users/arig23498/.cache/uv/archive-v0/mwFkb56SgsErYaHpkNGI2/bin/kernels", line 12, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/arig23498/.cache/uv/archive-v0/mwFkb56SgsErYaHpkNGI2/lib/python3.14/site-packages/kernels/cli.py", line 150, in main
    args.func(args)
    ~~~~~~~~~^^^^^^
  File "/Users/arig23498/.cache/uv/archive-v0/mwFkb56SgsErYaHpkNGI2/lib/python3.14/site-packages/kernels/cli.py", line 191, in kernel_versions
    print_kernel_versions(args.repo_id)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/Users/arig23498/.cache/uv/archive-v0/mwFkb56SgsErYaHpkNGI2/lib/python3.14/site-packages/kernels/versions_cli.py", line 12, in print_kernel_versions
    compatible_variants = set(build_variants())
                              ~~~~~~~~~~~~~~^^
  File "/Users/arig23498/.cache/uv/archive-v0/mwFkb56SgsErYaHpkNGI2/lib/python3.14/site-packages/kernels/utils.py", line 128, in build_variants
    return [build_variant(), build_variant_noarch(), build_variant_universal()]
            ~~~~~~~~~~~~~^^
  File "/Users/arig23498/.cache/uv/archive-v0/mwFkb56SgsErYaHpkNGI2/lib/python3.14/site-packages/kernels/utils.py", line 68, in build_variant
    import torch
ModuleNotFoundError: No module named 'torch'
```

This originates from the import
https://github.com/huggingface/kernels/blob/01fe10893ad76c0869a6be6806e417ca2b3fc755/kernels/src/kernels/utils.py#L68

I see that the project does not depend on `torch` and that leads to above mentioned error. What do you think should be a good solution to this, considering that we want to make the tool compatible with `uvx` (`uv tool run` that is).

danieldk:
If Torch is not available, we will not show a check mark for the compatible build variant.

Fixes #264.

### DIFF
--- a/.github/workflows/test_kernels.yaml
+++ b/.github/workflows/test_kernels.yaml
@@ -95,3 +95,9 @@ jobs:
         run: |
           uv pip uninstall torch
           uv run python -c "import kernels"
+
+      - name: Run command-line tools without Torch
+        working-directory: ./kernels
+        run: |
+          uv run kernels check kernels-community/activation
+          uv run kernels versions kernels-community/activation


kernels/src/kernels/cli/versions.py
```diff
@@ -9,7 +10,12 @@
 
 def print_kernel_versions(repo_id: str):
     api = HfApi()
-    compatible_variants = set(build_variants())
+
+    if find_spec("torch") is None:
+        # Do not mark compatible variants when Torch is not available.
+        compatible_variants = set()
```
MekkCyber, line 16: not sure if it's okay to add a warning here or not ?

danieldk: I don't think it's necessary, we still provide the correct list of versions. The check mark is more of a handy convenience.

MekkCyber: okay sg

PR #276, opened 2026-02-10
feat: add init and benchmarks commands to docs

drbh:
This PR simply adds new docs for the `init` and `benchmark` subcommands

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -63,4 +63,8 @@
       title: kernels lock
     - local: cli-download
       title: kernels download
+    - local: cli-skills
+      title: kernels skills
+    - local: cli-create-and-upload-card
+      title: kernels create-and-upload-card
   title: CLI Reference


docs/source/cli-init.md
```diff
@@ -0,0 +1,61 @@
...
+```bash
+cd my_kernel
+cachix use huggingface
+nix run -L --max-jobs 1 --cores 8 .#build-and-copy
+uv run example.py
+```
+
+## Notes
+
+- The `<repo>` part is normalized to lowercase, and `-` is replaced with `_`. For example, `my-user/My-Kernel` becomes a directory named `my_kernel` and a repo id `my-user/my_kernel`.
```
danieldk: I think I missed this when reviewing `init`, `-` should only be converted to lowercase in the Python package name (`build/<variant>/my_kernel`), not in the repo name, the kernel directory name, or the kernel name in `build.toml`. We prefer regular dashes there and maybe we should even do the opposite, let kernel init change `my-user/my_kernel` to `myuser/my-kernel`.

drbh: ah yea great point, just pushed changes to prefer dashes in the dir the kernel name and underscores/lowercase in the python naming

PR #278, opened 2026-02-11
[FEATURE] add `kernels skills add` to the cli

burtenshaw:
This PR adds a new cli command for installing an agent compatible skill. Things to note:

- cli implementation is based on https://github.com/huggingface/huggingface_hub/pull/3741/changes
- I have ported the skill from here: https://github.com/huggingface/huggingface_hub/pull/3741/changes. Tested and benchmarked with @sayakpaul 
- `kernels` will need to vendor this skill. I have added it to docs but not ToC. I think this is the easiest way, but I'm open to suggestions.

You can test this out with:

- `kernels skills --help`
- `kernels skills add --claude` etc.

### DIFF
--- a/docs/source/cli.md
+++ b/docs/source/cli.md
@@ -34,7 +34,7 @@ your kernel builds to the Hub. To know the supported arguments run: `kernels upl
 
 ### kernels skills add
 
-Use `kernels skills add` to install the skills for AI coding assistants like Claude, Codex, and OpenCode. For now, only the `cuda-kernels` skill is supported. Skill files are downloaded from the `huggingface/kernels` directory in this [repository](https://github.com/huggingface/kernels/blob/skills).
+Use `kernels skills add` to install the skills for AI coding assistants like Claude, Codex, and OpenCode. For now, only the `cuda-kernels` skill is supported. Skill files are downloaded from the `huggingface/kernels` directory in this [repository](https://github.com/huggingface/kernels/tree/main/skills).
 
 Skills instruct agents how to deal with hardware-specific optimizations, integrate with libraries like diffusers and transformers, and benchmark kernel performance in consistent ways.
 


docs/source/cli.md
```diff
@@ -31,3 +31,24 @@ your kernel builds to the Hub. To know the supported arguments run: `kernels upl
 - If a repo with the `repo_id` already exists and if it contains a `build` with the build variant
   being uploaded, it will attempt to delete the files existing under it.
 - Make sure to be authenticated (run `hf auth login` if not) to be able to perform uploads to the Hub.
+
+### kernels skills add
+
+Use `kernels skills add` to install the `cuda-kernels` skill for AI coding assistants.
```
sayakpaul: `cuda-kernels` could be renamed to `custom-kernels` to accommodate other accelerators as well.

(Edit: We're providing one example to write a CUDA kernel for a specific set of hardware for a specific model. I think it would make sense to clarify that here otherwise the impression might be a bit different).

docs/source/cli.md
```diff
@@ -31,3 +31,24 @@ your kernel builds to the Hub. To know the supported arguments run: `kernels upl
...
+Skill files are downloaded from the `skills/cuda-kernels` directory in this repository.
```
sayakpaul: Let's provide a bit more context about what this skill enables (perhaps in one/two lines)?

kernels/src/kernels/cli/skills.py
```diff
@@ -0,0 +1,118 @@
...
+        output_file = dest / rel_path
+        output_file.parent.mkdir(parents=True, exist_ok=True)
+        output_file.write_text(content, encoding="utf-8")
+
+    return dest
+
+
+def add_skill(args: Namespace) -> None:
+    if not (args.claude or args.codex or args.opencode or args.dest):
+        print(
```
sayakpaul, line 98: We should be able to do it via exposing a `vendor` option in the `args` and then restricting its `choices` to `claude, codex, etc.` for example? Do you reckon any disadvantages? The vendor could be an `nargs` to allow lists as an input.

docs/source/cli.md
burtenshaw: Ok. I'll clarify. We of course have not tested on other hardware, but in those cases I expect users would want separate skills.

kernels/src/kernels/cli/skills.py
burtenshaw: It's a makes sense, but this in consistent with other cli tools. like `hf` and `skills`

docs/source/cli.md
```diff
@@ -31,3 +31,25 @@ your kernel builds to the Hub. To know the supported arguments run: `kernels upl
 - If a repo with the `repo_id` already exists and if it contains a `build` with the build variant
   being uploaded, it will attempt to delete the files existing under it.
 - Make sure to be authenticated (run `hf auth login` if not) to be able to perform uploads to the Hub.
+
+### kernels skills add
+
+Use `kernels skills add` to install the skills for AI coding assistants like Claude, Codex, and OpenCode. For now, only the `cuda-kernels` skill is supported. Skill files are downloaded from the `skills/cuda-kernels` directory in this [repository](https://github.com/burtenshaw/kernels/blob/skills).
```
danieldk: I think this should be


```suggestion
Use `kernels skills add` to install the skills for AI coding assistants like Claude, Codex, and OpenCode. For now, only the `cuda-kernels` skill is supported. Skill files are downloaded from the `huggingface/kernels` directory in this [repository](https://github.com/huggingface/kernels/blob/skills).
```

danieldk (approved):
Looks great, thanks for adding this!

Will merge once CI is happy.

docs/source/cli.md
```diff
@@ -31,3 +31,25 @@ your kernel builds to the Hub. To know the supported arguments run: `kernels upl
 - If a repo with the `repo_id` already exists and if it contains a `build` with the build variant
   being uploaded, it will attempt to delete the files existing under it.
 - Make sure to be authenticated (run `hf auth login` if not) to be able to perform uploads to the Hub.
+
+### kernels skills add
+
+Use `kernels skills add` to install the skills for AI coding assistants like Claude, Codex, and OpenCode. For now, only the `cuda-kernels` skill is supported. Skill files are downloaded from the `huggingface/kernels` directory in this [repository](https://github.com/huggingface/kernels/blob/skills).
```
pcuenca:
```suggestion
Use `kernels skills add` to install the skills for AI coding assistants like Claude, Codex, and OpenCode. For now, only the `cuda-kernels` skill is supported. Skill files are downloaded from the `huggingface/kernels` directory in this [repository](https://github.com/huggingface/kernels/tree/main/skills).
```

Not sure `blob` would work here.

kernels/src/kernels/cli/skills.py
pcuenca: Yes, I think keeping consistency with `hf` is useful.

PR #281, opened 2026-02-13
gen-flake-outputs: add `ci-test` package

danieldk:
Running this output will run the kernel's tests. CI tests must be marked with `kernels_ci`. This avoids running all tests, which may be too expensive for CI.

Example run: https://github.com/huggingface/kernels-community/actions/runs/21980763830/job/63503495972?pr=364


docs/source/builder/writing-kernels.md
```diff
@@ -280,3 +284,27 @@ def relu(x: torch.Tensor, out: Optional[torch.Tensor] = None) -> torch.Tensor:
...
     return out
 ```
+
+## Kernel tests
+
+Kernel tests are stored in the `tests` directory. Since running all
+kernel tests in CI may be prohibitively expensive, the `pyproject.toml`
+generated by the builder adds support for the special `kernels_ci`
+PyTest marker that cam be used as follows:
+
```
MekkCyber:
```suggestion
PyTest marker that can be used as follows:

```

builder/examples/relu/tests/test_relu.py
```diff
@@ -19,6 +21,7 @@ def test_relu():
     torch.testing.assert_allclose(F.relu(x), relu.relu(x))
 
 
+@pytest.mark.kernels_ci
 def test_relu_layer():
```
sayakpaul, line 25: I am guessing this test suite is sufficiently fast?

docs/source/builder/writing-kernels.md
```diff
@@ -280,3 +284,27 @@ def relu(x: torch.Tensor, out: Optional[torch.Tensor] = None) -> torch.Tensor:
     ops.relu(out, x)
     return out
 ```
+
+## Kernel tests
+
+Kernel tests are stored in the `tests` directory. Since running all
```
sayakpaul, line 290: Should we add guidance on what kind of tests would be ideal for this? Like we would want tests that run fast yet provide some confidence?

docs/source/builder/writing-kernels.md
```diff
@@ -280,3 +284,27 @@ def relu(x: torch.Tensor, out: Optional[torch.Tensor] = None) -> torch.Tensor:
...
+  ...
+```
+
+You can run the tests (e.g. in CI) using:
+
+```bash
+$ nix run .#ci-test
+```
+
+On non-NixOS systems, make sure that
```
sayakpaul: So, currently, it seems limited to CUDA-only tests?

builder/examples/relu/tests/test_relu.py
danieldk: A second at most :).

docs/source/builder/writing-kernels.md
danieldk: It also works on non-CUDA, I should update the doc to mention this. Basically it is like this:

* If the kernels supports one of CUDA/ROCm/XPU/Metal, the test for the supported backend will be run.
* If the kernels supports multiple backends, CUDA is the default on Linux and Metal on macOS.
* You can always run the tests for any variant with `nix run .#ciTests.<variant>`, for instance: `nix run .#ciTests.torch210-cxx11-cpu-x86_64-linux`

sayakpaul (approved):

docs/source/builder/writing-kernels.md
sayakpaul: Thanks for explaining. Should we include this more explicitly in the docs then?

danieldk: Did it in the latest change.

PR #285, opened 2026-02-16
`kernels`: add support for Neuron and NKI

danieldk:
This change add support for Torch Neuron and NKI to the `kernels` Python package. Support for builder is not added in this change.

### DIFF
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -95,7 +95,7 @@ def _backend() -> str:
 
 def _build_variant(backend: str | None) -> str | None:
     """
-    Build vaariant for arch kernels, returns `None` when the backend
+    Build variant for arch kernels, returns `None` when the backend
     does not (yet) support arch kernels.
     """
     backend = _select_backend(backend)


kernels/src/kernels/utils.py
```diff
@@ -88,7 +92,11 @@ def _backend() -> str:
         return "cpu"
 
 
-def _build_variant(backend: str | None) -> str:
+def _build_variant(backend: str | None) -> str | None:
+    """
+    Build vaariant for arch kernels, returns `None` when the backend
+    does not (yet) support arch kernels.
```
MekkCyber:
```suggestion
    Build variant for arch kernels, returns `None` when the backend
    does not (yet) support arch kernels.
```

danieldk:
> thx for taking care of this! look good to me, except for the coming transition from torch xla. also wondering if you plan to add a ci with trn2 runner.

Hooking up CI would be great!

kernels/src/kernels/utils.py
```diff
@@ -89,7 +93,11 @@ def _backend() -> str:
         return "cpu"
 
 
-def _build_variant(backend: str | None) -> str:
+def _build_variant(backend: str | None) -> str | None:
+    """
+    Build vaariant for arch kernels, returns `None` when the backend
```
drbh: nit: tiny typo with `vaariant`

PR #287, opened 2026-02-17
Kernel docs followups

sayakpaul:
* Move to a fully Jinja-based template and remove `re` based parsing.
* Decouple commands into two: `kernels init-card` and `kernels fill-card`.
* Make kernel card generation a part of `kernels upload` (and test).
* Add an entry for generating kernel system cards to the "writing kernels" tutorial.

### DIFF
--- a/docs/source/cli-init.md
+++ b/docs/source/cli-init.md
@@ -15,7 +15,7 @@ This creates a new directory in the current working directory with the (normaliz
 - Downloads a project template and replaces placeholders in file names, paths, and file contents
 - Optionally restricts enabled backends (updates `build.toml` and removes unused backend folders)
 - Initializes a Git repo and stages the files (`git init`, `git add .`)
-- Initializes a trmplated system card for the kernel which is meant for the
+- Initializes a templated system card for the kernel which is meant for the
 kernel developer to further modify
 
 ## Examples


kernels/src/kernels/cli/__init__.py
```diff
@@ -344,6 +366,10 @@ def upload_kernels(args):
         branch=args.branch,
         private=args.private,
     )
+    if args.create_card:
+        if args.card_path is None:
+            args.card_path = str(Path(args.kernel_dir).resolve() / "README.md")
```
danieldk: Not for this PR, but it would be nice to make all card paths `Path` and not `str`.

kernels/src/kernels/cli/__init__.py
```diff
@@ -92,6 +92,28 @@ def main():
...
+        "--description",
+        type=str,
+        default=None,
+        help="Description to introduce the kernel (only used with --create-card).",
+    )
+    upload_parser.add_argument(
+        "--create-pr",
+        action="store_true",
+        help="If set, it will create a PR on the repo-id (only used with --create-card).",
+    )
```
danieldk: I'd leave this out for the upload command. Having dependent options is not nice if not needed and it's strange that the kernel upload would not generate a PR, but the card would.

danieldk: I was thinking more of just uploading the card, not creating one. E.g. if the user changed the card, it's nice that it gets uploaded as well along with the kernels. Presumably they already have created the card before. That means we have to settle on the name though, since I think on the Hub the card is always `README.md` and the `README.md` in a kernel Git repo may contain different information (more relevant to kernel developers/contributors).

So, maybe we should standardize on using `CARD.md` in the top-level directory of a kernel repo for the card and then `kernel upload` would upload it to the Hub as `README.md`. This is also in line with how the rest of how kernel building works - we go for very opinionated, there is only one way to do it, to enforce standardized, declarative source layouts.

I think orthogonal subcommands are the best, so ideally there would be a `create-card` subcommand that only creates a card. Then the `upload` subcommand uploads the card to the Kernel Hub along with the kernel. Maybe `upload` could have an `--only-card` option to only upload the card, but the default of `kernels upload` should do the right thing: upload the latest build variants and the latest card.

kernels/src/kernels/cli/__init__.py
```diff
@@ -346,35 +358,54 @@ def upload_kernels(args):
     )
 
 
-def create_and_upload_card(args):
-    if not args.repo_id and args.create_pr:
-        raise ValueError("`create_pr` cannot be True when `repo_id` is None.")
-
+def initialize_card(args):
     kernel_dir = Path(args.kernel_dir).resolve()
+    card_path = kernel_dir / "build" / SYSTEM_CARD_PATH
```
sayakpaul: Should we assume that `kernel_dir` has "build" directory present? Because none of the stuff related to kernel cards relies on that directory.

kernels/src/kernels/cli/__init__.py
```diff
@@ -17,13 +17,16 @@
 from kernels.cli.versions import print_kernel_versions
 from kernels.cli.doc import generate_readme_for_kernel
 from kernels.kernel_card_utils import (
```
drbh: maybe we can move the kernel card cli logic into the cli directory to align it with the other commands?

kernels/src/kernels/cli/__init__.py
drbh: > Should we assume that kernel_dir has "build" directory present? Because none of the stuff related to kernel cards relies on that directory.

no I don't think we should, currently the nix buiild process only creates this folder on the `.#build-and-copy` command. its probably best to create the CARD file in the project root

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -123,95 +123,85 @@ def _extract_functions_from_all(init_file_path: Path) -> list[str] | None:
...
-        return kernel_card
+def _extract_card_sections(card_content: str) -> dict:
+    """Extract named sections from a rendered kernel card markdown string."""
+    body = card_content
+    front_matter_match = re.match(r"^---\n.*?\n---\n", body, re.DOTALL)
+    if front_matter_match:
+        body = body[front_matter_match.end() :]
 
-    func_names = _extract_functions_from_all(init_file)
+    parts = re.split(r"\n## ", body)
```
drbh: any reason to prefer regex parsing over using jinja templates with explicit sections that will be injected into? 

I think we can avoid regex parsing all together with the placeholders that are in the `card_template.md` file. For example if we use `{{ kernel_description }}` then we can use jinja to hydrate the card in the `fill-card` command and it unambiguous as to where the generated text will be placed (for both the developer and for the command)

kernels/src/kernels/kernel_card_utils.py
drbh: if jinja is required it may be best to add it into the `pyproject.toml` file. normally we don't throw errors for individual dependencies.

the other cases where we throw id deps are missing, we add them to the `[project.optional-dependencies]` section of `pyproject.toml` and then the user can install them via `pip install kernels[optional_group]`, however in this case I think its reasonable to add jinja2 as a default dep since this seems like a default use case for the cli

kernels/src/kernels/kernel_card_utils.py
```diff
@@ -50,15 +50,15 @@ def _load_or_create_kernel_card(
...
             pass  # Will create from template below
 
     if kernel_card is None:
         kernel_description = kernel_description or DESCRIPTION
         kernel_card = ModelCard.from_template(
             card_data=ModelCardData(license=license, library_name=LIBRARY_NAME),
             template_path=str(KERNEL_CARD_TEMPLATE_PATH),
-            model_description=kernel_description,
+            modeld_description=kernel_description,
         )
```
drbh: I may not have enough context, but it's not fully clear to me if we need the `ModelCard` concept from the huggingface_hub. 

They way its used seems to essentially wraps the jinja templating functionality. And currently its only used to copy the `card_template.md` and inject two values, then the rest of the generated data is injected via regex. It may be more simple to use this to inject all of the values, or avoid the `ModelCard` all together and use jinja2 directly (especially since this code path requires the jinja2 dependency).

kernels/src/kernels/cli/__init__.py
```diff
@@ -346,35 +358,54 @@ def upload_kernels(args):
...
-
+def initialize_card(args):
     kernel_dir = Path(args.kernel_dir).resolve()
+    card_path = kernel_dir / "build" / SYSTEM_CARD_PATH
+
     kernel_card = _load_or_create_kernel_card(
-        kernel_description=args.description, license="apache-2.0"
+        repo_id_or_path=args.repo_id, license="apache-2.0"
     )
+    kernel_card.save(card_path)
```
drbh: it seems like the init command can be simplified to only copy the `card_template.md` into the current kernel dir. 

by calling `_load_or_create_kernel_card` the template is processed and it results in a blank card. 

ie.

```md
---
library_name: kernels
license: apache-2.0
---

<!-- This model card has automatically been generated. You
should probably proofread and complete it, then remove this comment. -->



## How to use



## Available functions



## Supported backends




## Benchmarks



## Source code



## Notes

```

I think we'd want to return a card with explicit sections that are contain place holders like `{{ kernel_description }}` this way its clear to the developer which sections will be injected into.

kernels/src/kernels/cli/kernel_card_utils.py
sayakpaul: Is it too bad though? Like Regex is only used to extract the header-level sections and nothing else. I think this much usage of regex is fine.

kernels/src/kernels/kernel_card_utils.py
sayakpaul: > They way its used seems to essentially wraps the jinja templating functionality. And currently its only used to copy the card_template.md and inject two values, then the rest of the generated data is injected via regex.

I don't think the rest of the generated data is injected through Regex. Please take a closer look :)
https://github.com/huggingface/kernels/blob/d6f40afc0b05dbf12a6142f57600d8c4e45dc36e/kernels/src/kernels/cli/__init__.py#L399C5-L405C6

None of this stuff is regex-injected here. `ModelCard` is already an established module from the `huggingface_hub` package, and it helps by providing programmatic interfaces for interacting with the Hub card (such as saving, loading from the Hub, accessing metadata, actual content, etc.).

Then we have:
https://github.com/huggingface/kernels/blob/d6f40afc0b05dbf12a6142f57600d8c4e45dc36e/kernels/src/kernels/kernel_card_utils.py#L153

Only the `_extract_functions_from_all` does regex stuff. Other than that, I think all the other things are parsed from the directory structure and the `build.toml` file.

In this light, could you take another look and LMK your thoughts?

drbh: thanks for the info. in terms of saving and loading from the hub, I was under the impression we were going to rely on using kernel upload to push to the hub, and we'd load the card from the copy we have in source in `kernels/src/kernels/cli/card_template.md` (or if it already exists it would be cloned with the rest of the repo source). 

and ahh great point, I totally made a mistake when I said "injected" I see that the variables are injected in the `ModelCard.from_template` call. my intention was to point out that we are parsing the file via regex, then using the parsed values as input when reconstructing the file. 

It seems like we can avoid the need to parse the file all together if we just leave the jinja template placeholders like `{{ kernel_description }}` and let jinja/model card only replace those sections. This way we wouldn't have to capture the existing text and use that in the `ModelCard.from_template` call.  my concern is that we can run into issue when trying to parse existing text.

### example

if I run 

```bash
kernels init-card --repo_id drbh/test-kernels .
```

then add a note under one of the sections like `## Supported backends` 

like this
```md
## Supported backends

will add ABD support in the future

```

and then run 

```bash
kernels fill-card --repo_id drbh/test-kernels .
```

it removes the user added comment completely. 

outputs:

```md
## Supported backends

- metal

```

I believe this happens because this logic

```python
    updated_card = ModelCard.from_template(
        card_data=existing_card.data,
        template_path=str(KERNEL_CARD_TEMPLATE_PATH),
        kernel_description=description,
        **preserved,
        **dynamic_vars,
    )
```

overwrites the `preserved` (parsed) data with the `dynamic_vars` inputs when they both refer to the same variable.

it seems like it would be less error prone to keep the `{{ supported_backends }}` in the card template that the user updates

like this:
```md
## Supported backends

{{ supported_backends }}
```

so a user could add notes like this
```md
## Supported backends

will add ABD support in the future

{{ supported_backends }}
```

and then 

```bash
kernels fill-card --repo_id drbh/test-kernels .
```

would result in 

```md
## Supported backends

will add ABD support in the future

- metal
```

following a pattern like this would remove the need to parse the file via regex, and would avoid reconstructing the whole file from a combination of regex parsed text and jinja variables, and has the benefit of clearly showing which sections of the card will be replaced.

kernels/src/kernels/cli/kernel_card_utils.py
drbh: I think my concern is mainly that we may not need to parse the file and can avoid rebuilding the file from parsed text. I added a couple more details and an example in the comment below

kernels/src/kernels/kernel_card_utils.py
sayakpaul: @drbh do the recent changes adhere to what you had in mind?

kernels/src/kernels/cli/kernel_card_utils.py
sayakpaul: All removed now.

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -0,0 +1,149 @@
...
+        kernel_configs = config.get("kernel", {})
+        cuda_capabilities: set[Any] = set()
+        for k in kernel_configs:
+            caps = kernel_configs[k].get("cuda-capabilities")
+            if caps:
+                cuda_capabilities.update(caps)
+        if cuda_capabilities:
+            vars["cuda_capabilities"] = "\n".join(
+                f"- {cap}" for cap in cuda_capabilities
+            )
```
sayakpaul, line 130: Based on https://huggingface.slack.com/archives/C090JN2P8NB/p1772345791192429, I think we're including CUDA capabilities in the `build.toml` for special cases (like FP8) for now? 

Once it's a part of `metadata.json`, we can modify this code to parse the capability info from there. WDYT?

kernels/src/kernels/cli/__init__.py
```diff
@@ -346,35 +359,44 @@ def upload_kernels(args):
...
-    )
-    updated_card = _update_benchmark(kernel_card=kernel_card, local_path=kernel_dir)
-    updated_card = _update_kernel_card_license(
-        kernel_card=kernel_card, local_path=kernel_dir
+    card_path = kernel_dir / SYSTEM_CARD_PATH
+
+    dynamic_vars = _build_kernel_card_vars(
+        kernel_dir, repo_id=args.repo_id or "REPO_ID"
     )
+    description = args.description or DESCRIPTION
```
drbh:
```suggestion
    repo_id = args.repo_id or "REPO_ID"
    dynamic_vars = _build_kernel_card_vars(kernel_dir, repo_id=repo_id)
    description = (args.description or DESCRIPTION).format(repo_id=repo_id)
```

kernels/src/kernels/cli/kernel_card_utils.py
drbh: sounds good to me

kernels/tests/test_status.py
```diff
@@ -78,7 +81,9 @@ def test_redirect_with_revision(self, tmp_path):
         from huggingface_hub.utils import EntryNotFoundError
 
         status_file = tmp_path / "kernel-status.toml"
-        status_file.write_text('kind = "redirect"\ndestination = "kernels-community/new-kernel"\nrevision = "v2"')
+        status_file.write_text(
+            'kind = "redirect"\ndestination = "kernels-community/new-kernel"\nrevision = "v2"'
+        )
```
sayakpaul, line 86: Result of running `make style`.

drbh (dismissed):
nice! lgtm

worked when tested locally

sayakpaul:
Thanks @drbh! Thanks for jamming with me on this one. 

I will leave it open for @danieldk's comments, too.

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -0,0 +1,160 @@
...
```
danieldk, line 130: We should read the capabilities from `metadata.json` as discussed, but I think it can be a follow-up PR.

kernels/src/kernels/cli/__init__.py
```diff
@@ -348,34 +362,44 @@ def upload_kernels(args):
...
+
+    card_data = ModelCardData(library_name=LIBRARY_NAME)
+    updated_card = ModelCard.from_template(
+        card_data=card_data,
+        template_path=str(card_path),
+        kernel_description=description,
+        **dynamic_vars,
+    )
+    _update_kernel_card_license(updated_card, kernel_dir)
+    updated_card.save(card_path)
```
danieldk, line 365: My idea was that the filled card should not overwrite the card in the repo. The card in the repo will always be a template, otherwise it can be out-of-sync with `build.toml` and the builds (for instance, when kernel-builder gets updated and adds a new Torch version), we should treat it as source code.

The fill-card command would accept an output file and then the builder uses it to write the card to `build/CARD.md` or something like that.

kernels/src/kernels/cli/__init__.py
```diff
@@ -247,36 +252,45 @@ def main():
...
-        help="If specified it will be pushed to a repository on the Hub.",
-    )
-    repocard_parser.add_argument(
-        "--create-pr",
-        action="store_true",
-        help="If specified it will create a PR on the `repo_id`.",
+        help="When specified, the example code will incorporate this `repo_id` to create a custom example snippet.",
     )
-    repocard_parser.set_defaults(func=create_and_upload_card)
+    fill_card_parser.set_defaults(func=fill_kernel_card)
```
danieldk: I don't think we need this option? The `repo_id` should be put in `build.toml`.

docs/source/cli-card.md
```diff
@@ -0,0 +1,6 @@
+### kernels card
+
+Use `kernels init-card kernel_dir`, to either instantiate a
+fresh kernel card or load an existing one.
+
+To fill the card, run `kernels fill-card kernel_dir`.
```
danieldk: I don't think we should document `fill-card`, it is part of the build and should only be used by the builder.

I think we should move move `fill-card` to the builder eventually, probably as small standalone Rust utility to properly distribute responsibility, `kernels` for client stuff, `builder` for build stuff.

But we can discuss and do this in a subsequent PR if we decide to do it.

docs/source/builder/writing-kernels.md
```diff
@@ -330,3 +330,16 @@ $ nix run .#ciTests.torch210-cxx11-cpu-x86_64-linux
...
+We provide a utility to generate a system card for a given kernel, utilizing
+information from its `build.toml` and metadata. This system card provides a
+reasonable starting point and is meant to be edited afterward by the kernel
+developer.
+
+To generate the card, first run `kernels init-card`, which will either
+instantiate a fresh kernel card or load an existing one. Then run
+`kernels fill-card` to update the card based on the build information.
+The existing user content is preserved. Refer to the [documentation](../cli.md)
+to know more.
```
danieldk:
```suggestion
To generate the card, first run `kernels init-card`, which will either
instantiate a fresh kernel card or load an existing one. The card will
be filled automatically by the builder when using the `build-and-upload`
or `build-and-copy` command.
```

kernels/src/kernels/cli/kernel_card_utils.py
sayakpaul: Yes, once that is merged, I will do it in a separate PR.

kernels/src/kernels/cli/__init__.py
sayakpaul: > the card in the repo

Which repo is being referred to here? The build repo must upload the card as a README.md to the root of it, otherwise it won't generate nicely on the Hub (which I believe is the purpose of the card)?

kernels/src/kernels/cli/__init__.py
sayakpaul: In case, it's absent, we will default to "none", then (which is fine IMO).

docs/source/cli-card.md
sayakpaul: I like the idea, but `kernels init-card ...` just gives us a template which is not that useful anyway. So, how do we get people to give importance to documenting their interfaces with a standard (which, IMO, should be treated with similar importance as developing the kernels)?

sayakpaul: (Edit): next suggestion makes sense!

kernels/src/kernels/cli/kernel_card_utils.py
sayakpaul: Added a note to deal with this later.

kernels/src/kernels/cli/__init__.py
sayakpaul: Resolved.

sayakpaul:
@danieldk I have some questions.

I did remove `repo_id` from `fill-card`. But how do we load an existing card from the Hub if `repo_id` isn't present in either `init-card` (that is what I am getting at after https://github.com/huggingface/kernels/pull/287#discussion_r2904928432) or `fill-card`?

If we were to just serialize the card template in the source repo (e.g., https://github.com/huggingface/kernels-community/), then `repo_id` would need to be removed from there (which is fine). However, if we also remove `repo_id` from `fill-card`, then there is no way for us to load an existing card from the Hub.

If we add `repo_id` back in `fill-card`, then I am guessing we'd only want to use it to load an existing card?

In any case, information that depends on build or metadata will be updated each time a new build is obtained and the `fill-card` utility is called (when needed). So, even if we're operating on an already loaded card, user-content won't be affected.

kernels/tests/test_layer.py
```diff
@@ -102,7 +102,7 @@ def __init__(self):
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         self.n_calls += 1
-        d = input.shape[-1] // 2
+        _ = input.shape[-1] // 2
```
sayakpaul: `make style`

kernels/src/kernels/cli/__init__.py
danieldk: The GitHub repo.

docs/source/cli-card.md
danieldk: > I like the idea, but `kernels init-card ...` just gives us a template which is not that useful anyway.

It is, because people can add additional things to the template, like acknowledgments, a longer introduction, their own instructions, etc.

sayakpaul:
I have updated the PR in accordance with https://github.com/huggingface/kernels/pull/287#issuecomment-4030311697. 

@drbh @danieldk PTAL when you have time.

docs/source/cli-init.md
```diff
@@ -15,6 +15,8 @@ This creates a new directory in the current working directory with the (normaliz
 - Downloads a project template and replaces placeholders in file names, paths, and file contents
 - Optionally restricts enabled backends (updates `build.toml` and removes unused backend folders)
 - Initializes a Git repo and stages the files (`git init`, `git add .`)
+- Initializes a trmplated system card for the kernel which is meant for the
```
danieldk:
```suggestion
- Initializes a templated system card for the kernel which is meant for the
```

PR #290, opened 2026-02-17
feat: enforce strict kernel name

drbh:
This PR enforces that kernel names:

- Start with a lowercase letter (a-z)
- Contain only lowercase letters, digits, and dashes
- End with a lowercase letter or digit
- Are at least 2 characters long

### DIFF
--- a/build2cmake/src/torch/noarch.rs
+++ b/build2cmake/src/torch/noarch.rs
@@ -24,7 +24,12 @@ pub fn write_torch_ext_noarch(
     let ops_name = kernel_ops_identifier(&target_dir, &build.general.name.python_name(), ops_id);
 
     write_compat_py(&mut file_set)?;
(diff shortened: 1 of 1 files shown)


build2cmake/src/main.rs
```diff
@@ -206,9 +207,29 @@ fn parse_and_validate(build_toml: impl AsRef<Path>) -> Result<BuildCompat> {
...
+             - Contain only lowercase letters, digits, and dashes\n\
+             - End with a lowercase letter or digit\n\
+             - Be at least 2 characters long\n\
+             Examples: `my-kernel`, `relu2d`, `flash-attention`"
+        );
+    }
+
+    Ok(())
+}
+
```
danieldk: Really nice! I like strong typing, maybe we could have a `KernelName` type where the constructor/deserializer does the validation? In that way we can guarantee that it's validated everywhere.

drbh: great idea, will update shortly

PR #293, opened 2026-02-19
feat: support redirection

drbh:
This PR adds a concept of redirection via a `REDIRECT.toml` file on the `main` branch.

The kernel tool will now first check for a redirection file, read and apply the redirection if it exists. The core use case is enable a way to manually redirect users from old repos to newer kernel repos with a custom message. 

Note: redirections are single hop only, its not possible to redirect to another redirect

example file 
```toml
destination = "kernels-community/activation"
message = "This kernel has been moved to kernels-community/activation. Please update your dependencies to use the new location."
```
this example file is currently hosted https://huggingface.co/drbh/test-kernel-redirect

### DIFF
--- a/kernels/src/kernels/status.py
+++ b/kernels/src/kernels/status.py
@@ -31,9 +31,8 @@ class Redirect:
 KernelStatusKind = Union[Redirect]
 
 
-@dataclass
 class KernelStatus:
-
+    @staticmethod
     def from_toml(content: str) -> KernelStatusKind:
         data = tomllib.loads(content)
 
@@ -47,6 +46,7 @@ class KernelStatus:
         raise ValueError(f"Unknown kernel status kind: {kind!r}")
 
     # Fetch the kernel status from the repository, if it exists
+    @staticmethod
     def check_status(
         api: HfApi, repo_id: str, revision: str
     ) -> KernelStatusKind | None:


kernels/src/kernels/redirect.py
```diff
@@ -0,0 +1,56 @@
+import logging
+import warnings
+from dataclasses import dataclass
+
+from huggingface_hub import HfApi
+from huggingface_hub.utils import EntryNotFoundError
+
+from kernels.compat import tomllib
+
+REDIRECT_FILENAME = "REDIRECT.toml"
```
danieldk: I think we should have a more generic name than that, so that we can use the same file for multiple signaling purposes.

Maybe `kernel-status.toml`?

kernels/src/kernels/redirect.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@dataclass
+class RedirectInfo:
+    destination: str
+    message: str
```
danieldk: Redirect should be one of a union (so that we can also have e.g. deprecated/unmaintained in the future).

kernels/src/kernels/redirect.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def parse_redirect_file(content: str) -> RedirectInfo:
```
danieldk: I think we should have this similar to https://github.com/huggingface/kernels/blob/main/kernels/src/kernels/metadata.py where these are methods.

Maybe we should consider switching to pydantic for these deserialization/ validation purposes.

kernels/src/kernels/redirect.py
```diff
@@ -0,0 +1,56 @@
...
+    if redirect is None:
+        return repo_id, revision
+
+    msg = f"WARNING: '{repo_id}' redirected to '{redirect.destination}'"
+    if redirect.message:
+        msg = f"WARNING: {redirect.message}"
+
+    RED_COLOR = "\033[91m"
+    RESET_COLOR = "\033[0m"
+    msg = f"\n{RED_COLOR}{msg}{RESET_COLOR}"
```
danieldk: Do these work well in Windows shells, I vaguely remember ASCII sequences giving issues and you need a library to make them work on *nix and Windows.

kernels/src/kernels/redirect.py
```diff
@@ -0,0 +1,56 @@
...
+
+    warnings.warn(msg, UserWarning)
+
+    if "@" in redirect.destination:
+        return redirect.destination.rsplit("@", 1)
+    return redirect.destination, "main"
```
danieldk: `@` in the string data seems very ad-hoc if we have a structured data format like TOML?

kernels/src/kernels/utils.py
```diff
@@ -207,6 +208,7 @@ def install_kernel(
     backend: str | None = None,
     variant_locks: dict[str, VariantLock] | None = None,
     user_agent: str | dict | None = None,
+    follow_redirects: bool = True,
```
danieldk: I would leave this out. We already have plenty of arguments and we are probably going to add `framework` soon. I think we should just go for doing the right thing by default. If a kernel developer sets a repo to redirect, they probably _really_ want to redirect.

kernels/src/kernels/redirect.py
danieldk: I would leave out a message for now.

kernels/src/kernels/redirect.py
drbh: good point, updated in latest and renamed the file to `status` rather than `redirect`

kernels/src/kernels/redirect.py
drbh: this makes sense, I've updated in latest for the following

```python
@dataclass
class Redirect:
    kind: str  # must be "redirect"
    destination: str
    revision: str

    ...
```

kernels/src/kernels/redirect.py
drbh: good point, I've updated to follow a pattern like metadata, and added a `kind` key so we know which type to deserialize the file to.

pydantic could be nice to add, however i've left that out of this PR and can be followed up in the future to clean things up

kernels/src/kernels/redirect.py
drbh: hmm I did a bit of research and I think they should work on any windows version `>=10` however I've removed for now to simplify

kernels/src/kernels/redirect.py
drbh: good point, `revision` is now explicitly on the `Redirect` class now

kernels/src/kernels/utils.py
drbh: agreed, updated in latest

kernels/src/kernels/redirect.py
drbh: updated in latest

PR #300, opened 2026-02-27
feat: upload in multiple commits when many files

drbh:
This PR adds a path to use the `upload_large_folder` api when there are more than 200 files in the build output. This helps avoid timeouts when many files are in the build. Otherwise the normal `upload_folder` is preferred since it has a bit more flexibility around `delete_patterns` and the `commit_message`

### DIFF
--- a/kernels/src/kernels/cli/upload.py
+++ b/kernels/src/kernels/cli/upload.py
@@ -35,7 +35,7 @@ def _upload_build_dir(
         f"build/{variant.relative_to(build_dir).as_posix()}/" for variant in variants
     )
     delete_prefixes = ("build/",) if is_new_branch else variant_prefixes
-    operations = [
+    operations: list[CommitOperationAdd | CommitOperationDelete] = [
         CommitOperationDelete(path_in_repo=repo_file)
         for repo_file in sorted(
             api.list_repo_files(repo_id=repo_id, revision=revision, repo_type="model")


sayakpaul (dismissed):
Should we add a test as well? No strong opinions, of course.

kernels/src/kernels/cli/upload.py
```diff
@@ -70,13 +70,33 @@ def upload_kernels_dir(
...
+            f"⚠️  Found {file_count} files to upload, which exceeds the 200 file limit for a single commit. Deleting old build files and re-uploading the whole build folder to avoid hitting file limits."
+        )
+        kernel_root_dir = build_dir.parent
+        api.upload_large_folder(
+            repo_id=repo_id,
+            folder_path=kernel_root_dir,
+            revision=branch,
+            repo_type="model",
+            allow_patterns=["build/torch*"],
+        )
```
danieldk: We should not just remove old build files, since it can break our version contract. Also, if we automatically switch to a different upload type, the behavior should be _exactly_ the same.

Is it possible to add `delete_patterns` to `upload_large_folder`?

kernels/src/kernels/cli/upload.py
```diff
@@ -70,13 +70,33 @@ def upload_kernels_dir(
...
-        delete_patterns=list(delete_patterns),
-        commit_message="Build uploaded using `kernels`.",
-        allow_patterns=["torch*"],
+    file_count = sum(
+        1
+        for p in build_dir.rglob("*")
+        if p.is_file() and p.relative_to(build_dir).as_posix().startswith("torch")
     )
+
+    if file_count > 200:
```
danieldk: Docs:

> When dealing with a large folder (thousands of files or hundreds of GB), we recommend using [upload_large_folder()](https://huggingface.co/docs/huggingface_hub/v1.5.0/en/package_reference/hf_api#huggingface_hub.HfApi.upload_large_folder) instead.

kernels/src/kernels/cli/upload.py
```diff
@@ -70,13 +70,33 @@ def upload_kernels_dir(
...
 
-    api.upload_folder(
-        repo_id=repo_id,
-        folder_path=build_dir,
-        revision=branch,
-        path_in_repo="build",
-        delete_patterns=list(delete_patterns),
-        commit_message="Build uploaded using `kernels`.",
-        allow_patterns=["torch*"],
+    file_count = sum(
```
danieldk: I think this should be a separate function.

kernels/src/kernels/cli/upload.py
drbh: unfortunately  upload_large_folder does not contain delete_patterns.  and I actually misunderstood the upload functionality and added an incorrect comment. The upload_large_folder function doesn't seem to delete any files. It upserts the files that match the allow pattern. 

I fully agree they should act exactly the same, however don't we want to delete the older build within a version? For example my understanding is the `upload_folder` path deletes old files like 

```
{'torch210-cxx11-cu130-x86_64-linux/**', '.cache/**', 'torch210-cxx11-cu126-x86_64-linux/**', 'torch29-cxx11-cu126-x86_64-linux/**', 'torch29-cxx11-cu128-x86_64-linux/**', 'torch210-cxx11-cu128-x86_64-linux/**', 'torch29-cxx11-cu130-x86_64-linux/**'}
```

so the old `.so` is removed and replaced with the new `.so` (since they will have unique suffixes). I think to fulfill this logic with the `upload_large_folder` there will need to be an explicit delete call made before or after the large folder upload.

another possible option is to avoid `upload_large_folder` all together and in the case of many files, we either retry the upload command a couple times (this works in practice) or we iterate over each build variant and upload each in its own commit (but this may still suffer if a single build variant is too large)

what do you think is the best approach?

kernels/src/kernels/cli/upload.py
drbh: good point, I've bumped the number to 1000 files in the latest commit

originally just arbitrarily followed the number in huggingface_hub that promoted the text to a warning https://github.com/huggingface/huggingface_hub/blob/ed43eb67f2c42b8ce14114a867d7ece012bd2a79/src/huggingface_hub/hf_api.py#L9837-L9846

kernels/src/kernels/cli/upload.py
drbh: agreed, I've moved this logic to a new function in the latest commit

kernels/src/kernels/cli/upload.py
```diff
@@ -70,13 +77,29 @@ def upload_kernels_dir(
...
-        folder_path=build_dir,
-        revision=branch,
-        path_in_repo="build",
-        delete_patterns=list(delete_patterns),
-        commit_message="Build uploaded using `kernels`.",
-        allow_patterns=["torch*"],
-    )
+    file_count = get_file_count_in_build(build_dir)
+
+    if file_count > 1_000:
```
sayakpaul: (nit): 1_000 could be made a constant.

kernels/tests/test_kernel_upload.py
```diff
@@ -120,3 +122,27 @@ def test_kernel_upload_deletes_as_expected():
         str(filename_to_change) in k for k in repo_filenames
     ), f"{repo_filenames=}"
     _get_hf_api().delete_repo(repo_id=REPO_ID)
+
+
+def test_large_kernel_upload_uses_kernel_root_path(monkeypatch, tmp_path):
```
sayakpaul: Test is perfect!

kernels/src/kernels/cli/upload.py
```diff
@@ -70,13 +77,29 @@ def upload_kernels_dir(
...
+        print(
+            f"⚠️  Found {file_count} files to upload, which exceeds the 1,000 file limit for a single commit."
+        )
+        kernel_root_dir = build_dir.parent
```
sayakpaul: We can call `delete_files()` here to maintain parity with the existing behaviour of `upload_kernels()`.

sayakpaul (commented):
Left some more comments. LMK if they're clear.

kernels/src/kernels/cli/upload.py
danieldk: Yes, we need two different behaviors:

* We need to if a variant `torch210-cxx11-cu130-x86_64-linux` is uploaded, we need to remove all stale files like you mention, e.g. the old `.so` file.
* At the same time, if we e.g. upload only `torch210-cxx11-cu130-x86_64-linux` and `torch210-cxx11-cu126-x86_64-linux`, we should not remove `torch29-cxx11-cu126-x86_64-linux`.

So the logic is: remove stale files of build variants that are uploaded, but do not remove any files of build variants that are not uploaded.

> I think to fulfill this logic with the upload_large_folder there will need to be an explicit delete call made before or after the large folder upload.

But I think deletes should be atomic or at least after the large upload to avoid having an undownloadable kernel if upload fails.

For now, for large folders doing one upload per build variant sounds like the best option? We guarantee the existing semantics and it might be enough for most cases?

sayakpaul: > For now, for large folders doing one upload per build variant sounds like the best option? We guarantee the existing semantics and it might be enough for most cases?

Good point! I agree this should cut it for now. If anything more advanced arises, we can also discuss a bit with the Hub team to figure out what's best.

julien-c: (cc @Wauplin just in case for viz)

Wauplin: Thanks for the ping! General note, never hesitate to ping me or @hanouticelina on such topics either in Github or directly on Slack so we can either advice on a solution or adapt `huggingface_hub` if some inconsistencies can be resolved in it.

Wauplin: Regarding this specific case, the "why" upload_large_folder do not have `delete_patterns` it's because the method is meant for heavy uploads, with a resumable and multi-commit process. Primary purpose is for example to upload very large datasets _once_, not necessarily as an automated process. In such cases, deleting files as part of the upload is not really important (repo author can always clean-up their files afterwards). If integrating in a more automated workflow -typically in this PR-, then we would change a bit the API I think.

In your very specific case, I think the best is to use the low-level API `create_commit` with a list of `CommitOperationAdd` and `CommitOperationDelete`. Something like this should work:

```py
# build list of delete ops from previous build
delete_operations = [CommitOperationDelete(...) for file in list_repo_files(...) if file ...]

# build list of add ops
files_to_add = [...]  # to implement
add_operations = [CommitOperationAdd(...) for file in files_to_add]

# commit by batch of 200 ops (nb to tweak?)
for chunk_operations in chunk_iterator(delete_operations + add_operations, n=200):
    create_commit(...operations=chunk_operations)
```

This is a bit more work but you have full flexibility and it works well. And you can adapt the commit message as you want. The good part is that you have exactly the same logic no matter the number of ops (i.e. no "if ... then use upload_folder else use upload_large_folder").

Let me know what you think!

kernels/src/kernels/cli/upload.py
```diff
@@ -70,13 +77,29 @@ def upload_kernels_dir(
...
+        api.upload_large_folder(
+            repo_id=repo_id,
+            folder_path=kernel_root_dir,
+            revision=branch,
+            repo_type="model",
+            allow_patterns=["build/torch*"],
+        )
+    else:
```
Wauplin: see my comment above to avoid having 2 different upload logic https://github.com/huggingface/kernels/pull/300#discussion_r2903723010

kernels/src/kernels/cli/upload.py
drbh: I really like this approach, thanks for the tip! I've updated to follow this pattern in the latest commits

kernels/src/kernels/cli/upload.py
```diff
@@ -1,16 +1,80 @@
...
+
+    if not operations:
+        return
+
+    batch_count = (
+        len(operations) + BUILD_COMMIT_BATCH_SIZE - 1
+    ) // BUILD_COMMIT_BATCH_SIZE
+    if batch_count > 1:
+        print(
+            f"⚠️  Found {len(operations)} build operations, uploading in {batch_count} commits."
```
sayakpaul, line 58: Should this use `logging` instead?

kernels/src/kernels/cli/upload.py
```diff
@@ -86,13 +140,13 @@ def upload_kernels_dir(
             allow_patterns=["benchmark*.py"],
         )
 
-    api.upload_folder(
+    assert variants is not None
```
sayakpaul, line 143: Do we want to accompany this with an informative error message?

kernels/tests/test_kernel_upload.py
```diff
@@ -168,3 +172,108 @@ def test_kernel_upload_deletes_as_expected():
...
+        "benchmarks/benchmark_old.py",
+        "build/torch-cpu/stale.py",
+        "build/torch-cuda/inherited.py",
+    ]
+    monkeypatch.setattr("kernels.cli.upload._get_hf_api", lambda: api)
+
+    upload_kernels(UploadArgs(kernel_root, REPO_ID, False, "v2"))
+
+    api.upload_folder.assert_called_once()
+    assert api.upload_folder.call_args.kwargs["delete_patterns"] == ["**"]
```
sayakpaul, line 266: Is it because for the benchmarks, we rely on `upload_folder()`?

sayakpaul (commented):
Cool work! I see that we're relying on the commit ops in this PR. Could I get a gist of what the rationale was behind it?

(edit: nevermind https://github.com/huggingface/kernels/pull/300#discussion_r2903723010)

kernels/src/kernels/cli/upload.py
drbh: maybe, but in this case I think print is okay since the messages are outputs from the cli command directly to the user, as opposed to a log from an internal library. additionally this follow the same pattern as some of the other user facing messages like the success message. 

happy to revisit in the future in another PR that cleans up the repo, but will avoid adding the looger to this file for now and keep the changes focused on the new commit logic

kernels/src/kernels/cli/upload.py
drbh: theoretically this will never fire and was just added to make the type checker happy. the logic above (copied) should fires the error in the case that variants's were None before this assert is reached. 

```python
    build_dir = None
    variants = None
    for candidate in [kernel_dir / "build", kernel_dir]:
        variants = [
            variant_path
            for variant_path in chain(
                candidate.glob("torch*"), candidate.glob("tvm-ffi*")
            )
            if BUILD_VARIANT_REGEX.match(variant_path.name) is not None
            and (variant_path / "metadata.json").is_file()
        ]
        if variants:
            build_dir = candidate
            break
    if build_dir is None:
        raise ValueError(
            f"Couldn't find any build variants in: {kernel_dir.absolute()} or {(kernel_dir / 'build').absolute()}"
        )
```

however I think maybe the overall logic could be cleaned up to avoid the need for an assert, but would like to keep this PR focused on the commit changes and can revisit a refactor soon. Also in that refactor it may be nice to check out `ty` instead of `mypy`

kernels/tests/test_kernel_upload.py
drbh: yea currently the benchmarks are uploaded separately from the build files via `upload_folder`, however they can be rolled into the new commit logic. I also think this might be best as a follow up PR. I mainly want to keep this one focused on replacing the build logic with the new api and happy to clean this up more in another PR

PR #303, opened 2026-03-02
add terraform scripts

### LINKED ISSUE #236: Terraform scripts for setting up instances for building, development, etc.
Nix setup can be painful. Automating the whole process through Terraform will help us to establish structure as well as reproducibility. It will also allow for flexible configurations.

sayakpaul:
Resolves #236 

HF folks should consult https://www.notion.so/huggingface2/Terraform-scripts-for-kernels-3141384ebcac809589c6e2df6108f4de to refer to the session-specific security instructions.

Currently, `terraform apply` would create an `m7i.8xlarge` NixOS instance (32 vCPUs, 128 GiB RAM) is provisioned with a 200 GiB encrypted root volume and a 1 TiB encrypted gp3 data
volume mounted at `/data`. On first boot, the bundled NixOS configuration
is applied via `nixos-rebuild switch` and the Hugging Face Cachix binary
cache is registered. 

Screenshot 2026-03-02 at 3 56 36 PM

Screenshot 2026-03-02 at 4 01 27 PM

Will update the docs after a review.

Cc: @drbh @danieldk

### DIFF
--- a/build2cmake/src/templates/noarch/setup.py
+++ b/build2cmake/src/templates/noarch/setup.py
@@ -87,7 +87,7 @@ class BuildKernel(build):
             raise ValueError(
                 f"Metadata file {metadata_file} does not exist, run build2cmake to create it"
             )
-        shutil.copy2(metadata_file, variant_dir / f"metadata-{backend}.json")
+        shutil.copy2(metadata_file, variant_dir / "metadata.json")
 
 
 if __name__ == "__main__":


terraform/nixos-configuration.nix
```diff
@@ -0,0 +1,192 @@
...
+  # -------------------------------------------------------------------------
+  environment.systemPackages = with pkgs; [
+    # Build toolchain
+    gcc
+    clang
+    cmake
+    ninja
+    pkg-config
+    gnumake
+    llvm
```
danieldk: These should not be needed, `nix develop` in a kernel directory should give you the right environment.

terraform/nixos-configuration.nix
```diff
@@ -0,0 +1,192 @@
...
+
+    # Python (used for CI validation scripts and kernel tests)
+    python3
+    python3Packages.pip
+    python3Packages.virtualenv
```
danieldk: Same here `nix develop`. Global state for development is not great, since they may be incompatible versions.

terraform/nixos-configuration.nix
```diff
@@ -0,0 +1,192 @@
...
+
+  # -------------------------------------------------------------------------
+  # Shell environment
+  # -------------------------------------------------------------------------
+
+  # direnv hooks for bash and zsh so `nix develop` shells activate automatically.
+  programs.direnv = {
+    enable = true;
+    nix-direnv.enable = true;
+  };
```
danieldk, line 151: Yes 🚀, everybody needs direnv in their lives :).

terraform/nixos-configuration.nix
```diff
@@ -0,0 +1,192 @@
...
+
+  # Useful shell aliases for kernel dev workflow.
+  environment.shellAliases = {
+    nbd = "nix build -L"; # build with logs
```
danieldk, line 155: Maybe something like

```
  dinit = "echo 'use nix' > .envrc ; direnv allow";
```

(untested, not sure if the quoting is

terraform/nixos-configuration.nix
```diff
@@ -0,0 +1,192 @@
...
+  #   substituters = huggingface cachix
+  # -------------------------------------------------------------------------
+  nix.settings = {
+    experimental-features = [
+      "nix-command"
+      "flakes"
+    ];
+    # Use every available core; individual flakes can override with their own settings.
+    max-jobs = "auto";
+    cores = 0;
```
danieldk: Hmmm, this often makes machines run out of memory with larger kernels. Maybe:

```
    cores = 16;
    max-jobs = 4;
```

terraform/nixos-configuration.nix
```diff
@@ -0,0 +1,192 @@
...
+    python3Packages.virtualenv
+
+    # Version control & productivity
+    git
+    git-lfs
+    curl
+    wget
+    jq
+    ripgrep
+    htop
```
danieldk, line 119: If we are adding tops, `btop` as well :). Maybe also zellij (as an alternative to `tmux`)?

terraform/terraform.tfvars.example
```diff
@@ -0,0 +1,20 @@
...
+# subnet_id              = "subnet-0123456789abcdef0"  # subnet used when launching manually
+# security_group_id      = "sg-0123456789abcdef0"      # existing SG that allows SSH (port 22)
+# key_pair_name          = "my-key-pair"               # existing EC2 key pair name
+# ssh_private_key_path   = "~/.ssh/id_rsa"             # local path to the corresponding private key (".pem"), 
+# this should correspond to the one being specified to `key_pair_name`
+
+# Optional overrides (all have sensible defaults):
+# instance_name          = "kernels-dev"
+# aws_region             = "us-east-1"
+# nixos_channel          = "25.11"         # or "24.11" for the stable LTS channel
```
danieldk:
```suggestion
# nixos_channel          = "25.11"
```

24.11 is not maintained, there is no LTS.

terraform/README.md
```diff
@@ -0,0 +1,105 @@
...
+```bash
+journalctl -u amazon-init -f
+```
+
+The setup is complete when the service reaches `Finished` state, which you can confirm with:
+
+```bash
+systemctl status amazon-init
+```
+
```
drbh, line 65: nit: this process was much faster than 10 minutes for me, unsure if some cache was warmed but maybe we can lower the expected time `~a couple minutes max`?

terraform/README.md
```diff
@@ -0,0 +1,105 @@
...
+Once done, reload the shell to pick up the aliases and settings:
+
+```bash
+exec bash
+```
+
+### Inside the VM
+
+Once the setup is complete, a few useful aliases are available:
```
drbh, line 76: nit: maybe we can have an example of pulling the kernels-community and building a kernel? or maybe using the kernel cli to init a new kernel and build it? just so the example is end to end?

terraform/README.md
```diff
@@ -0,0 +1,105 @@
...
+
```
drbh, line 77: nit: another thought is maybe we can provide simple instructions to connect to the device with an idea like vscode and/or zed?

terraform/README.md
sayakpaul: Very good idea. Would you like to open a PR to this branch?

terraform/README.md
sayakpaul: Since we can already SSH, maybe we can link to the remote server extension docs for VSCode so that users know that this is a possibility?

terraform/README.md
drbh: sure, however it might be best to include in a follow up PR along with an update to the docs. I'll play with this a bit more this/next week and will have a better idea of any examples/docs that should be included

terraform/README.md
drbh: yea sounds good to me, I'll include this in the docs update (however if I remember correctly theres some small gotcha with using vscode... will make sure to add this in the next PR)

PR #304, opened 2026-03-02
build2cmake: add GPU archs to metadata.json

danieldk:


build2cmake/src/templates/build-variants.cmake
```diff
@@ -117,10 +117,13 @@ endfunction()
...
 #   BUILD_VARIANT_NAME - Build variant name (e.g., "torch271-cxx11-cu124-x86_64-linux")
 #   INSTALL_PREFIX - Base installation directory (defaults to CMAKE_INSTALL_PREFIX)
+#   GPU_ARCHS - List of GPU architectures that were compiled
+#               (optional; when provided for CUDA/ROCm, metadata.json will include
+#               a "backend" key with the type and arch list)
 #
 function(add_kernels_install_target TARGET_NAME PACKAGE_NAME BUILD_VARIANT_NAME)
     set(oneValueArgs INSTALL_PREFIX)
-    set(multiValueArgs DATA_EXTENSIONS)
+    set(multiValueArgs DATA_EXTENSIONS GPU_ARCHS)
```
sayakpaul, line 126: Should there be any post-validation here on the derived capabilities? My comment is based on one of your earlier comments ([internal](https://huggingface.slack.com/archives/C083BS3K2F9/p1772308315259519?thread_ts=1772215901.984289&cid=C083BS3K2F9) comment):

> We intersect the capabilities in build.toml with the actual capabilities supported by a CUDA version. So, the set of supported capabilities may be smaller.

sayakpaul (commented):
Just one clarification question. Also, do we need to test this?

build2cmake/src/templates/build-variants.cmake
danieldk: The list of archs passed through `GPU_ARCHS` come from the intersection.

PR #308, opened 2026-03-04
Fix Metal MPS encoder lifecycle and broaden macOS compatibility

### LINKED ISSUE #307: Metal example uses unsafe MPS encoder pattern that crashes on sequential calls
## Problem

The Metal ReLU example (`examples/relu/relu_metal/relu.mm`) creates compute command encoders directly on the command buffer:

```objc
id<MTLComputeCommandEncoder> computeEncoder =
    [commandBuffer computeCommandEncoder];
```

This bypasses PyTorch's MPS stream encoder lifecycle management (kernel coalescing). PyTorch's `MPSStream` maintains an internal `_commandEncoder` that it reuses across operations. Creating an encoder directly can conflict with this active encoder, causing:

```
[AGXG16XFamilyCommandBuffer tryCoalescingPreviousComputeCommandEncoderWithConfig:nextEncoderClass:]:
  failed assertion 'A command encoder is already encoding to this command buffer'
```

This crashes the process (not a catchable exception — it's a Metal framework assertion).

## When it triggers

Any time a Metal kernel is called twice in sequence without an intervening operation that flushes PyTorch's encoder state (e.g., `.cpu()`, `torch.mps.synchronize()`). This is common in real workloads.

## Fix

Use `stream->commandEncoder()` from PyTorch's `MPSStream` API instead:

```objc
at::mps::MPSStream *stream = at::mps::getCurrentMPSStream();
dispatch_sync(stream->queue(), ^{
    id<MTLComputeCommandEncoder> enc = stream->commandEncoder();
    // ... encode work ...
    // Do NOT call [enc endEncoding] — stream manages encoder lifecycle
});
stream->synchronize(at::mps::SyncType::COMMIT_AND_CONTINUE);
```

This properly integrates with PyTorch's encoder coalescing: it reuses the stream's active encoder (or creates one if needed), and `synchronize()` handles ending it via `endKernelCoalescing()`.

## Impact

Since the ReLU example serves as the reference implementation for all Metal kernels built with kernel-builder, any downstream kernel that follows this pattern will have the same bug. We discovered this while developing Metal kernels for `rotary-embedding` and `fused-rms-norm` — the crash is 100% reproducible when calling a kernel function twice in sequence.

robtaylor:
## Summary

- **Fix MPS encoder lifecycle**: Template and examples create compute encoders directly via `[commandBuffer computeCommandEncoder]`, bypassing PyTorch's MPS stream encoder management (kernel coalescing). This causes a fatal crash when any kernel is called twice in sequence: `A command encoder is already encoding to this command buffer`. Fixed by using `stream->commandEncoder()` from PyTorch's `MPSStream` API.
- **Lower Metal standard to metal3.1** (macOS 14+): All current kernel features (`bfloat16_t`, `simd_sum`, `simd_shuffle`, `threadgroup_barrier`) are available in Metal 3.1. Previous `metal4.0` required macOS 26.
- **Multi-strategy Metal toolchain detection**: Support macOS 14/15 (metal bundled in Xcode) and macOS 26 (separate Metal toolchain cryptex). Falls back to scanning `/Applications/Xcode*.app` when `xcrun`/`xcode-select` are unavailable in the Nix sandbox.
- **macOS CI matrix**: Build-test on macos-14, macos-15, and macos-26-xlarge (arm64 with GPU). macOS 14 is marked best-effort (`continue-on-error`) due to MPS OOM on 16GB runners.
- **Update Metal docs**: Document macOS 15+ as supported baseline, macOS 14 as best-effort.

## Files changed

- `build2cmake/src/templates/metal/compile-metal.cmake` — metal3.1 default, multi-strategy toolchain detection
- `builder/lib/torch-extension/arch.nix` — toolchain fallback for macOS 14/15, clear SDKROOT in xcrunHost
- `template/__KERNEL_NAME_NORMALIZED___metal/__KERNEL_NAME_NORMALIZED__.mm` — use MPSStream encoder API
- `builder/examples/relu/relu_metal/relu.mm` — same fix
- `builder/examples/extra-data/relu_metal/relu.mm` — same fix
- `.github/workflows/build_kernel_macos.yaml` — macOS version matrix, sandbox=relaxed, macos-14 best-effort
- `docs/source/builder/metal.md` — macOS version support table, updated requirements

## Test plan

- [x] Build-tested on macos-14 (Xcode 15), macos-15 (Xcode 16), macos-26 (Xcode 26) via CI matrix
- [x] GPU tests pass on macos-15 and macos-26 (relu + relu-metal-cpp)
- [x] macos-14 builds succeed, test_relu_layer OOM (known, MPS memory on 16GB runner)
- [x] Verified `fused-rms-norm` kernel: 74/74 tests pass (was crashing on sequential calls)
- [x] Verified `rotary-embedding` kernel: 217/217 tests pass
- [x] Crash is 100% reproducible before fix, 0% after fix

Fixes #307

### DIFF
--- a/.github/workflows/build_kernel_macos.yaml
+++ b/.github/workflows/build_kernel_macos.yaml
@@ -17,10 +17,14 @@ jobs:
         include:
           - os: macos-14-xlarge
             xcode: "/Applications/Xcode_15.4.app"
+            # macOS 14 is best-effort: builds work but MPS tests may OOM
+            # on runners with limited unified memory.
+            allow-failure: true
           - os: macos-15-xlarge
             xcode: "/Applications/Xcode_16.2.app"
           - os: macos-26-xlarge
             xcode: "/Applications/Xcode_26.0.app"
+    continue-on-error: ${{ matrix.allow-failure || false }}
     steps:
       - name: "Select Xcode"
         run: sudo xcrun xcode-select -s ${{ matrix.xcode }}
--- a/docs/source/builder/metal.md
+++ b/docs/source/builder/metal.md
@@ -5,24 +5,29 @@ Instructions on this page assume that you installed Nix with the
 
 ## Targeted macOS versions
 
-Since new macOS versions get [adopted quickly](https://telemetrydeck.com/survey/apple/macOS/versions/),
-we only support the latest major macOS version except for the first weeks
-after a release, when we also support the previous major version.
+Metal kernels are compiled with `-std=metal3.1` (AIR v26), which requires
+macOS 15 or later on ARM64 (Apple Silicon).
 
-We currently support macOS 26.0 and later on ARM64 (Apple silicon).
+| macOS version | Support level |
+|---------------|---------------|
+| macOS 26+     | Fully supported and tested in CI |
+| macOS 15      | Fully supported and tested in CI |
+| macOS 14      | Best-effort (builds work, some tests may fail due to MPS memory limits) |
 
 ## Requirements
 
 To build a Metal kernel, the following requirements must be met:
 
-- Xcode 26.x must be available on the build machine.
-- `xcode-select -p` must point to the Xcode 26 installation, typically
+- An Xcode installation with the Metal compiler must be available. The build
+  system automatically detects the Metal toolchain from available Xcode
+  installations.
+- On macOS 26+, the Metal Toolchain is a separate download from Xcode:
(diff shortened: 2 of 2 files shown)


danieldk:
Nice work! One question, isn't Metal 4 required for the new Neural Accelerators of M5 GPUs?

PR #320, opened 2026-03-09
kernels: add support from tvm-ffi build variants

danieldk:
This change adds support for tvm-ffi kernels to the `kernels` Python module. It combines the following changes:

- A new submodule `backends` is added. The backend detection is moved from `utils` to this module. The backend detection first tries to use Torch to detect the framework and framework version and falls back to probing the framework itself if Torch is not installed (only CUDA is supported for now).
- The backends are now represented as data classes that also carry the version when applicable.
- The build variant name construction was moved from the `utils` submodule to the `variants` submodule. The `tvm-ffi` variants are added as an additional option with a lower priority than Torch variants.
- Added tests and run them in CI without Torch as well.

### DIFF
--- a/kernels/src/kernels/backends.py
+++ b/kernels/src/kernels/backends.py
@@ -2,23 +2,41 @@ import ctypes
 import ctypes.util
 import warnings
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, Protocol
 
 from packaging.version import Version
 
 from kernels.compat import has_torch
 
 
+class Backend(Protocol):
+    @property
+    def name(self) -> str:
+        """
+        Short name of the backend, e.g. "cuda", "rocm", "cpu", etc.
+        """
+        ...
+
+    @property
+    def variant(self) -> str:
+        """
+        The name of the backend as used in a build variant, e.g. `cu128`
+        for CUDA 12.8.
+        """
+        ...
+
+
 @dataclass
 class CANN:
     version: Version
 
-    def __str__(self) -> str:
-        return f"cann{self.version.major}{self.version.minor}"
-
     @property
     def name(self) -> str:
         return "cann"
+
+    @property
+    def variant(self) -> str:
+        return f"cann{self.version.major}{self.version.minor}"
 
 
 @dataclass
@@ -27,7 +45,8 @@ class CPU:
     def name(self) -> str:
         return "cpu"
 
-    def __str__(self) -> str:
+    @property
+    def variant(self) -> str:
         return "cpu"
 
 
@@ -39,7 +58,8 @@ class CUDA:
     def name(self) -> str:
         return "cuda"
 
-    def __str__(self) -> str:
+    @property
+    def variant(self) -> str:
         return f"cu{self.version.major}{self.version.minor}"
 
 
@@ -49,7 +69,8 @@ class Metal:
     def name(self) -> str:
         return "metal"
 
-    def __str__(self) -> str:
+    @property
+    def variant(self) -> str:
         return "metal"
 
 
@@ -59,7 +80,8 @@ class Neuron:
     def name(self) -> str:
         return "neuron"
 
-    def __str__(self) -> str:
+    @property
+    def variant(self) -> str:
         return "neuron"
 
 
@@ -71,7 +93,8 @@ class ROCm:
     def name(self) -> str:
         return "rocm"
 
-    def __str__(self) -> str:
+    @property
+    def variant(self) -> str:
         return f"rocm{self.version.major}{self.version.minor}"
 
 
@@ -83,11 +106,9 @@ class XPU:
     def name(self) -> str:
         return "xpu"
 
-    def __str__(self) -> str:
+    @property
+    def variant(self) -> str:
         return f"xpu{self.version.major}{self.version.minor}"
-
-
-Backend = CANN | CPU | CUDA | Metal | Neuron | ROCm | XPU
 
 
 def _backend() -> Backend:
--- a/kernels/src/kernels/variants.py
+++ b/kernels/src/kernels/variants.py
@@ -17,7 +17,7 @@ def _torch_build_variant(backend: str | None) -> list[str]:
 
     selected_backend = _select_backend(backend)
 
-    compute_framework = str(selected_backend)
+    backend_variant = selected_backend.variant
 
     import torch
 
@@ -28,17 +28,17 @@ def _torch_build_variant(backend: str | None) -> list[str]:
     if os == "darwin":
         cpu = "aarch64" if cpu == "arm64" else cpu
         return [
-            f"torch{torch_version.major}{torch_version.minor}-{compute_framework}-{cpu}-{os}"
+            f"torch{torch_version.major}{torch_version.minor}-{backend_variant}-{cpu}-{os}"
         ]
     elif os == "windows":
         cpu = "x86_64" if cpu == "AMD64" else cpu
         return [
-            f"torch{torch_version.major}{torch_version.minor}-{compute_framework}-{cpu}-{os}"
+            f"torch{torch_version.major}{torch_version.minor}-{backend_variant}-{cpu}-{os}"
         ]
 
     cxxabi = "cxx11" if torch.compiled_with_cxx11_abi() else "cxx98"
     return [
-        f"torch{torch_version.major}{torch_version.minor}-{cxxabi}-{compute_framework}-{cpu}-{os}"
+        f"torch{torch_version.major}{torch_version.minor}-{cxxabi}-{backend_variant}-{cpu}-{os}"
     ]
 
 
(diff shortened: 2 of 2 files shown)


kernels/src/kernels/backends.py
```diff
@@ -0,0 +1,179 @@
...
+        return "xpu"
+
+    def __str__(self) -> str:
+        return f"xpu{self.version.major}{self.version.minor}"
+
+
+Backend = CANN | CPU | CUDA | Metal | Neuron | ROCm | XPU
+
+
+def _backend() -> Backend:
```
danieldk, line 114: We should probably rename this to `backend` or `find_backend` in the future, but I wanted to keep changes minimal and there may be variables with the same name as well.

kernels/src/kernels/variants.py
```diff
@@ -1,3 +1,85 @@
...
+    else:
+        return [f"torch-{selected_backend.name}"]
+
+
+def _build_variant_universal() -> list[str]:
+    # Once we support other frameworks, detection goes here.
+    return ["torch-universal"] if has_torch else []
+
+
+def _build_variants(backend: str | None) -> list[str]:
```
danieldk, line 78: We should probably rename this to `build_variants` in the future, but I wanted to keep changes minimal and there may be variables with the same name as well.

kernels/tests/conftest.py
```diff
@@ -36,17 +46,19 @@ def pytest_addoption(parser):
...
         return "xpu"
-    elif _get_privateuse_backend_name() == "npu":
+    elif has_npu:
         return "npu"
 
     return "cpu"
 
 
 def pytest_runtest_setup(item):
+    if "torch_only" in item.keywords and torch is not None:
```
danieldk, line 60: There are many tests that are torch-only, but we need a way to run without torch in `test_tvm_ffi.py`.

kernels/tests/conftest.py
```diff
@@ -1,29 +1,39 @@
...
+from kernels.backends import _get_torch_privateuse_backend_name
+
+try:
+    import torch
+except ImportError:
+    torch = None
 
 has_cuda = (
-    hasattr(torch.version, "cuda")
+    torch is not None
```
danieldk, line 13: We might want to uncouple these device detections from Torch in the future, but for now all the tests also assume Torch. So we need to make this generic + add `has_torch` on most of the existing tests. But it's a bit out-of-scope for this PR.

kernels/src/kernels/cli/upload.py
```diff
@@ -20,7 +21,9 @@ def upload_kernels_dir(
     for candidate in [kernel_dir / "build", kernel_dir]:
         variants = [
             variant_path
-            for variant_path in candidate.glob("torch*")
+            for variant_path in chain(
+                candidate.glob("torch*"), candidate.glob("tvm-ffi*")
```
sayakpaul, line 25: Not a strong opinion but should `torch` and `tvm-ffi` come from a `set` constant?

Or even better could be to maintain a set of framework where each framework would be a dataclass having a field for its identifier name. But I understand this is probably out of scope for this PR and introduces complexity that I am not if we want.

kernels/src/kernels/cli/upload.py
```diff
@@ -77,6 +80,6 @@ def upload_kernels_dir(
         path_in_repo="build",
         delete_patterns=list(delete_patterns),
         commit_message="Build uploaded using `kernels`.",
-        allow_patterns=["torch*"],
+        allow_patterns=["torch*", "tvm-ffi*"],
```
sayakpaul, line 83: Same as above.

kernels/src/kernels/backends.py
```diff
@@ -0,0 +1,179 @@
...
+import warnings
+from dataclasses import dataclass
+from typing import Optional
+
+from packaging.version import Version
+
+from kernels.compat import has_torch
+
+
+@dataclass
```
sayakpaul, line 29: Should we have a base `Backend` class and have the actual backends subclass it?

kernels/src/kernels/compat.py
```diff
@@ -1,8 +1,14 @@
+import importlib.util
```
sayakpaul, line 1: Should these changes to `import_utils.py`?

kernels/src/kernels/variants.py
```diff
@@ -1,3 +1,85 @@
...
 import re
 
-BUILD_VARIANT_REGEX = re.compile(r"^(torch\d+\d+|torch-(cpu|cuda|metal|rocm|xpu))")
+from packaging.version import parse
+
+from kernels.backends import _select_backend
+from kernels.compat import has_torch, has_tvm_ffi
+
+BUILD_VARIANT_REGEX = re.compile(
+    r"^(torch\d+\d+|torch-(cpu|cuda|metal|neuron|rocm|xpu)|tvm-ffi\d+\d+)"
```
sayakpaul, line 10: Same here. The backends supported for torch could come from a constant or a function like `get_supported_backends()`. Ideally, this should also be done for the frameworks? 

This way, we won't have to remember that this regex should be changed when supporting a new framework or a backend.

sayakpaul: Does not have to be this PR but we should probably also have a function that gives us all the backends we support per framework. That would be useful for many things I reckon.

kernels/tests/conftest.py
```diff
@@ -36,17 +46,19 @@ def pytest_addoption(parser):
...
+    if has_cuda:
         return "cuda"
-    elif hasattr(torch, "xpu") and torch.xpu.is_available():
+    elif has_xpu:
         return "xpu"
-    elif _get_privateuse_backend_name() == "npu":
+    elif has_npu:
         return "npu"
 
     return "cpu"
```
sayakpaul, line 56: Doesn't have to be this PR but this kinda utility is generally also useful out of tests.

kernels/tests/conftest.py
sayakpaul: How do we feel about decorators / markers like `require_torch`, `require_tvm_ffi`, etc.? This way, we know upfront what is the primary dependency for a test suite.

kernels/pyproject.toml
```diff
@@ -29,6 +29,7 @@ dev = [
   # Whatever version is compatible with pytest.
   "pytest-benchmark",
   "torch>=2.5",
+  "apache-tvm-ffi>=0.1.9,<0.2.0",
```
sayakpaul, line 32: Is there a strong reason for fixing when there's no upper limit to `torch`?

flake.nix
```diff
@@ -161,6 +164,7 @@
               docutils
               huggingface-hub
               kernel-abi-check
+              matplotlib
```
sayakpaul, line 167: Why is this needed? 🤯

kernels/tests/conftest.py
danieldk: I am not sure I understand, these are used as decorators. We should probably rename this one to `has_torch` like the others.

kernels/pyproject.toml
danieldk: Yes, tvm-ffi applies semantic versioning, so there may be incompatible changes in 0.2.

flake.nix
danieldk: The benchmarks use matplotlib, so we want it in the standard Nix development environment so that mypy can do typechecks, we can run benchmarks, etc. We want the Nix devshell to provide the full development environment without having to `pip install` things (which may not work due to non-FHS on Nix/NixOS).

kernels/src/kernels/cli/upload.py
danieldk: Yeah, I agree, we may want to do something like

```
class Torch:
  GLOB = "torch*"
  REGEXP = ...
```

kernels/src/kernels/variants.py
danieldk: Yeah, I agree, this should probably be a function like

```
class Torch:
  version: Version
  backend: Backend

  @staticmethod
  def parse_variant(self, variant: str) -> Self:
      ...

class TorchNoArch:
   ...

  @staticmethod
  def parse_variant(self, variant: str) -> Self:
      ...
```

We should also move away from regexes for this, they are really fragile. We could let the parse function split parts, parse/check parts. But sounds like something nice to have in a future PR and very much in line with making backends more formalized in this PR.

kernels/src/kernels/backends.py
danieldk: Hmmm, subclassing sounds kind of heavyweight for what is essentially a small interface. Let me add a protocol and see what you think.

danieldk: It's a protocol.

sayakpaul: Okay so, it's currently used for duck typing then. Fine by me.

kernels/src/kernels/cli/upload.py
sayakpaul: Opened https://github.com/huggingface/kernels/issues/323 to track.

PR #321, opened 2026-03-10
[chore] catch styling bugs in the prs better.

sayakpaul:
Currently, our CI doesn't properly catch the styling issues despite having the `lint.yml` workflow. This PR fixes that.

Additionally, this PR:

* Introduces a `quality` command in the `Makefile`. So, when developers run `make style && make quality`, they know that the CI will be through with this.
* Enforces styling checks to pass first before running the `test_kernels.yml` workflow.


kernels/tests/test_kernel_card.py
```diff
@@ -23,8 +23,7 @@ def mock_kernel_dir():
         kernel_dir = Path(tmpdir)
 
         build_toml = kernel_dir / "build.toml"
-        build_toml.write_text(
-            """[general]
+        build_toml.write_text("""[general]
```
sayakpaul: Result of `make style`.

We could consider fixing the `black` version (same for isort and ruff) but I don't think it's a big deal.

sayakpaul:
Doesn't look like we have `update_python_depends.py`?

.github/workflows/test_kernels.yaml
```diff
@@ -13,7 +13,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  styling-and-lints:
+    uses: ./.github/workflows/lint.yml
```
danieldk: Not sure I fully understand this what's the benefit of running it as a job here, since it's already run as a separate lint job?

Also, this workflow is really for running tests, not for styling. Even if the style workflow fails, it's nice to still have the outcome of testing.

kernels/tests/test_layer.py
```diff
@@ -102,7 +102,7 @@ def __init__(self):
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         self.n_calls += 1
-        d = input.shape[-1] // 2
+        _ = input.shape[-1] // 2
```
danieldk: I think we can just remove this line. This doesn't do anything.

.github/workflows/test_kernels.yaml
sayakpaul: Oh we usually make the actual testing conditioned on styling passes in other repos like `transformers`. I can remove it.

.github/workflows/lint.yml
```diff
@@ -24,28 +25,22 @@ jobs:
...
 
-      - name: Install black
-        run: uv pip install black
-
       - name: Check formatting
-        run: |
-          uv run black --check kernels
+        run: uv run --with black black --check kernels/src kernels/tests
 
-  validate-dependencies:
```
danieldk, line 34: The dependency validation gets removed?

sayakpaul: I removed it because `update_python_depends.py` (the script this step uses) is missing.

danieldk: Ah, right. But we should still check that the files are the same in the Rust crate and the Python packages (ideally we'd only have one file, but both don't deal well with out-of-tree files).

danieldk: I think something like `cmp -s file1 file2` should do the trick, since it returns exit code != 0 when the files differ.

sayakpaul: Do the latest changes work for you? 👀

sayakpaul:
Closing in favor of https://github.com/huggingface/kernels/pull/462. Thanks @drbh

drbh:
closing as mentioned above. please feel free to reopen if needed @sayakpaul. just trying to clean up some of the stale PRs

PR #325, opened 2026-03-10
[docs] start a doc on why kernels.

sayakpaul:


docs/source/why_kernels.md
```diff
@@ -0,0 +1,28 @@
+## Why kernels?
+
+Why do we want yet-another standard? Our goal with the `kernels`
+package is two-fold:
```
danieldk: I'd remove _Why do we want yet-another standard?_, it immediately invokes https://xkcd.com/927/ for me, which is not what we want.

docs/source/why_kernels.md
```diff
@@ -0,0 +1,28 @@
...
+
+* Establish a much needed standardization in the ecosystem of
+custom kernels. This is enforced in the build structure, support
+matrices, reproducibility in the package. We are trying to do this
+through `builder` component.
```
danieldk:
```suggestion
* Establish a standard way of structuring and building kernels. The `builder` component takes kernel source in a pre-defined layout, with a declarative build configuration, and produces compiled kernels for a wide matrix of compute backends (e.g. CUDA, ROCM and XPU), operating systems, and architectures. The builder also enforces reproducibility of the build environment and build steps.
```

docs/source/why_kernels.md
```diff
@@ -0,0 +1,28 @@
...
+* Provide an easy way for users to load and use pre-built compute
+kernels, which, otherwise can take a very long time to build. For
+example, building something like Flash Attention 3 from source can
+take a couple of hours to build locally.
```
danieldk: * Provide a standard way of distributing and loading kernels. Kernels are distributed through the Hugging Face Hub and can be loaded through the `kernels` Python package. `kernels` fetches the right kernel build for the system that it runs, avoiding long local, sometimes hours-long, builds. It also supports loading of multiple versions of the same kernel, effectively eliminating 'dependency hell'.

docs/source/why_kernels.md
```diff
@@ -0,0 +1,28 @@
...
+* Provide kernel builds across all supported PyTorch versions,
+accelerators, and capabilities. This is particularly important
+because local kernel builds can become unusable once the base
+runtime (e.g., CUDA) is updated to a recent version.
```
danieldk:
```suggestion
because local kernel builds can become unusable once the base
machine learning framework (e.g. Torch) is updated to a recent version.
```

(CUDA is a bit tricky as an example, since e.g. CUDA 12.6 would still work after updating to 12.8.)

docs/source/why_kernels.md
```diff
@@ -0,0 +1,28 @@
...
+
+Additionally, there are several advantages to hosting the pre-built
+kernels on the Hugging Face Hub platform:
+
+* Trends -- kinds of models using a kernel more than others, for
+example.
```
danieldk: I think there is something missing here?

docs/source/why_kernels.md
sayakpaul: TIL.

docs/source/why_kernels.md
sayakpaul:
```suggestion
* Provide a standard way of distributing and loading kernels. Kernels are distributed through the Hugging Face Hub and can be loaded through the kernels Python package. kernels fetches the right kernel build for the system that it runs, avoiding long local, sometimes hours-long, builds. It also supports loading of multiple versions of the same kernel, effectively eliminating 'dependency hell'.
```

Issue #315, opened 2026-03-06
README utility for `main` branches on the build repos on Hub


sayakpaul:
After https://github.com/huggingface/kernels/pull/287 is done, add a utility to generate a README for the `main` branch that points to the revisions because we have started recommending `version`s when downloading a kernel via `get_kernel()`.

In doing so, it might so happen that the user gets treated with a blank page when opening a kernel repo (example https://huggingface.co/kernels-community/sgl-flash-attn3) even though it might have content in other `revision`s. This could be misleading.

PR #329, opened 2026-03-11
feat: update `main` revision README to the most updated version one

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/315

I am guessing we should incorporate this utility within the builder (perhaps the build and upload command) or something?


docs/source/cli-collate-readme.md
```diff
@@ -0,0 +1,17 @@
+### kernels collate-readme
+
+Use `kernels collate-readme <repo_id>` to generate a README for the `main`
```
danieldk: Repo READMEs are instantiated cards. Shouldn't the versions be enumerated by card filling?

sayakpaul: Sorry didn't get it. Did you mean https://github.com/huggingface/kernels/pull/287/ should have a section on "Versions" in the card template and populate it automatically?

Even if that's the case, I think the PR could still have its own scope because we're tackling the README that will be included in the `main` revision of the kernel repo. The builder will automatically generate the cards and push to particular revisions in which case the README will still likely stay blank. Or am I missing something?

danieldk: For now I think `main` will be the default branch that is shown? Then `main`'s README should be the instantiated card of the most recent version. I am not sure if all the version branches need individual cards, since we do not want people to actually use older versions. We leave them around to avoid breakage for people who haven't updated their code to use the latest kernel version yet, other than that, we should hide the old versions as much as possible.

julien-c: i think it makes sense ^

Issue #326, opened 2026-03-10
Support resolving the CUDA version to another compatible version


danieldk:
E.g. if a system uses Torch build with CUDA 12.9, but we only have 12.6, 12.8, and 13.0 builds, we should resolve to 12.8. That is if the CUDA version is x.y, we find the build with the largest version x.z, such that z <= y.

PR #330, opened 2026-03-11
[core] feat: improve variant resolution

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/326

Should we document it somewhere?

### DIFF
--- a/builder/build-variants.json
+++ b/builder/build-variants.json
@@ -20,6 +20,7 @@
       "torch210-cxx11-cu130-aarch64-linux",
       "torch29-cxx11-cu126-aarch64-linux",
       "torch29-cxx11-cu128-aarch64-linux",
+      "torch29-cxx11-cu129-aarch64-linux",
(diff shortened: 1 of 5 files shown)


kernels/tests/test_variants.py
```diff
@@ -0,0 +1,131 @@
...
+    def test_neuron_returns_single_variant(self):
+        assert _compatible_backend_variants(Neuron()) == ["neuron"]
+
+    def test_rocm_returns_single_variant(self):
+        backend = ROCm(version=Version("6.2"))
+        assert _compatible_backend_variants(backend) == ["rocm62"]
+
+    def test_xpu_returns_single_variant(self):
+        backend = XPU(version=Version("2024.2"))
+        assert _compatible_backend_variants(backend) == [backend.variant]
```
danieldk: Probably best to put a literal here as well on the RHS of the equality check?

PR #335, opened 2026-03-11
python3Packages.nvidia-cutlass-dsl: 4.3.0 -> 4.4.1

danieldk:


nix/pkgs/python-modules/nvidia-cutlass-dsl/default.nix
```diff
@@ -1,61 +1,49 @@
...
-    "libcuda.so.1"
+  pythonRemoveDeps = lib.optionals (cudaPackages.cudaAtLeast "13.0") [
+    # nvidia-cutlass-dsl-libs-cu13 has a dependency on the base package,
+    # but it has the same contents + CUDA 13 extensions.
+    "nvidia-cutlass-dsl-libs-base"
   ];
 
   meta = {
-    broken = !(cudaPackages.cudaAtLeast "12.8");
+    broken = nvidia-cutlass-dsl-libs.meta.broken;
```
sayakpaul, line 47: Wow, this is quite cool!

nix/pkgs/python-modules/nvidia-cutlass-dsl/default.nix
```diff
@@ -1,61 +1,49 @@
...
-  };
-  hash =
-    hashes."${pyShortVersion}-${stdenv.system}"
-      or (throw "Unsupported Python version: ${pyShortVersion}-${stdenv.system}");
-
 in
 buildPythonPackage rec {
   pname = "nvidia-cutlass-dsl";
-  version = "4.3.0";
+  version = "4.4.1";
```
sayakpaul, line 21: Pardon my unwisdom but is this sufficiently BW-compatible?

sayakpaul (approved):

nix/pkgs/python-modules/nvidia-cutlass-dsl/default.nix
danieldk: I think so, kernels using this are going to use this as a Python API, so it's kind of similar to how we handle Triton or Torch Python dependencies (i.e. not at all 😅, it's up to kernels to check the version and have specific code paths if they want to support old Triton/Torch versions). In this case, I think there isn't going to be any breakage at all, because most kernel couldn't use this yet, since JIT-ing was broken in cutlass-dsl in kernels due to not supporting relative imports (fixed in 4.4).

PR #339, opened 2026-03-12
Refactor variant handling and add CUDA fallback

danieldk:
Build variants were stringly-typed throughout kernels, with custom parsing and serialization sprinkled everywhere. This change adds proper/strong typing to variants adding a `Variant` class. This also centers parsing/serialization in one place and allows code to easily query various parts of of a variant.

This also fundamentally changes how we deal with getting variants from the Hub. Rather than casting a wide net with all possible variants and using allow patterns based on that, we query the hub for variants of a kernel, parse them and can decide if there is an applicable variant ahead of time. If there are multiple applicable variants, we can select the best one (e.g. arch before noarch or recent CUDA version before older versions).

The filtering and sorting logic is now completely centralized in two functions (`_filter_variants` and `_sort_variants`).

### DIFF
--- a/kernels/src/kernels/variants.py
+++ b/kernels/src/kernels/variants.py
@@ -166,8 +166,6 @@ def get_variants(api: HfApi, *, repo_id: str, revision: str) -> list[Variant]:
             logging.warning(
                 f"Repository {repo_id} (revision: {revision}) contains invalid build variant variant: {variant_str!r}"
             )
-            log
-            pass
     return variants
 
 


kernels/src/kernels/variants.py
```diff
@@ -1,85 +1,324 @@
...
+            return f"{self.backend.variant_str}-{self.platform}-{self.os}"
+        else:
+            return f"{'cxx11' if self.cxx11_abi else 'cxx98'}-{self.backend.variant_str}-{self.platform}-{self.os}"
+
+    @staticmethod
+    def parse(parts: list[str]) -> "Arch":
+        # Handle Linux with cxx11 marker.
+        if len(parts) == 4:
+            cxx11_abi = parts[0] == "cxx11"
+            parts = parts[1:]
```
drbh, line 91: not sure if we want to enforce that the abi is one of the two we support


```suggestion
        if len(parts) == 4:
            if parts[0] not in {"cxx11", "cxx98"}:
                raise ValueError(f"Invalid Linux ABI marker: {parts[0]!r}")        
            cxx11_abi = parts[0] == "cxx11"
            parts = parts[1:]
```

danieldk: Good catch. I think we should enforce a default, so that we can drop the `-cxx11` tag in the future, but we should flip it around so that cxx11 is the default.

kernels/src/kernels/cli/versions.py
```diff
@@ -1,44 +1,22 @@
-from importlib.util import find_spec
-from pathlib import Path
-
-from huggingface_hub import HfApi
-
 from kernels._versions import _get_available_versions
-from kernels.utils import _build_variants, _get_hf_api
-from kernels.variants import BUILD_VARIANT_REGEX
+from kernels.utils import _get_hf_api
+from kernels.variants import get_variants, resolve_variant
```
sayakpaul: Should we consider making them private?

kernels/src/kernels/cli/versions.py
```diff
@@ -1,44 +1,22 @@
...
-
     versions = _get_available_versions(repo_id).items()
     if not versions:
         print(f"Repository does not support kernel versions: {repo_id}")
         return
 
     for version, ref in sorted(versions, key=lambda x: x[0]):
+        variants = get_variants(api, repo_id=repo_id, revision=ref.ref)
+        best = resolve_variant(variants)
         print(f"Version {version}: ", end="")
```
sayakpaul, line 22: I think we can remove this?

kernels/src/kernels/cli/versions.py
```diff
@@ -1,44 +1,22 @@
...
-        variants = [
-            f"{variant} ✅" if variant in compatible_variants else f"{variant}"
-            for variant in _get_build_variants(api, repo_id, ref.ref)
+        variant_strs = [
+            f"{variant.variant_str} ✅" if variant == best else f"{variant.variant_str}"
```
sayakpaul: (nit): should this a bit more descriptive rather than just the variant name? Something like `"✅ {variant.variant_str} available" if variant == best else f"{variant.variant_str}"`?

kernels/src/kernels/cli/versions.py
```diff
@@ -1,44 +1,22 @@
...
-    for filename in api.list_repo_files(repo_id, revision=revision):
-        path = Path(filename)
-        if len(path.parts) < 2 or path.parts[0] != "build":
-            continue
-
-        match = BUILD_VARIANT_REGEX.match(path.parts[1])
-        if match:
-            variants.add(path.parts[1])
-    return sorted(variants)
+        print(", ".join(variant_strs))
```
sayakpaul, line 31: (nit): `logging`?

kernels/src/kernels/backends.py
```diff
@@ -46,9 +59,15 @@ def name(self) -> str:
         return "cpu"
 
     @property
-    def variant(self) -> str:
+    def variant_str(self) -> str:
         return "cpu"
 
+    @staticmethod
+    def parse(s: str) -> "CPU":
+        if s != "cpu":
```
sayakpaul, line 64:
```suggestion
        if s != self.name:
```

kernels/src/kernels/backends.py
```diff
@@ -70,9 +96,15 @@ def name(self) -> str:
         return "metal"
 
     @property
-    def variant(self) -> str:
+    def variant_str(self) -> str:
         return "metal"
 
+    @staticmethod
+    def parse(s: str) -> "Metal":
+        if s != "metal":
```
sayakpaul, line 103:
```suggestion
        if s != self.name:
```

kernels/src/kernels/backends.py
```diff
@@ -81,9 +113,15 @@ def name(self) -> str:
         return "neuron"
 
     @property
-    def variant(self) -> str:
+    def variant_str(self) -> str:
         return "neuron"
 
+    @staticmethod
+    def parse(s: str) -> "Neuron":
+        if s != "neuron":
```
sayakpaul, line 120:
```suggestion
        if s != self.name:
```

kernels/src/kernels/backends.py
```diff
@@ -8,6 +9,11 @@
 
 from kernels.compat import has_torch
 
+_CUDA_VARIANT_REGEX = re.compile(r"cu(\d+)(\d+)")
+_ROCM_VARIANT_REGEX = re.compile(r"rocm(\d+)(\d+)")
+_XPU_VARIANT_REGEX = re.compile(r"xpu(\d+)(\d+)")
+_CANN_VARIANT_REGEX = re.compile(r"cann(\d+)(\d+)")
```
sayakpaul: Okay to skip "npu" for now?

kernels/src/kernels/utils.py
```diff
@@ -155,30 +172,21 @@ def _find_kernel_in_repo_path(
...
-            break
-
-    if variant is None:
-        raise FileNotFoundError(
-            f"Kernel at path `{repo_path}` does not have one of build variants: {', '.join(variants)}"
-        )
-
-    assert variant_path is not None
+    variant_str = variant.variant_str
+    variant_path = repo_path / "build" / variant_str
```
sayakpaul, line 179: So, by this point, we want the `variant` to always be resolved unlike previously?

sayakpaul: If so, could I get a brief rundown of what got changed here?

kernels/src/kernels/variants.py
```diff
@@ -1,85 +1,324 @@
...
 BUILD_VARIANT_REGEX = re.compile(
     r"^(torch\d+\d+|torch-(cpu|cuda|metal|neuron|rocm|xpu)|tvm-ffi\d+\d+)"
 )
 
+_TORCH_VARIANT_REGEX = re.compile(r"torch(\d+?)(\d+)")
+_TVM_FFI_VARIANT_REGEX = re.compile(r"tvm-ffi(\d+?)(\d+)")
+
+
+@dataclass(unsafe_hash=True)
+class Torch:
```
sayakpaul, line 29: Would it make sense to have a `Framework` baseclass and have these (`Torch` and `TvmFfi`) implement it? 

Skeleton of `Framework` could be:

```py
@dataclass(unsafe_hash=True)
class Framework:
    version: Version
    
     @property
     def variant_str(): ...
     
     @staticmethod
     def parse(): ...
   
     ... 
```

sayakpaul: I think the framework-specific classes could also hold the regex pattern as a class attribute.

kernels/src/kernels/variants.py
```diff
@@ -1,85 +1,324 @@
...
+        elif parts[0] == "tvm" and parts[1].startswith("ffi"):
+            framework = TvmFfi.parse(f"tvm-{parts[1]}")
+            arch = Arch.parse(parts[2:])
+        else:
+            raise ValueError(f"Unknown framework in variant string: {variant_str!r}")
+
+        return Variant(framework=framework, arch=arch)
+
+
+def get_variants(api: HfApi, *, repo_id: str, revision: str) -> list[Variant]:
```
sayakpaul, line 153: Beautiful!

kernels/src/kernels/variants.py
```diff
@@ -1,85 +1,324 @@
...
+    """Get all the build variants available from a kernel repository."""
+
+    try:
+        tree = api.list_repo_tree(repo_id, path_in_repo="build", revision=revision)
+        variant_strs = {
+            item.path.split("/")[-1] for item in tree if isinstance(item, RepoFolder)
+        }
+    except Exception:
```
sayakpaul: Perhaps a `logger.warning()` with a descriptive message here?

kernels/src/kernels/variants.py
```diff
@@ -1,85 +1,324 @@
...
         return []
 
-    selected_backend = _select_backend(backend)
+    variants = []
+    for variant_str in variant_strs:
+        try:
+            variants.append(Variant.parse(variant_str))
+        except ValueError:
+            pass
```
sayakpaul: Same as above.

kernels/src/kernels/variants.py
```diff
@@ -1,85 +1,324 @@
...
+    )
 
-    tvm_ffi_version = parse(tvm_ffi.__version__)
-    cpu = platform.machine()
-    os = platform.system().lower()
 
-    return [
-        f"tvm-ffi{tvm_ffi_version.major}{tvm_ffi_version.minor}-{backend_variant}-{cpu}-{os}"
-    ]
+def _resolve_variant_for_system(
```
sayakpaul, line 242: This is very idiomatic!

kernels/tests/test_variants.py
```diff
@@ -0,0 +1,322 @@
+import pytest
+from huggingface_hub import HfApi
+from packaging.version import Version
+
+from kernels.backends import CPU, CUDA, ROCm
+from kernels.variants import Variant, _resolve_variant_for_system, get_variants
+
+VARIANT_STRINGS = [
```
sayakpaul, line 8: tvm variants as well?

kernels/tests/test_variants.py
```diff
@@ -0,0 +1,322 @@
...
+        variants=RESOLVE_VARIANTS,
+        selected_backend=CUDA(Version("11.8")),
+        cpu="x86_64",
+        os="linux",
+        torch_version=Version("2.10"),
+        torch_cxx11_abi=True,
+        tvm_ffi_version=None,
+    )
+    assert result is not None
+    assert result.variant_str == "torch-cuda"
```
sayakpaul: Just for curiosity. Is this a reasonable resolution or should it have errored out saying no matching variants found?

sayakpaul (commented):
Great refactor! I think it will be much more sustainable long-term. 

I think we can close https://github.com/huggingface/kernels/pull/336 as well in favor of this PR?

kernels/src/kernels/cli/versions.py
danieldk: I think we should add a note to the documentation that only the symbols exported at the top level are considered stable.

kernels/src/kernels/cli/versions.py
danieldk: Which part? We are printing the version before printing all variants.

kernels/src/kernels/cli/versions.py
danieldk: I think logging shouldn't be used for typical Unix command-line utilities where output goes to stdout and you don't want to filter on level. Logging adds a lot of noise.

kernels/src/kernels/backends.py
danieldk: CANN == NPU

Neuron is not needed, since we don't have version specifiers.

kernels/src/kernels/cli/versions.py
sayakpaul: > Logging adds a lot of noise.

Out of curiosity: how so?

kernels/src/kernels/backends.py
danieldk: We can't 😢, `parse` is a static method, so there is no self. We could move it to the class, but then we need a class and an instance version of the name variable.

sayakpaul: Oh very sorry

kernels/src/kernels/utils.py
danieldk: Yeah, the responsibility of finding the variant to-be-used has moved to the caller. The reasoning is two-fold:

* We have to resolve the variant in the caller anyway to narrow down the allow pattern to just the relevant variant.
* It makes this function more orthogonal, if fulfills just what it says on the tin: finding the given variant in the repo path.

kernels/src/kernels/variants.py
danieldk: I would like to do this, but then `Framework` needs to be an abstract base class (since otherwise you could instantiate `Framework`). But ABCs don't work well with dataclasses and I think you need something like https://github.com/MichaelSchneeberger/dataclass-abc

Since it's only two classes now, a union should be ok I think? We could reconsider when we add more frameworks.

What do you think?

danieldk: > I think the framework-specific classes could also hold the regex pattern as a class attribute.

Great idea!

sayakpaul: > Since it's only two classes now, a union should be ok I think? We could reconsider when we add more frameworks.

Fine by me.

kernels/tests/test_variants.py
danieldk: Yeah, noarch are always an option. If we didn't resolve to that, we could never use a noarch kernel, even in a repo that only has noarch.

kernels/src/kernels/variants.py
danieldk: Fixed

kernels/src/kernels/cli/versions.py
danieldk: Prefixes to the actual output

```
>>> import logging                                                                                                                                                                                                                                                                                                                                                                                                                                     
>>> logging.warning("Yo!")                                                                                                                                                                                                           
WARNING:root:Yo!
```

kernels/src/kernels/variants.py
danieldk: I think the error here was to catch the exception at all, it should bubble up in this case, because the user may have used an invalid repo name, revision, etc. Changed it to bubble up.

kernels/src/kernels/variants.py
danieldk: Logging a warning now.

kernels/tests/test_variants.py
danieldk: Added 👍 .

kernels/src/kernels/variants.py
danieldk: Moved the regexes to the classes, also did the same for the backends in `backends.py`.

kernels/src/kernels/backends.py
```diff
@@ -18,98 +19,172 @@ def name(self) -> str:
...
         The name of the backend as used in a build variant, e.g. `cu128`
         for CUDA 12.8.
         """
         ...
 
 
-@dataclass
+@dataclass(unsafe_hash=True)
 class CANN:
+    _VARIANT_REGEX: ClassVar[re.Pattern] = re.compile(r"cann(\d+)(\d+)")
```
sayakpaul, line 32: Ah lovely!

PR #357, opened 2026-03-14
kernel-builder: add CUDA capability detection for tvm-ffi builds

### LINKED ISSUE #354: Add tvm-ffi CUDA capability detection (for devshells)

danieldk:
Make the CMake build to detect the CUDA capability of the GPU by default. This speeds up builds when developing, since we only need to compile for a single capability. Kernels still get built for the full of capabilities when doing a regular kernel build, since it passes `BUILD_ALL_SUPPORTED_ARCHS=ON`.

Fixes #354.


kernel-builder/src/pyproject/templates/tvm_ffi/preamble.cmake
```diff
@@ -36,7 +36,7 @@ endif()
 set(KERNEL_REVISION "{{ revision }}" CACHE STRING "Kernel revision, must be unique")
 set(OPS_NAME "_{{python_name}}_${BACKEND}_{{ revision }}")
 
-option(BUILD_ALL_SUPPORTED_ARCHS "Build all supported architectures" on)
+option(BUILD_ALL_SUPPORTED_ARCHS "Build all supported architectures" off)
```
sayakpaul, line 39: When is this turned back on?

danieldk: This mirrors the Torch version. It was set to `on` for tvm-ffi because the detection is not implemented yet. We set it to `on` in the Nix builder, so that official builds get the full range of capabilities: 

https://github.com/huggingface/kernels/blob/289fc4be66d6151b1e8277519cbbc40bb952603e/nix-builder/lib/extension/tvm-ffi/arch.nix#L219

PR #358, opened 2026-03-15
kernel-builder: restructure Torch templates

danieldk:
This change does not add any new code, but restructures the project a bit. A lot of templates were in the top-level directory from before we supported tvm-ffi, but they are Torch-specific, so this change:

* Moves Torch-specific templates to `src/templates/torch`.
* Splits up the `torch/common.rs` module. Functions that are used for Torch and tvm-ffi are moved to `common.rs`, functions that are only used by Torch are moved to `torch/mod.rs`.
* Move all code related to pyproject/CMake generation to `pyproject` module.
* Rename the `clean` subcommand to `clean-pyproject` to mirror `create-pyproject`.

### DIFF
--- a/kernel-builder/src/main.rs
+++ b/kernel-builder/src/main.rs
@@ -57,7 +57,7 @@ enum Commands {
     },
 
     /// Clean generated artifacts.
-    Clean {
+    CleanPyproject {
         #[arg(name = "BUILD_TOML")]
         build_toml: PathBuf,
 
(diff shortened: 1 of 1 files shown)


kernel-builder/src/pyproject/torch/deps.rs
```diff
@@ -16,7 +16,7 @@ pub fn render_deps(env: &Environment, build: &Build, write: &mut impl Write) ->
     for dep in deps {
         match dep {
             Dependency::Cutlass2_10 => {
-                env.get_template("cuda/dep-cutlass.cmake")
+                env.get_template("torch/cuda/dep-cutlass.cmake")
```
sayakpaul, line 19: Pardon my unwisdom in Rust but can't we define it as a global constant in the script and use it from there? Or maybe even as a dictionary like:

{"cuda": "torch/cuda/dep-cutlass.cmake", "xpu": "torch/xpu/dep-sycl-tla.cmake"}

danieldk: Yeah, we could do that. We have to see a bit how this particular module will pan out, since it'll probably move up to be used by tvm-ffi as well.

PR #362, opened 2026-03-16
[Terraform] Prevent user processes from killing on SSH disconnect

adarshxs:
tmux got killed, after SSH disconnection. This should fix it for instances created via terraform


terraform/nixos-configuration.nix
```diff
@@ -159,6 +159,11 @@
     dinit = "echo 'use nix' > .envrc && direnv allow"; # init direnv for a flake dir
   };
 
+  # -------------------------------------------------------------------------
+  # Keep user processes (e.g. tmux) alive after SSH disconnect
+  # -------------------------------------------------------------------------
+  services.logind.killUserProcesses = false;
```
danieldk: According to the docs this is the default?

https://nixos.org/manual/nixos/stable/options#opt-services.logind.settings.Login.KillUserProcesses

> Defaulted to false in nixpkgs because many tools that rely on persistent user processes—like tmux, screen, mosh, VNC, nohup, and more — would break by the systemd-default behavior.

adarshxs: nvm. different reason. enabled lingering instead

drbh:
following up on this PR, is this change still needed?

tmux should survive SSH disconnect by default, and this config doesn’t appear to set `KillUserProcesses=true` or anything else that would explain tmux being killed.

additionally im not sure if this change would resolve the issue anyway, since its setting linger to be on for the root user, and I believe the default user is nixos? lastly if we do want to enable linger I think we may want to set it in the config like `users.users.nixos.linger = true;`

@adarshxs let me know what your thoughts are. just trying to close/clean up stale PRs. 

Thanks!

adarshxs:
Yes we can close this

PR #364, opened 2026-03-16
Refine dependency specification

danieldk:
Refine the dependency specification, so that we can actually map a missing module to a package in error messages.

### DIFF
--- a/kernels/src/kernels/deps.py
+++ b/kernels/src/kernels/deps.py
@@ -13,7 +13,9 @@ except FileNotFoundError:
     )
 
 
-def validate_dependencies(dependencies: list[str], backend: Backend):
+def validate_dependencies(
+    kernel_module_name: str, dependencies: list[str], backend: Backend
+):
     """
     Validate a list of dependencies to ensure they are installed.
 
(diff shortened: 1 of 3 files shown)


drbh (dismissed):
lgtm, just looks like the test needs to be updated for the latest error message

kernels/src/kernels/deps.py
```diff
@@ -28,18 +28,31 @@ def validate_dependencies(dependencies: list[str], backend: Backend):
...
+                print("wut?")
+                # These are typically packages that do not provide any Python
+                # code, but get installed to Python's library dirctory. E.g.
+                # OneAPI.
+                continue
+
             if importlib.util.find_spec(module_name) is None:
                 raise ImportError(
-                    f"Kernel requires Python dependency `{python_package}`. Please install with: pip install {python_package}"
+                    f"Kernel requires Python dependency `{pkg_name}`. Please install with: pip install {pkg_name}"
```
sayakpaul: Is there any way to also provide the name of the kernel (maybe from its root module)?

sayakpaul (approved):
Can we also test this reliably?

kernels/src/kernels/deps.py
danieldk: At that point we only have the kernel module name, but that's usually the Pythonification of the kernel name, so I think it's already informative, so added that.

PR #367, opened 2026-03-17
replace workflow with huggingface/cuda-toolkit

sayakpaul:
Cc: @mfuntowicz

### DIFF
--- a/.github/workflows/build_kernel_windows.yaml
+++ b/.github/workflows/build_kernel_windows.yaml
@@ -37,7 +37,7 @@ jobs:
       - uses: actions/checkout@v6
 
       # CUDA environment setup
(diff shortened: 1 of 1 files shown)


.github/workflows/build_kernel_windows.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
       - uses: actions/checkout@v6
 
       # CUDA environment setup
-      - uses: Jimver/cuda-toolkit@v0.2.30
+      - uses: huggingface/cuda-toolkit@714c97b32958862237b96401fb253a4261453c3b # windows-only fork
```
danieldk: I think we can now add a tag and use it here?

sayakpaul: Done

PR #371, opened 2026-03-17
feat: remove docker files and references

drbh:
The PR removes docker support and references in the docs.

Docker has been removed because it adds an unneeded wrapper around the nix build. nix already sandboxes builds and does in a more reproducible and reliable way. Additionally docker support was always a bit hacky since it required workarounds to allow nix to run correctly.

The kernels project is focused on first class support of the nix builder and we want to avoid the buggy docker approach. For those interested in a docker build, please feel free to wrap the nix builder in your own containers, we just have decided to avoid maintaining this non recommend path in the kernels project itself.


.github/workflows/build_kernel.yaml
```diff
@@ -97,33 +97,3 @@ jobs:
             relu-kernel-cpu
             relu-backprop-compile-kernel
             silu-and-mul-kernel
-
-  test:
```
danieldk, line 101: We should keep the kernel tests though!

drbh: haha very very good point, mistake has been corrected! thanks for catching!!

nix-builder/tests/Dockerfile.test-kernel
```diff
@@ -1,78 +0,0 @@
-# syntax=docker/dockerfile:1.4
```
danieldk, line 1: This file shouldn't be removed, since it's used for running tests.

drbh: ah yea thanks for pointing out (totally missed in the previous fix) thanks for catching! 

fixed in the latest commits

PR #372, opened 2026-03-18
[card] read cuda capabilities from metadata.json

sayakpaul:


kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -81,6 +82,31 @@ def _extract_functions_from_all(init_file_path: Path) -> list[str] | None:
         return None
 
 
+def _read_cuda_capabilities_from_metadata(local_path: Path) -> set[str]:
+    """Read CUDA capabilities from metadata.json files in each build variant."""
+    cuda_capabilities: set[str] = set()
+    build_dir = local_path / "build"
```
danieldk, line 88: Only reading from the local `build` is going to give the wrong results. E.g. if we build and upload from macOS, it would clear all the capabilities, because none of the builds in the `build` directory have them (since they'll be Metal builds).

sayakpaul: What would you recommend then? Check if the `repo_id` parsed from the `build.toml` has any other builds?

PR #378, opened 2026-03-18
feat: add init and upload in kernel-builder cli

drbh:
~This PR is a WIP to add the init command to the kernel-builder rust cli. This branch mainly ports the init command from the python cli, with some improvements to the interface and updates to the template~

This PR adds an the `init` and `upload` command to the rust kernel-builder cli similar to implementation in the current python cli. It also adds a `build` command that is a facade for the `nix run` so devs can interact with a single tool for the full creation lifecycle of a kernel.

Mainly it aims to have better defaults to make the commands easier to use
1. the `init` command only requires the name of the kernels (folder) and the owner is pulled via a `whoami` request if the user is already logged into hf
2. the `init` command creates a new dir if a name is specified, if called within a dir, the cli will use the dir name as the kernel name 
3. the `upload` command does not require a repo id if the id can be read from the build
4. the `upload` command does not require a path to the build folder and will default to looking for the `build` dir if no explicit path is specified

example usage

init from inside of a existing dir
```bash
mkdir some-kernel
cd some-kernel
kernel-builder init
# Downloading template from kernels-community/template...
# Initialized `drbh/some-kernelt` at /home/drbh/Projects/kernels/kernel-builder/some-kernel
```

init from outside of a dir
```bash
kernel-builder init some-kernel
# Downloading template from kernels-community/template...
# Initialized `drbh/some-kernelt` at /home/drbh/Projects/kernels/kernel-builder/some-kernel
```

build
```bash
kernel-builder build
# ...
# some-kernel-torch-ext> no Makefile or custom installCheckPhase, doing nothing
# some-kernel-torch-ext> Checking of ABI compatibility
# some-kernel-torch-ext> 🐍 Checking for compatibility with manylinux_2_28 and Python ABI version 3.9
# some-kernel-torch-ext> ✅ No compatibility issues found
# some-kernel-torch-ext> Checking loading kernel with get_kernel
# some-kernel-torch-ext> Check whether the kernel can be loaded with get-kernel: some_kernel
# some-kernel-torch-ext> Running phase: removeBytecodeHook
# some-kernel-torch-ext> Removing Python bytecode
```

upload
```bash
kernel-builder upload
# Found 7 build variant(s) in /home/drbh/Projects/kernels/kernel-builder/some-kernel/build
# Using branch `v1` (new)
# Uploading 36 operations...
# Kernel upload successful. Find the kernel at: https://hf.co/drbh/some-kernel
```

### DIFF
--- a/kernel-builder/src/upload.rs
+++ b/kernel-builder/src/upload.rs
@@ -652,8 +652,7 @@ mod tests {
 
     const METADATA_V3: &str =
         r#"{"version": 3, "python-depends": [], "backend": {"type": "cuda"}}"#;
-    const METADATA_NO_VERSION: &str =
-        r#"{"python-depends": [], "backend": {"type": "cuda"}}"#;
+    const METADATA_NO_VERSION: &str = r#"{"python-depends": [], "backend": {"type": "cuda"}}"#;
 
     #[test]
     fn test_detect_branch_from_metadata() {


template/CARD.md
```diff
@@ -0,0 +1,38 @@
+---
```
sayakpaul: Do we want to maintain two copies? We have one in https://github.com/huggingface/kernels/blob/main/kernels/src/kernels/cli/card_template.md

sayakpaul (commented):
Very minor comments. No RUST expertise yet.

kernel-builder/src/build.rs
```diff
@@ -0,0 +1,89 @@
...
+
+#[derive(Debug, Args)]
+pub struct BuildArgs {
+    /// Directory of the kernel project (defaults to current directory).
+    #[arg(value_name = "KERNEL_DIR", default_value = ".")]
+    pub kernel_dir: PathBuf,
+
+    /// Nix flake target to run.
+    #[arg(long, default_value = "build-and-copy")]
+    pub target: BuildTarget,
```
danieldk: I think top-level

```
kernel-builder build
kernel-builder build-and-copy
kernel-builder build-and-upload
```

are nicer, easier to discover. I think we can use `#[command(flatten)]` to share arguments between different build commands, something like:

```
// ...
Build {
  #[command(flatten)]
  pub common_build_args: CommonBuildArgs,
  // ...
}

File: kernel-builder/src/build.rs [hunk 3]
```diff
@@ -0,0 +1,89 @@
...
+    #[arg(long, default_value = "4")]
+    pub max_jobs: u32,
+
+    /// Number of CPU cores per build job.
+    #[arg(long, default_value = "4")]
+    pub cores: u32,
+
+    /// Additional arguments passed through to `nix run`.
+    #[arg(last = true)]
+    pub nix_args: Vec<String>,
```
danieldk: Not sure if we want this?

kernel-builder/src/init.rs
```diff
@@ -0,0 +1,638 @@
...
+
+            // Try parsing the positional arg directly as "owner/repo".
+            let info = if let Ok(info) = RepoInfo::from_str(&path_str) {
+                info
+            } else {
+                // Bare name like "my-kernel" — look up the HF user as the owner.
+                let bare_name = path.file_name().and_then(OsStr::to_str).ok_or_else(|| {
+                    eyre::eyre!("Cannot determine directory name from `{path_str}`")
+                })?;
+                let owner = hf::whoami_username()?;
```
danieldk, line 139: nice!

kernel-builder/src/init.rs
```diff
@@ -0,0 +1,638 @@
...
+
+        let contents = fs::read(&entry_path)
+            .wrap_err_with(|| format!("Cannot read template file `{}`", entry_path.display()))?;
+
+        match String::from_utf8(contents) {
+            Ok(mut text) => {
+                for (from, to) in replacements {
+                    text = text.replace(from, to);
+                }
+                fs::write(&destination, text).wrap_err_with(|| {
```
danieldk: I possible (not sure if it supports everything needed), I think it would be nice to use the fileset API that is also used for generating pyproject files. The nice thing is that it is atomic, if something fails while still preparing the output, nothing is written.

I think we would have to extend it with the notion of files that should not be overwritten (if a project already exists and you only want to add new files). But I think making it impossible to have an operation half-completed is a nice feature.

kernel-builder/src/init.rs
```diff
@@ -0,0 +1,638 @@
...
+        return Ok(());
+    }
+
+    let build_toml = fs::read_to_string(build_toml_path)
+        .wrap_err_with(|| format!("Cannot read `{}`", build_toml_path.display()))?;
+    let mut document: toml::Value = toml::from_str(&build_toml)
+        .wrap_err_with(|| format!("Cannot parse TOML in `{}`", build_toml_path.display()))?;
+
+    // Update [general].backends
+    if let Some(general) = document.get_mut("general").and_then(|v| v.as_table_mut()) {
```
danieldk: I forget if we discussed this before, but it would be worthwhile putting jinja templates in the template, so that we can render everything editing TOML (but maybe parse + write to format).

This has the benefit that if we ever change the build.toml format, we don't have to change it both in the template and here in the code.

E.g., suppose that `backends` moves somewhere else, we have to change it in the template and here.

kernel-builder/src/init.rs
```diff
@@ -0,0 +1,638 @@
...
+        impl_block.push_str(condition);
+        impl_block.push_str("\n  ops.impl(\"");
+        impl_block.push_str(func_name);
+        impl_block.push_str("\", ");
+        impl_block.push_str(device);
+        impl_block.push_str(", ");
+        impl_block.push_str(ref_prefix);
+        impl_block.push_str(func_name);
+        impl_block.push_str(");\n");
+    }
```
danieldk: Could be a jinja template with `if` blocks.

kernel-builder/src/init.rs
```diff
@@ -0,0 +1,638 @@
...
+    };
+
+    let end = start + end_offset + END_MARKER.len();
+
+    let mut result = String::with_capacity(content.len());
+    result.push_str(&content[..start]);
+    result.push_str(replacement);
+    result.push_str(&content[end..]);
+    result
+}
```
danieldk: Would be easier to maintain with a jinja template and some if blocks I think.

kernel-builder/src/main.rs
```diff
@@ -24,6 +35,16 @@ struct Cli {
...
+    /// Initialize a new kernel project from template.
+    Init(InitArgs),
+
+    /// Build the kernel using Nix.
+    Build(BuildArgs),
+
+    /// Upload kernel build artifacts to the Hugging Face Hub.
+    Upload(UploadArgs),
+
+    #[command(hide = true)]
```
danieldk: If I understand correctly, this removes the subcommand from help. Sounds bad? Leftover from testing?

kernel-builder/src/upload.rs
```diff
@@ -0,0 +1,439 @@
...
+
+use crate::hf;
+
+const BUILD_COMMIT_BATCH_SIZE: usize = 1_000;
+
+#[derive(Debug, Args)]
+pub struct UploadArgs {
+    /// Directory of the kernel build (defaults to current directory).
+    #[arg(value_name = "KERNEL_DIR", default_value = ".")]
+    pub kernel_dir: PathBuf,
```
danieldk: Only a small nit, I'd use


```suggestion
    #[arg(value_name = "KERNEL_DIR")]
    pub kernel_dir: Option<PathBuf>,
```

It expresses the semantics a bit better, there is now a helper function to get the path from that. (No strong opinions though.)

kernel-builder/src/upload.rs
```diff
@@ -0,0 +1,439 @@
...
+        api.list_repo_files(&params)
+            .await
+            .wrap_err("Cannot list repository files")
+    })?;
+
+    // Collect all operations: benchmarks, README, and build artifacts.
+    let mut operations: Vec<CommitOperation> = Vec::new();
+
+    collect_benchmark_ops(&kernel_dir, &existing_files, is_new_branch, &mut operations)?;
+    collect_readme_ops(&kernel_dir, &mut operations);
```
danieldk: README should go to the `main` branch. I think probably the most elegant way is to make `operations` a `HashMap<String, Vec<CommitOperation>>` with the branch name as the key. That way we can have a nested loop to handle everything.

danieldk: For later: we may want to do the card generation and upload in one go as part of upload, since we cannot do the card generation in the Nix sandbox iff we need to fetch something from the main repo to do it (depends a bit on whether we still need to enumerate all the build variants or whether that information is handled by the Hub itself through the new kernel type, if the remaining information in the card is relatively static, we can handle it during build as well).

Not for this PR though.

kernel-builder/src/upload.rs
```diff
@@ -0,0 +1,439 @@
...
+            eprintln!("  Uploaded batch {}/{batch_count}.", batch_index + 1);
+        }
+    }
+
+    println!("Kernel upload successful. Find the kernel at: https://hf.co/{repo_id}");
+    Ok(())
+}
+
+// Collect benchmark file operations: add matching files, delete stale ones.
+fn collect_benchmark_ops(
```
danieldk: Maybe we should call them `commit_ops`? I was initially a bit confused, because we already use the term `ops` for ops in the kernel itself. Adding the `commit_` prefix may make it a bit clearer that we are referring to different ops here.

```suggestion
fn collect_benchmark_commit_ops(
```

kernel-builder/src/upload.rs
```diff
@@ -0,0 +1,439 @@
...
+        .and_then(|v| v.as_str())
+        .filter(|s| !s.is_empty())
+        .map(|s| s.to_owned())
+        .ok_or_else(|| {
+            eyre::eyre!(
+                "No `general.hub.repo-id` in `{}`. Use --repo-id to specify it.",
+                build_toml_path.display()
+            )
+        })
+}
```
danieldk: Use `util::parse_build`. We could also implement a `fn repo_id(&self) -> Option<&str>)` on `Build` if it is helpful and then we probably do not need this function at all.

kernel-builder/src/upload.rs
```diff
@@ -0,0 +1,439 @@
...
+            .map(|v| match v {
+                Some(n) => n.to_string(),
+                None => "none".to_owned(),
+            })
+            .collect();
+        bail!(
+            "Found multiple versions in build variants: {}",
+            version_strs.join(", ")
+        );
+    }
```
danieldk, line 414: Use `pyproject::Metadata` to deserialize `metadata.json`. If helpful, we could add a function similar to `parse_build`.

kernel-builder/src/upload.rs
```diff
@@ -0,0 +1,439 @@
...
+        let entry = entry?;
+        let path = entry.path();
+        if path.is_dir() {
+            walk_recursive(&path, files)?;
+        } else {
+            files.push(path);
+        }
+    }
+    Ok(())
+}
```
danieldk: `is_dir` follows symbolic links, so if a build directory contains a link like `ln -s . foo`, this will get stuck in an infinite loop (until it runs out of stack or heap). So we either need to skip symbolic links or we need loop detection. Maybe it's worth considering BurntSushi's `walkdir`, since it covers a lot of edge cases and doesn't follow symlinks by default. It also returns an iterator, which is more efficient than pushing everything into a `Vec`. Not that we care too much about performance here, but it is a nice property.

https://docs.rs/walkdir/latest/walkdir/

kernel-builder/src/upload.rs
sayakpaul: Some relevant internal discussions: https://huggingface.slack.com/archives/C090JN2P8NB/p1774236408982349

kernel-builder/src/build.rs
drbh: makes sense, updated in latest and added those three commands at the top level

kernel-builder/src/build.rs
drbh: yea I guess if the user wants more advance flags they should call nix run directly. removed in latest

kernel-builder/src/init.rs
drbh: oh yea thats a nice improvement, updated to use the file set api in the latest changes. this is a better ux and pattern so we avoid getting into a strange state

kernel-builder/src/init.rs
drbh: ah yes thats a great improvement. updated the kernel template to use jinja in the latest changes

kernel-builder/src/init.rs
drbh: updated to prefer jinja in latest

kernel-builder/src/init.rs
drbh: updated to prefer jinja in latest

kernel-builder/src/main.rs
drbh: removed in latest

kernel-builder/src/upload.rs
drbh: agreed this is a more idiomatic, updated in latest

kernel-builder/src/upload.rs
drbh: yea thats much cleaner, updated in latest! thanks

kernel-builder/src/upload.rs
drbh: nice point, added a `parse_metadata` function in latest commits.

kernel-builder/src/upload.rs
drbh: very good catch, updated to use walkdir in latest. thanks for pointing out

kernel-builder/src/pyproject/card.rs
```diff
@@ -0,0 +1,107 @@
...
+fn extract_functions(kernel_dir: &Path, module_name: &str) -> Option<Vec<String>> {
+    let init_path = kernel_dir
+        .join("torch-ext")
+        .join(module_name)
+        .join("__init__.py");
+
+    let content = fs::read_to_string(&init_path).ok()?;
+
+    let re = Regex::new(r#"__all__\s*=\s*\[\s*([^\]]*)\s*\]"#).ok()?;
+    let list_content = re.captures(&content)?.get(1)?.as_str();
```
danieldk: Regex parsing is really fragile. Not for this PR, but we should use a Python AST parser. Maybe tree-sitter-python or whatever ruff/ty uses (if it's available from crates.io). Maybe it's worthwhile creating a ticket?

kernel-builder/src/build.rs
```diff
@@ -0,0 +1,44 @@
...
+    #[arg(value_name = "KERNEL_DIR")]
+    pub kernel_dir: Option<PathBuf>,
+
+    /// Maximum number of Nix build jobs.
+    #[arg(long, default_value = "4")]
+    pub max_jobs: u32,
+
+    /// Number of CPU cores per build job.
+    #[arg(long, default_value = "4")]
+    pub cores: u32,
```
danieldk: Should use `Option`, we want to use the default from `nix.conf` when the user has not specified any arguments. E.g. I have these turned on my builder for the number of cores/RAM available.

kernel-builder/src/main.rs
```diff
@@ -12,6 +12,17 @@ use completions::print_completions;
 mod develop;
 use develop::{devshell, testshell};
 
+mod build;
+use build::{run_build, CommonBuildArgs};
```
danieldk:
```suggestion
use build::run_build;
```

kernel-builder/src/main.rs
```diff
@@ -34,7 +45,47 @@ enum Commands {
...
     Completions { shell: Shell },
 
-    /// Create pyproject and CMake files for a kernel development.
+    /// Initialize a new kernel project from template.
+    Init(InitArgs),
+
+    /// Build the kernel locally (alias for build-and-copy).
+    Build {
+        #[command(flatten)]
+        args: CommonBuildArgs,
```
danieldk: Use https://github.com/huggingface/kernels/blob/826d53607e09536c198178a2bc1294cd68ba7ec1/kernel-builder/src/main.rs#L26

kernel-builder/src/main.rs
```diff
@@ -125,6 +176,28 @@ fn main() -> Result<()> {
...
+        }
+        Commands::BuildAndUpload {
+            build_args,
+            repo_id,
+            branch,
+            private,
+            repo_type,
+        } => {
+            let kernel_dir = build_args.kernel_dir.clone();
+            run_build(build_args, "build-and-copy")?;
```
danieldk:
```suggestion
```

Remove, the idea behind `build-and-upload` is to upload directly from the Nix store without using duplicate storage for both the store version and the `build` version.

kernel-builder/src/main.rs
```diff
@@ -125,6 +176,28 @@ fn main() -> Result<()> {
             print_completions(&mut Cli::command(), shell);
             Ok(())
         }
+        Commands::Init(args) => run_init(args),
+        Commands::Upload(args) => run_upload(args),
+        Commands::Build { args } | Commands::BuildAndCopy { args } => {
+            run_build(args, "build-and-copy")
```
danieldk: I think I would only do a `nix build` for this subcommand. This unnecessarily creates another copy, which can be very large (e.g. FA) and the kernels tools work with the result directory/symlink as well (at least `kernels upload`).

kernel-builder/src/upload.rs
```diff
@@ -0,0 +1,422 @@
...
+
+/// Recursively walk a directory and return all file paths.
+fn walk_files(dir: &Path) -> impl Iterator<Item = PathBuf> {
+    WalkDir::new(dir)
+        .into_iter()
+        .filter_map(|e| e.ok())
+        .filter(|e| e.file_type().is_file())
+        .map(|e| e.into_path())
+}
```
danieldk, line 430: We should probably copy the upload test from the Python tests?

kernel-builder/src/pyproject/card.rs
drbh: good point, I opt'ed to use https://crates.io/crates/rustpython-parser which is what ruff forked, only because ruff's parser has git dependencies and our use case is relatively simple

kernel-builder/src/build.rs
drbh: good point, agree that option is a better pattern (as noted earlier) thanks for catching. updated to prefer the `NixArgs` struct where these values are optional

kernel-builder/src/main.rs
drbh: removed in latest changes

kernel-builder/src/main.rs
drbh: nice! updated to use this struct

kernel-builder/src/main.rs
drbh: good point, updated to only call build then upload from the symlinked result dir

kernel-builder/src/main.rs
drbh: makes sense, updated in latest

kernel-builder/src/upload.rs
drbh: agreed, copied in latest changes

kernel-builder/src/upload.rs
sayakpaul: @drbh I think the most sensible thing for now would be to keep the card contents limited to what we can fetch from the build (after the build is successful). We will render capabilities, frameworks etc. on the Hub side upon each commit, anyway. So, the card could contain an example snippet, available function interfaces, etc. This is all implemented in Python CLI. 

Totally okay if we want to do the segregation in a separate PR.

sayakpaul: And then there is this other thing of always uploading the card to `main` revision.

kernel-builder/src/pyproject/card.rs
```diff
@@ -0,0 +1,153 @@
...
+use kernels_data::config::Build;
+
+fn extract_functions(kernel_dir: &Path, module_name: &str) -> Option<Vec<String>> {
+    let init_path = kernel_dir
+        .join("torch-ext")
+        .join(module_name)
+        .join("__init__.py");
+
+    let content = fs::read_to_string(&init_path).ok()?;
+    let stmts = ast::Suite::parse(&content, "<module>").ok()?;
```
danieldk, line 16: Nice!!!

kernel-builder/src/upload.rs
drbh: > @drbh I think the most sensible thing for now would be to keep the card contents limited to what we can fetch from the build (after the build is successful). 

oh we are actually writing the card in the build step now! for exactly the reasons you mention above. Basically the card just extracts the functions from __all__ and shows snippets/any description the user adds

> And then there is this other thing of always uploading the card to main revision. 

and we also do now push the card to main https://github.com/huggingface/kernels/blob/6e14158fe3651b186d047dbad6fd767b1861f8ba/kernel-builder/src/upload.rs#L140

sayakpaul: The Python CLI does have a template for the snippet and derives it without having to rely on any fragile `re`. Are we replicating that as well? 

https://github.com/huggingface/kernels/blob/6e14158fe3651b186d047dbad6fd767b1861f8ba/kernels/src/kernels/cli/kernel_card_utils.py#L177-L180

drbh: yep thats is done in this file https://github.com/huggingface/kernels/blob/6e14158fe3651b186d047dbad6fd767b1861f8ba/kernel-builder/src/pyproject/card.rs#L52 and we also now avoid `re` for extracting the functions from `__all__` by parsing the python ast. so generally I think this approach should be solid

sayakpaul: Perfect. I think that should be it.

PR #379, opened 2026-03-19
[builder] feat: implement upstream field for build.toml.

sayakpaul:
Similar to https://github.com/huggingface/kernel-builder/pull/343

### DIFF
--- a/kernel-builder/src/config/compat.rs
+++ b/kernel-builder/src/config/compat.rs
@@ -5,10 +5,11 @@ use serde_value::Value;
 use super::{v1, v2, v3, Build};
 
 #[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
 pub enum BuildCompat {
     V1(v1::Build),
     V2(v2::Build),
-    V3(Box<v3::Build>),
+    V3(v3::Build),
 }
 
 impl<'de> Deserialize<'de> for BuildCompat {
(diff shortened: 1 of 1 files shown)


kernel-builder/src/config/v3.rs
```diff
@@ -38,6 +38,8 @@ pub struct General {
 
     pub license: Option<String>,
 
+    pub upstream: Option<String>,
```
danieldk: If this is a URL, we should consider using the `url` crate. I think it also supports serde, so deserialization should not be an issue. Gives us validation of what people put in `upstream` for free.

sayakpaul: Wow, cool. Let me try it out!

kernel-builder/src/config/compat.rs
```diff
@@ -8,7 +8,7 @@ use super::{v1, v2, v3, Build};
 pub enum BuildCompat {
     V1(v1::Build),
     V2(v2::Build),
-    V3(v3::Build),
+    V3(Box<v3::Build>),
```
sayakpaul: Otherwise https://github.com/huggingface/kernels/actions/runs/23290500058/job/67724598491?pr=379. Any better way?

danieldk: I would just use `#[allow(clippy::large_enum_variant)]` on the enum. This isn't performance-sensitive code and having a Box in between is annoying.

danieldk (approved):

PR #383, opened 2026-03-19
Backport `kernels` changes for FA4 support

danieldk:


kernels/src/kernels/deps.py
```diff
@@ -31,13 +33,27 @@ def validate_dependencies(dependencies: list[str], backend: str):
...
 
         # Check if each python package is installed
         for python_package in python_packages:
             # Convert package name to module name (replace - with _)
-            module_name = python_package.replace("-", "_")
+            pkg_name = python_package.get("pkg")
+            # Assertion because this should not happen and is a bug.
+            assert (
+                pkg_name is not None
+            ), f"Invalid dependency data for `{dependency}`: missing `pkg` field."
```
sayakpaul, line 47: Yes, this is what confused me yesterday during debugging

kernels/src/kernels/python_depends.json
```diff
@@ -2,29 +2,33 @@
   "general": {
     "einops": {
       "nix": ["einops"],
-      "python": ["einops"]
```
sayakpaul, line 5: Do we generate this manually? Would it be prudent to generate it automatically if it makes sense?

danieldk: Yes, this is hand-written. I don't know an easy way to do it automatically, also not sure if it is needed, I don't think we should go beyond a handful of dependencies. Otherwise we are back to Python wheels.

kernels/src/kernels/deps.py
danieldk: But this one shouldn't occur, unless the JSON file is corrupt somehow.

PR #388, opened 2026-03-23
kernel-builder: add `completions` subcommand for generating shell completions

danieldk:
The new `completion` subcommand creates completions for your favorite shell. Also hook it up in Nix, so that dev shells have completions.

### DIFF
--- a/nix-builder/pkgs/kernel-builder/default.nix
+++ b/nix-builder/pkgs/kernel-builder/default.nix
@@ -1,5 +1,6 @@
 {
   lib,
+  installShellFiles,
   rustPlatform,
   pkg-config,
   libgit2,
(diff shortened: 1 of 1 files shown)


kernel-builder/src/main.rs
```diff
@@ -2,9 +2,13 @@ use std::fs::File;
 use std::io::{BufWriter, Write};
 use std::path::PathBuf;
 
-use clap::{Parser, Subcommand};
+use clap::{CommandFactory, Parser, Subcommand};
```
sayakpaul, line 5: Can the users configure anything pertaining to autocompleitions?

danieldk: No, you just let the command generate it and let the shell load it. Since completions are standardized, I don't think there is much to configure.

PR #391, opened 2026-03-23
Fix handling of noarch and tvm-ffi variants

### LINKED ISSUE #356: Fix noarch devshell + testshell

danieldk:
- For development and test shells, we want the arch variant names, because we'd like to be able to test a noarch kernel with multiple Torch/CUDA versions. This change also fixes `nix develop` without specifying a variant.
- For kernel CI tests, make sure that we generate the tvm-ffi variant names for tvm-ffi kernels.

Fixes #356

### DIFF
--- a/nix-builder/lib/extension/torch/no-arch.nix
+++ b/nix-builder/lib/extension/torch/no-arch.nix
@@ -100,5 +100,6 @@ stdenv.mkDerivation (prevAttrs: {
   passthru = {
     inherit dependencies;
     variant = torch.noarchVariant;
+    archVariant = torch.variant;
   };
 })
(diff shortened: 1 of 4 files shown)


nix-builder/lib/extension/torch/no-arch.nix
```diff
@@ -100,5 +100,6 @@ stdenv.mkDerivation (prevAttrs: {
   passthru = {
     inherit dependencies;
     variant = torch.noarchVariant;
+    archVariant = torch.variant;
```
sayakpaul, line 103: Out of curiosity. I find `python3.pkgs.torch.variant` more consistent with how it's being done for TVM FFI.

danieldk: We should probably do this project-wide for `buildSet.torch` at some point. This originates from when we didn't overlay Torch, so `python3.pkgs.torch` would not be the derivation that we want.

PR #393, opened 2026-03-24
feat: implement strict type validation using `strict` from `huggingface_hub`

sayakpaul:
As discussed internally.

Note that I have chosen NOT to use `pydantic` for this as per https://huggingface.co/docs/huggingface_hub/package_reference/dataclasses#why-not-use-pydantic--or-attrs--or-marshmallowdataclass-. Plus, we don't have to add a new dependency and get away with what we already have (which is always nice). But I can switch to `pydantic` should we mutually decide.

Some inline comments.

## Notes

* Bump `huggingface_hub` minimum version to use `@strict`. 
* `@strict` can't validate packaging.version.Version (a third-party type). So, I had to remove `@strict` from the `Torch` class, for example, in `variants.py`. If we decided to go the `@strict` way, we could follow up with the Hub team for adding support.
* Had to add some dependencies to `overlay.nix` due to the version bump in Hugging Face Hub. I think that they should be quite safe.
* Didn't do the nix related updates in terraform yet in case we want to use pydantic.

### DIFF
--- a/nix-builder/overlay.nix
+++ b/nix-builder/overlay.nix
@@ -130,6 +130,11 @@ in
           xpuPackages = final.xpuPackages_2025_3;
         };
 
+        torch-bin_2_11 = mkTorch {
+          version = "2.11";
+          xpuPackages = final.xpuPackages_2025_3;
+        };
+
         triton-xpu_2_9 = callPackage ./pkgs/python-modules/triton-xpu {
           torchVersion = "2.9";
           xpuPackages = final.xpuPackages_2025_2;
--- a/.github/workflows/build_kernel_macos.yaml
+++ b/.github/workflows/build_kernel_macos.yaml
@@ -28,7 +28,7 @@ jobs:
       # For now we only test that there are no regressions in building macOS
       # kernels. Also run tests once we have a macOS runner.
       - name: Build relu kernel
-        run: ( cd examples/kernels/relu && nix build .\#redistributable.torch29-metal-aarch64-darwin -L )
+        run: ( cd examples/kernels/relu && nix build .\#redistributable.torch210-metal-aarch64-darwin -L )
 
       - name: Build relu metal cpp kernel
-        run: ( cd examples/kernels/relu-metal-cpp && nix build .\#redistributable.torch29-metal-aarch64-darwin -L )
+        run: ( cd examples/kernels/relu-metal-cpp && nix build .\#redistributable.torch210-metal-aarch64-darwin -L )
--- a/docs/source/builder/build-variants.md
+++ b/docs/source/builder/build-variants.md
@@ -8,42 +8,42 @@ available. This list will be updated as new PyTorch versions are released.
 ## CPU aarch64-darwin
 
 - `torch210-cpu-aarch64-darwin`
-- `torch29-cpu-aarch64-darwin`
+- `torch211-cpu-aarch64-darwin`
 
 ## Metal aarch64-darwin
 
 - `torch210-metal-aarch64-darwin`
-- `torch29-metal-aarch64-darwin`
+- `torch211-metal-aarch64-darwin`
 
 ## CPU aarch64-linux
 
 - `torch210-cxx11-cpu-aarch64-linux`
-- `torch29-cxx11-cpu-aarch64-linux`
+- `torch211-cxx11-cpu-aarch64-linux`
 
 ## CUDA aarch64-linux
 
 - `torch210-cxx11-cu126-aarch64-linux`
 - `torch210-cxx11-cu128-aarch64-linux`
 - `torch210-cxx11-cu130-aarch64-linux`
-- `torch29-cxx11-cu126-aarch64-linux`
-- `torch29-cxx11-cu128-aarch64-linux`
+- `torch211-cxx11-cu126-aarch64-linux`
+- `torch211-cxx11-cu128-aarch64-linux`
+- `torch211-cxx11-cu130-aarch64-linux`
 - `torch29-cxx11-cu129-aarch64-linux`
-- `torch29-cxx11-cu130-aarch64-linux`
 
 ## CPU x86_64-linux
 
 - `torch210-cxx11-cpu-x86_64-linux`
-- `torch29-cxx11-cpu-x86_64-linux`
+- `torch211-cxx11-cpu-x86_64-linux`
 
 ## CUDA x86_64-linux
 
 - `torch210-cxx11-cu126-x86_64-linux`
 - `torch210-cxx11-cu128-x86_64-linux`
 - `torch210-cxx11-cu130-x86_64-linux`
-- `torch29-cxx11-cu126-x86_64-linux`
-- `torch29-cxx11-cu128-x86_64-linux`
+- `torch211-cxx11-cu126-x86_64-linux`
+- `torch211-cxx11-cu128-x86_64-linux`
+- `torch211-cxx11-cu130-x86_64-linux`
 - `torch29-cxx11-cu129-x86_64-linux`
-- `torch29-cxx11-cu130-x86_64-linux`
 
 ## ROCm x86_64-linux
 
--- a/examples/kernels/flake.nix
+++ b/examples/kernels/flake.nix
@@ -15,7 +15,7 @@
       inherit (kernel-builder.inputs.nixpkgs) lib;
 
       cudaVersion = "cu126";
-      torchVersion = "29";
+      torchVersion = "210";
       tvmFfiVersion = "01";
 
       # All example kernels to build in CI.
@@ -79,7 +79,7 @@
           drv = sys: out: out.packages.${sys}.default;
           torchVersions = _defaultVersions: [
             {
-              torchVersion = "2.9";
+              torchVersion = "2.10";
               cudaVersion = "12.8";
               systems = [
                 "x86_64-linux"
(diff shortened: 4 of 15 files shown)


kernels/src/kernels/backends.py
```diff
@@ -3,13 +3,15 @@
...
-from typing import ClassVar, Optional, Protocol
+from typing import ClassVar, Optional, Protocol, runtime_checkable
 
+from huggingface_hub.dataclasses import strict
 from packaging.version import Version
 
 from kernels.compat import has_torch
 
 
+@runtime_checkable
```
sayakpaul, line 14: Because `@strict` validates field types using `isinstance()` checks at construction time. When `Arch.backend` is typed as `Backend` (a `Protocol`), `@strict` tries to do `isinstance(value, Backend)` — and Python's `typing.Protocol` raises TypeError unless it's decorated with `@runtime_checkable`.

kernels/src/kernels/cli/benchmark.py
```diff
@@ -83,7 +85,18 @@ class TimingResults:
     verified: bool | None = None  # None = no verify fn, True = passed, False = failed
     ref_mean_ms: float | None = None  # Reference implementation mean time
 
+    def validate_iterations(self):
```
sayakpaul, line 88: These are "magic" methods. `@strict` auto-discovers any method named `validate_*` and calls it during `__init__` (and on `__setattr__`). They replaced the manual checks that were previously inside `from_dict()`.

kernels/pyproject.toml
```diff
@@ -10,7 +10,7 @@ license = { text = "Apache-2.0" }
 readme = "README.md"
 requires-python = ">= 3.9"
 dependencies = [
-  "huggingface_hub>=0.26.0,<2.0",
+  "huggingface_hub>=0.31.4,<2.0",
```
sayakpaul: Because this is the version where `strict` was introduced. I think this is a relatively simple change.

nix-builder/overlay.nix
```diff
@@ -77,6 +77,22 @@ in
...
+          version = "1.3.0";
+          src = python-super.fetchPypi {
+            pname = "huggingface_hub";
+            inherit version;
+            hash = "sha256-KJ4qNYb98B41iClE6qBvvVdDbeJLbmU9H6skhYSs1ms=";
+          };
+          dependencies = (old.dependencies or [ ]) ++ [
+            python-self.httpx
+            python-self.shellingham
+            python-self.typer-slim
```
sayakpaul, line 90: Can add their `hash` as well.

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -22,18 +24,85 @@
 LIBRARY_NAME = "kernels"
 
 
-def _parse_build_toml(local_path: str | Path) -> dict | None:
-    local_path = Path(local_path)
-    build_toml_path = local_path / "build.toml"
+@strict
+@dataclass
+class HubConfig:
+    repo_id: str = ""
```
danieldk:
```suggestion
    repo_id: str | None
```

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -22,18 +24,85 @@
...
+
+    @staticmethod
+    def from_dict(data: dict) -> "HubConfig":
+        return HubConfig(repo_id=data.get("repo-id", ""))
```
danieldk:
```suggestion
        return HubConfig(repo_id=data.get("repo-id"))
```

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -22,18 +24,85 @@
...
+
+
+@strict
+@dataclass
+class GeneralConfig:
+    name: str = ""
+    version: int | None = None
+    license: str | None = None
+    backends: list[str] | None = None
+    hub: HubConfig | None = None
```
danieldk:
```suggestion
    version: int | None
    license: str | None
    backends: list[str] | None
    hub: HubConfig | None
```

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -22,18 +24,85 @@
...
+
+    @staticmethod
+    def from_dict(data: dict) -> "GeneralConfig":
+        hub_data = data.get("hub")
+        return GeneralConfig(
+            name=data.get("name", ""),
```
danieldk:
```suggestion
            name=data.get("name"),
```

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -22,18 +24,85 @@
...
+            version=data.get("version"),
+            license=data.get("license"),
+            backends=data.get("backends"),
+            hub=HubConfig.from_dict(hub_data) if hub_data else None,
```
danieldk, line 56: Note on the discussion from Slack, this is one thing that goes wrong with hand-rolled deserialization. What if the user put in something that makes `hub_data` truthy, but not a dict. Then `HubConfig.from_dict` will fail with a hard-to-understand error, whereas it should just say what the field/section is expected to be.

You can avoid this by programming very defensively (e.g. here checking that it is a dict first), but those things you get for free from a library that does deserialization for you.

I think for now it's ok, since we are going to replace this deserialization with Rust as discussed.

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -22,18 +24,85 @@
...
 
-    try:
-        with open(build_toml_path, "rb") as f:
-            return tomllib.load(f)
-    except Exception:
-        return None
+@strict
+@dataclass
+class KernelConfig:
+    cuda_capabilities: list[str] | None = None
```
danieldk:
```suggestion
    cuda_capabilities: list[str] | None
```

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -22,18 +24,85 @@
...
+    def from_dict(data: dict) -> "KernelConfig":
+        return KernelConfig(cuda_capabilities=data.get("cuda-capabilities"))
+
+
+@strict
+@dataclass
+class BuildConfig:
+    general: GeneralConfig = field(default_factory=GeneralConfig)
+    kernel: dict[str, KernelConfig] | None = None
+    upstream: str | None = None
```
danieldk:
```suggestion
    kernel: dict[str, KernelConfig] | None
    upstream: str | None
```

danieldk: Removing all the initializers, since we want these to be instance variables, not class variables.

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -88,8 +157,10 @@ def _parse_repo_id(local_path: str | Path) -> str | None:
     if not config:
         return None
 
-    repo_id = config.get("general", {}).get("hub", {}).get("repo-id", None)
-    return repo_id
+    if config.general.hub is None:
+        return None
+    repo_id = config.general.hub.repo_id
+    return repo_id if repo_id else None
```
danieldk:
```suggestion
    return config.general.hub.repo_id
```

kernels/src/kernels/cli/kernel_card_utils.py
```diff
@@ -151,12 +219,12 @@ def _build_kernel_card_vars(
 def _update_kernel_card_license(
     kernel_card: ModelCard, local_path: str | Path
```
danieldk: Maybe we should


```suggestion
    kernel_card: ModelCard, local_path: Path
```

and make the caller responsible for the conversion?

kernels/src/kernels/layer/device.py
```diff
@@ -34,6 +37,12 @@ class CUDAProperties:
     min_capability: int
     max_capability: int
 
+    def validate_capability_range(self):
+        if self.min_capability > self.max_capability:
+            raise ValueError(
+                f"min_capability ({self.min_capability}) must be <= max_capability ({self.max_capability})"
+            )
```
danieldk, line 44: Nice!

kernels/src/kernels/deps.py
```diff
@@ -1,12 +1,65 @@
...
+from huggingface_hub.dataclasses import strict
+
 from kernels.backends import Backend
 
+
+@strict
+@dataclass
+class PythonPackage:
+    pkg: str
+    import_name: str | None = None
```
danieldk:
```suggestion
    import_name: str | None
```

kernels/src/kernels/deps.py
```diff
@@ -1,12 +1,65 @@
...
+        return PythonPackage(
+            pkg=data["pkg"],
+            import_name=data.get("import"),
+        )
+
+
+@strict
+@dataclass
+class DependencyInfo:
+    nix: list
```
danieldk:
```suggestion
    nix: list[str]
```

kernels/src/kernels/deps.py
```diff
@@ -1,12 +1,65 @@
...
+    python: list
```
danieldk:
```suggestion
    python: list[str]
```

nix-builder/overlay.nix
danieldk: Not needed if we use them from nixpkgs as-is.

kernels/src/kernels/cli/kernel_card_utils.py
sayakpaul: Made a note about it as well in bc042be.

PR #397, opened 2026-03-24
kernel-builder: share common Nix options and add option to print build logs

danieldk:


kernel-builder/src/develop.rs
```diff
@@ -35,15 +38,23 @@ pub fn devshell(
...
     cores: Option<u32>,
+    print_build_logs: bool,
 ) -> Result<()> {
-    run_develop(kernel_dir, max_jobs, cores, Some("test".to_string()))
+    run_develop(
+        kernel_dir,
+        max_jobs,
+        cores,
+        print_build_logs,
+        Some("test".to_string()),
```
sayakpaul, line 58: Wonder why is this needed.

danieldk: "test" because it's a test shell, `to_string` because `run_develop` wants an owned string. I'll change it to a string reference.

danieldk: Forgot: not worth it, we have to do the conversion at some point, because the string gets stored in a struct.

PR #398, opened 2026-03-25
ci: speed up and improve kernels tests CI

danieldk:
* Do not build test image inside a buildx container. We don't use its features and it adds a lot of serialization time.
* Build all kernels using a single Nix command to get build parallelism.
* Make kernel test failures fail the workflow.
* Fix the failures that were silently discarded. Also uncovered a bug in the `compat` module used by all kernels.

### DIFF
--- a/examples/kernels/flake.nix
+++ b/examples/kernels/flake.nix
@@ -29,17 +29,20 @@
         {
           name = "relu-kernel";
           path = ./relu;
-          drv = sys: out: out.packages.${sys}.redistributable.${"torch${torchVersion}-cxx11-${cudaVersion}-${sys}"};
+          drv =
+            sys: out: out.packages.${sys}.redistributable.${"torch${torchVersion}-cxx11-${cudaVersion}-${sys}"};
         }
         {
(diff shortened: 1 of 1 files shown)


.github/workflows/build_kernel.yaml
```diff
@@ -118,17 +73,11 @@ jobs:
...
       - name: Build Docker image
-        uses: docker/build-push-action@v6
-        with:
-          context: .
-          file: nix-builder/tests/Dockerfile.test-kernel
-          platforms: linux/amd64
-          load: true
-          push: false
-          tags: kernel-builder:latest
+        run: |
```
sayakpaul, line 77: I guess you meant 

> Do_not_ build test image inside a buildx container. We don't use its features and it adds a lot of serialization time.

examples/kernels/flake.nix
```diff
@@ -0,0 +1,159 @@
+{
+  description = "All example kernels";
```
sayakpaul, line 2: TIL. Thanks for this.

PR #401, opened 2026-03-26
Add ROCm support for Torch 2.11

danieldk:
Adds ROCm support for Torch 2.11. In order to do so:

* Add ROCm 7.2.1.
* Add aotriton 0.11.2b.
* Remove old ROCm/aotriton versions.

### DIFF
--- a/.github/workflows/build_kernel_rocm.yaml
+++ b/.github/workflows/build_kernel_rocm.yaml
@@ -34,7 +34,7 @@ jobs:
       # For now we only test that there are no regressions in building ROCm
       # kernels. Also run tests once we have a ROCm runner.
       - name: Build relu kernel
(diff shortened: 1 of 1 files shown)


nix-builder/pkgs/aotriton/default.nix
```diff
@@ -32,68 +32,21 @@ let
     };
 in
 {
-  aotriton_0_10 = generic rec {
```
sayakpaul, line 35: Do we not need it anymore?

sayakpaul (approved):
All good pending the test failures.

nix-builder/pkgs/aotriton/default.nix
danieldk: No, Torch 2.10 and 2.11 use aotriton 0.11.1 and 0.11.2 respectively.

sayakpaul: Ah yeah. I see the modifications now.

PR #406, opened 2026-03-26
Use `kernel-builder upload` in the Nix builder

danieldk:
To get there, support uploads without `build.toml`.

### DIFF
--- a/kernel-builder/src/upload.rs
+++ b/kernel-builder/src/upload.rs
@@ -72,7 +72,7 @@ pub fn run_upload(args: UploadArgs) -> Result<()> {
     let arg_repo_id = match args.repo_id {
         Some(id) => id,
         None =>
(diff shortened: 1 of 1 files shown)


kernel-builder/src/upload.rs
```diff
@@ -69,15 +69,23 @@ pub fn run_upload(args: UploadArgs) -> Result<()> {
...
     let arg_repo_id = match args.repo_id {
         Some(id) => id,
-        None => build
-            .repo_id()
-            .ok_or_else(|| {
-                eyre::eyre!("No `general.hub.repo-id` in build.toml. Use --repo-id to specify it.")
-            })?
-            .to_owned(),
+        None =>
+        // WARN: parsing must not be moved out op this branch, we want users
```
sayakpaul: Sorry

```suggestion
        // WARN: parsing must not be moved out of this branch, we want users
```

PR #409, opened 2026-03-27
Add XPU support for Torch 2.11

danieldk:
Changes:

* Update RHEL to Nix script to support multiple versions of the same package names.
* Add oneAPI 2025.3.2.
* Add Torch 2.11 XPU.
* Build `triton-xpu` with binary wheels, replacing outdated versions.
* Fix issue with wheel dependency relaxation hook (e.g. `triton-xpu` getting relaxed to `triton`).
* Remove old, now-unused oneAPI versions.

### DIFF
--- a/.github/workflows/build_kernel_xpu.yaml
+++ b/.github/workflows/build_kernel_xpu.yaml
@@ -34,13 +34,13 @@ jobs:
       # For now we only test that there are no regressions in building XPU
       # kernels. Also run tests once we have a XPU runner.
       - name: Build relu kernel
-        run: ( cd examples/kernels/relu && nix build .\#redistributable.torch210-cxx11-20253-x86_64-linux -L )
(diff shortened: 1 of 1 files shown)


nix-builder/pkgs/python-modules/torch/binary/generic.nix
```diff
@@ -228,26 +229,46 @@ buildPythonPackage {
...
+            "nvidia-nccl"
+            "nvidia-nccl"
+            "nvidia-nvshmem"
+            "nvidia-nvshmem"
+            "nvidia-nvtx"
+            "nvidia-nvjitlink"
+            "nvidia-cufile"
+          ];
+          suffix = [
+            ""
```
drbh, line 261: is this empty suffix used? just looks a bit strange

danieldk: Yep, in some cases the dependencies do not have a `-cuNN` suffix. But it's really weird, because the package might have it for CUDA 12 and not for CUDA 13.

PR #416, opened 2026-03-30
[ci] e2e kernel builder tests

sayakpaul:
Adds fast e2e testing for the kernel builder CLI:

* Initialization + build
* Upload the build to the Hub
* Download the build from the Hub (through `get_kernel()`) and use it

To keep the CI runtime manageable, we only build against a single backend. But I think we should also build against a single variant now that https://github.com/huggingface/kernels/pull/419/ is in.

### DIFF
--- a/docs/source/builder/nix.md
+++ b/docs/source/builder/nix.md
@@ -1,5 +1,10 @@
 # Using the kernel builder with Nix
 
+> [!NOTE] 
+> The [install script](writing-kernels.md#quick-install) automates
+> the Nix and kernel-builder setup described below. Use these manual
+> instructions if you prefer step-by-step control.
+
 The kernel builder uses Nix for building kernels. You can build or
 run the kernels directly if you have Nix installed on your system.
 We recommend installing Nix in the following way:
--- a/docs/source/builder/writing-kernels.md
+++ b/docs/source/builder/writing-kernels.md
@@ -37,6 +37,34 @@ support.
 
 ## Setting up environment
 
+### Quick install
+
+The fastest way to get started is to run the install script. This
+installs [Determinate Nix](https://docs.determinate.systems/determinate-nix/)
+and `kernel-builder` in a single command:
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/huggingface/kernels/main/install.sh | bash
+```
+
+This will:
+
+1. Install Determinate Nix (if not already installed).
+2. Configure the Hugging Face binary cache (to avoid building dependencies from
+   source).
+3. Install `kernel-builder` via `nix profile install`.
+
+To update `kernel-builder` later:
+
+```bash
+nix profile upgrade --all
+```
+
+For a step-by-step breakdown of what the script does, see
+[Using the kernel builder with Nix](nix.md).
+
+### Cloud environment
+
 In the [`terraform`](https://github.com/huggingface/kernels/tree/main/terraform) directory, we provide an
 example of programatically spinning up an EC2 instance that is ready
 with everything needed for you to start developing and building
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "flake-compat": {
       "locked": {
-        "lastModified": 1765121682,
-        "narHash": "sha256-4VBOP18BFeiPkyhy9o4ssBNQEvfvv1kXkasAYd0+rrA=",
+        "lastModified": 1767039857,
+        "narHash": "sha256-vNpUSpF5Nuw8xvDLj2KCwwksIbjua2LZCqhV1LNRDns=",
         "owner": "edolstra",
         "repo": "flake-compat",
-        "rev": "65f23138d8d09a92e30f1e5c87611b23ef451bf3",
+        "rev": "5edf11c44bc78a0d334f6334cdaf7d60d732daab",
         "type": "github"
       },
       "original": {
@@ -35,11 +35,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1766341660,
-        "narHash": "sha256-4yG6vx7Dddk9/zh45Y2KM82OaRD4jO3HA9r98ORzysA=",
+        "lastModified": 1774935083,
+        "narHash": "sha256-Mh6bLcYAcENBAZk3RoMPMFCGGMZmfaGMERE4siZOgP4=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "26861f5606e3e4d1400771b513cc63e5f70151a6",
+        "rev": "2f4fd5e1abf9bac8c1d22750c701a7a5e6b524c6",
         "type": "github"
       },
       "original": {
@@ -64,11 +64,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1769050281,
-        "narHash": "sha256-1H8DN4UZgEUqPUA5ecHOufLZMscJ4IlcGaEftaPtpBY=",
+        "lastModified": 1774926780,
+        "narHash": "sha256-JMdDYn0F+swYBILlpCeHDbCSyzqkeSGNxZ/Q5J584jM=",
         "owner": "oxalica",
         "repo": "rust-overlay",
-        "rev": "6deef0585c52d9e70f96b6121207e1496d4b0c49",
+        "rev": "962a0934d0e32f42d1b5e49186f9595f9b178d2d",
         "type": "github"
       },
       "original": {
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,150 @@
+#!/bin/bash
+set -euo pipefail
+
+# kernel-builder installer
+# Usage: curl -fsSL https://raw.githubusercontent.com/huggingface/kernels/main/install.sh | bash
+
+FLAKE_REF="github:huggingface/kernels"
+NIX_PROFILE_SCRIPT="/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh"
+
+# --- Colors (respect NO_COLOR) ---
+
+if [ -z "${NO_COLOR:-}" ] && [ -t 1 ]; then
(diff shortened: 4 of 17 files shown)


.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,155 @@
...
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    paths:
+      - "kernel-builder/**"
+      - "kernels/src/**"
+      - "nix-builder/**"
+      - "kernels-data/**"
+      - ".github/workflows/test_e2e.yaml"
```
sayakpaul, line 13: Trigger on core changes.

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  init-build-upload:
+    name: Init, build, and upload kernel
+    outputs:
+      torch_version: ${{ steps.detect-variant.outputs.torch_version }}
```
sayakpaul: This is because we cannot always fixate on the torch version. Today the latest stable is 2.11 and that can change.

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,155 @@
...
+        run: nix build .#kernel-builder -L
+
+      - name: Init kernel project
+        run: |
+          cd /tmp
+          $GITHUB_WORKSPACE/result/bin/kernel-builder init \
+            --name kernels-test/e2e-test-kernel \
+            --backends cuda
+
+      - name: Validate scaffold
```
sayakpaul, line 51: Happy to remove this check but it's not adding much to the runtime.

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,155 @@
...
+          TORCH_MAJOR=${TORCH_DIGITS:0:1}
+          TORCH_MINOR=${TORCH_DIGITS:1}
+          TORCH_VERSION="${TORCH_MAJOR}.${TORCH_MINOR}.0"
+          echo "Detected Torch version: $TORCH_VERSION"
+          echo "torch_version=$TORCH_VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Upload kernel to Hub
+        run: |
+          $GITHUB_WORKSPACE/result/bin/kernel-builder upload /tmp/e2e-test-kernel \
+            --repo-id kernels-test/kernels-upload-test
```
sayakpaul: This is a separate org entirely so I guess we should safe regarding security?

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,149 @@
...
+            sandbox-fallback = false
+      - uses: cachix/cachix-action@v16
+        with:
+          name: huggingface
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
+        env:
+          USER: runner
+
+      - name: Build kernel-builder CLI
+        run: nix build .#kernel-builder -L
```
danieldk: Not needed, see below.

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Init kernel project
+        run: |
+          cd /tmp
+          $GITHUB_WORKSPACE/result/bin/kernel-builder init \
+            --name kernels-test/e2e-test-kernel \
+            --backends cuda
```
danieldk:
```suggestion
          nix run $GITHUB_WORKSPACE#kernel-builder -- init \
            --name kernels-test/e2e-test-kernel \
            --backends cuda
```

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Patch flake.nix to use local nix-builder
+        run: |
+          cd /tmp/e2e-test-kernel
+          sed -i 's|github:huggingface/kernels|path:'"$GITHUB_WORKSPACE"'|' flake.nix
+
+      - name: Build one CUDA variant
+        run: |
+          cd /tmp/e2e-test-kernel
+          nix build .#ci -L
```
danieldk: I think we want to use `kernel-builder build` here for the e2e test?

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,149 @@
...
+          ls "$VARIANT_DIR"/*.so
+
+          # Extract Torch version from variant dir name (e.g. torch210 -> 2.10.0)
+          VARIANT_NAME=$(basename "$VARIANT_DIR")
+          TORCH_DIGITS=$(echo "$VARIANT_NAME" | grep -oP '(?<=torch)\d+')
+          TORCH_MAJOR=${TORCH_DIGITS:0:1}
+          TORCH_MINOR=${TORCH_DIGITS:1}
+          TORCH_VERSION="${TORCH_MAJOR}.${TORCH_MINOR}.0"
+          echo "Detected Torch version: $TORCH_VERSION"
+          echo "torch_version=$TORCH_VERSION" >> "$GITHUB_OUTPUT"
```
danieldk: I think much of this (but not all) could be replaced by testing with `get_local_kernels` maybe?

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - name: Install uv and set Python version
+        uses: astral-sh/setup-uv@v7
+        with:
+          python-version: "3.12"
+
+      - name: Install Python deps
+        working-directory: ./kernels
+        run: |
+          uv sync --all-extras --dev
+          uv pip install "torch==${{ needs.init-build-upload.outputs.torch_version }}"
```
danieldk: I don't think we'd need this if we do a general build. A generic "uv pip install torch` would work since we'd build all variants. If `uv pip install torch` (what most users do) doesn't work, we want to see that in this test, because it means that we failed to support the latest Torch.

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,149 @@
...
+          TORCH_MAJOR=${TORCH_DIGITS:0:1}
+          TORCH_MINOR=${TORCH_DIGITS:1}
+          TORCH_VERSION="${TORCH_MAJOR}.${TORCH_MINOR}.0"
+          echo "Detected Torch version: $TORCH_VERSION"
+          echo "torch_version=$TORCH_VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Upload kernel to Hub
+        run: |
+          $GITHUB_WORKSPACE/result/bin/kernel-builder upload /tmp/e2e-test-kernel \
+            --repo-id ${{ env.E2E_REPO_ID }}
```
danieldk: This and the code below has a race condition. If two PRs run this same workflow, they'll trample on eachother's uploads.

We probably need to upload to a random branch (and remove that afterwards).

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,149 @@
...
+          result = kernel.e2e_test_kernel(x)
+          expected = x + 1.0
+          torch.testing.assert_close(result, expected)
+          print('E2E test passed: get_kernel + correctness check')
+          "
+
+      - name: Cleanup Hub repo
+        if: always()
+        working-directory: ./kernels
+        run: uv run huggingface-cli delete-repo ${{ env.E2E_REPO_ID }} --yes || true
```
danieldk: See above, worsens the race condition.

.github/workflows/test_e2e.yaml
sayakpaul: This is regarding detecting torch version. Did you mean to comment on a different LOC?

.github/workflows/test_e2e.yaml
sayakpaul: Great catch! Should be addressed now.

.github/workflows/test_e2e.yaml
sayakpaul: Since we're now uploading a revision, I don't think we have to delete the repo? LMK.

.github/workflows/test_e2e.yaml
sayakpaul: Indeed. Addressed.

.github/workflows/test_e2e.yaml
```diff
@@ -91,18 +83,9 @@ jobs:
...
-          TORCH_MAJOR=${TORCH_DIGITS:0:1}
-          TORCH_MINOR=${TORCH_DIGITS:1}
-          TORCH_VERSION="${TORCH_MAJOR}.${TORCH_MINOR}.0"
-          echo "Detected Torch version: $TORCH_VERSION"
-          echo "torch_version=$TORCH_VERSION" >> "$GITHUB_OUTPUT"
-
       - name: Upload kernel to Hub
         run: |
-          $GITHUB_WORKSPACE/result/bin/kernel-builder upload /tmp/e2e-test-kernel \
+          nix run $GITHUB_WORKSPACE#kernel-builder -- upload /tmp/e2e-test-kernel \
```
danieldk: I think we need `--repo-id kernels-test/kernels-upload-test` here?

danieldk: Never mind, it gets added through the template.

.github/workflows/test_e2e.yaml
```diff
@@ -0,0 +1,129 @@
...
+          name: huggingface
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
+        env:
+          USER: runner
+
+      - name: Init kernel project
+        run: |
+          cd /tmp
+          nix run $GITHUB_WORKSPACE#kernel-builder -- init \
+            --name kernels-test/e2e-test-kernel \
```
danieldk:
```suggestion
            --name ${{ env.E2E_REPO_ID }} \
```

PR #418, opened 2026-03-31
add an onboarding script for new kernel-builder users

### LINKED ISSUE #414: Create onboarding script with the new kernel-builder cli

sayakpaul:
Fix #414 

Screenshot 2026-03-31 at 1 29 22 PM


install.sh
```diff
@@ -0,0 +1,184 @@
...
+    x86_64-linux|aarch64-linux|aarch64-darwin) ;;
+    *)
+      error "Unsupported platform: $PLATFORM"
+      echo "  Supported platforms: x86_64-linux, aarch64-linux, aarch64-darwin"
+      exit 1
+      ;;
+  esac
+
+  info "Detected platform: $PLATFORM"
+}
```
danieldk: Doesn't the Determinate installer do platform checks?

install.sh
```diff
@@ -0,0 +1,184 @@
...
+
+install_nix() {
+  if find_nix; then
+    info "Nix is already installed: $(nix --version)"
+    return 0
+  fi
+
+  info "Installing Determinate Nix..."
+  curl -fsSL https://install.determinate.systems/nix | sh -s -- install --no-confirm
```
danieldk, line 60: On macOS we need to run with `--extra-conf "sandbox = relaxed"`, otherwise we cannot access the Metal compiler from the build sandbox.

danieldk: Not on other platforms though, where we really really want full sandboxing (I think it's enabled by default).

install.sh
```diff
@@ -0,0 +1,184 @@
...
+  fi
+
+  info "Configuring Hugging Face binary cache..."
+  # Use 'trusted-substituters' and 'trusted-public-keys' — these are daemon-level
+  # settings (written to nix.conf as root) so they don't require the user to be
+  # in 'trusted-users'.
+  echo "extra-trusted-substituters = $HF_SUBSTITUTER" | sudo tee -a /etc/nix/nix.conf >/dev/null
+  echo "extra-trusted-public-keys = $HF_PUBLIC_KEY" | sudo tee -a /etc/nix/nix.conf >/dev/null
+  sudo systemctl restart nix-daemon 2>/dev/null || sudo pkill -HUP nix-daemon || true
+  sleep 2
```
danieldk: Doesn't `systemctl restart` block by default until the unit is restarted?

install.sh
```diff
@@ -0,0 +1,184 @@
...
+  RESET=""
+fi
+
+info()  { echo -e "${BOLD}${GREEN}==>${RESET} ${BOLD}$1${RESET}"; }
+warn()  { echo -e "${BOLD}${YELLOW}warning:${RESET} $1"; }
+error() { echo -e "${BOLD}${RED}error:${RESET} $1" >&2; }
+
+# --- Platform detection ---
+
+detect_platform() {
```
drbh: seems like we can avoid this function since I don't think the `PLATFORM` is used elsewhere in the script.

I do see the use case for flagging that a system is not supported, however I think we should still let the user install nix and the cli tool, and simply throw in the builder if the platform isn't supported

install.sh
```diff
@@ -0,0 +1,184 @@
...
+    info "Hugging Face binary cache is already configured"
+    return 0
+  fi
+
+  info "Configuring Hugging Face binary cache..."
+  # Use 'trusted-substituters' and 'trusted-public-keys' — these are daemon-level
+  # settings (written to nix.conf as root) so they don't require the user to be
+  # in 'trusted-users'.
+  echo "extra-trusted-substituters = $HF_SUBSTITUTER" | sudo tee -a /etc/nix/nix.conf >/dev/null
+  echo "extra-trusted-public-keys = $HF_PUBLIC_KEY" | sudo tee -a /etc/nix/nix.conf >/dev/null
```
drbh: maybe the keys should be `trusted-substituters` and `trusted-public-keys` (no extra)?

currently I get the following warning

```
warning: ignoring untrusted substituter 'https://huggingface.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
warning: ignoring untrusted substituter 'https://huggingface.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
```

install.sh
```diff
@@ -0,0 +1,184 @@
...
+  nix profile add --accept-flake-config "${FLAKE_REF}#kernel-builder"
+
+  if ! command -v kernel-builder &>/dev/null; then
+    error "kernel-builder was installed but is not in PATH."
+    echo "  Try restarting your shell or running:"
+    echo "    . $NIX_PROFILE_SCRIPT"
+    exit 1
+  fi
+
+  info "kernel-builder installed: $(kernel-builder --version)"
```
drbh: not sure we need the first check, I think we can rely on the `$(kernel-builder --version)` call, or mainly I just think we can avoid the error message since this install script will run in its own bash shell if the user is installing via `curl ... | bash` so the restarting shell suggestion likely wouldn't fix if the `nix profile` command fails. 

tldr; not sure if we need the check/error

install.sh
```diff
@@ -0,0 +1,184 @@
...
+  echo "    1. Create a new kernel:     kernel-builder init my-kernel"
+  echo "    2. Build your kernel:       cd my-kernel && nix run .#build-and-copy -L"
+  echo "    3. Read the docs:           https://huggingface.co/docs/kernels/"
+  echo ""
+  echo "  To update kernel-builder later:"
+  echo "    nix profile upgrade --all"
+  echo ""
+  echo "  Note: you may need to restart your shell or run:"
+  echo "    . $NIX_PROFILE_SCRIPT"
+  echo ""
```
drbh, line 147: I think this will likely be required since we cant update the users shell from within this bash script install (i think...)

when I tested locally I needed to run `. /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` since the cli was not found.

another possible idea is that we add a symlink to the profile path into a location that is likely loaded into the users env already. maybe something like

```bash
sudo ln -sf "$HOME/.nix-profile/bin/kernel-builder" /usr/local/bin/kernel-builder
```

although I'm not sure if this a bad practice when it comes to nix profiles. just a thought

drbh (changes requested):

install.sh
sayakpaul: I am still getting

```
warning: ignoring untrusted substituter 'https://huggingface.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
```

after trying out a bunch of options. I am afraid my knowledge about nix is too limited at this point to keep on digging. Any suggestions?

install.sh
sayakpaul: The script already sources the Nix profile via `find_nix / . "$NIX_PROFILE_SCRIPT"` within its own process, so `kernel-builder --version` works during the script.

I would prefer this to symlinking, though.

install.sh
```diff
@@ -0,0 +1,151 @@
...
+  # 'extra-trusted-users' appends to the existing trusted-users list.
+  # Using 'trusted-users' would override the default (root), which could
+  # break the Nix installation.
+  sudo tee -a /etc/nix/nix.conf >/dev/null <<EOF
+extra-trusted-users = $user
+extra-substituters = $HF_SUBSTITUTER
+extra-trusted-public-keys = $HF_PUBLIC_KEY
+EOF
+  sudo systemctl restart nix-daemon 2>/dev/null || sudo pkill -HUP nix-daemon || true
+  sleep 3
```
sayakpaul, line 98: The issue is that `systemctl restart` might return before the Nix daemon is actually ready to serve requests with the new config — despite blocking on the unit state, the daemon's socket may not be ready yet.

This is needed in order to propagate the substitutes above to Nix config.

install.sh
drbh: yea sorry I wasn't clear, I meant it is likely required for the user to run, if we don't have some other method to make it available automatically. might be worth exploring the symlink or some other approach. 

However if this is too tedious, I also think its okay if we require the user to run `. /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` for now

install.sh
drbh: ooh I think we may need to add the user as a `trusted-users = <username>` otherwise the those values are ignored

drbh (approved):
lgtm, working as expected with 

```
curl -Ls https://raw.githubusercontent.com/huggingface/kernels/6c4a107ba2c488ce50b9a19994e0a1e008fd2e08/install.sh | bash
```

install.sh
danieldk: Maybe there is a principled way of waiting until the socket is ready?

sayakpaul: For an onboarding script, I think this is fine to relax a bit of complexity.

PR #419, opened 2026-03-31
Support fast kernel build variant enumeration and expose variants

danieldk:
The goal of this PR is two-fold:

1. We want to expose the build variant names of a kernel in an easy way for the Rust CLI.
2. We want to speed up the evaluation of the build variant names, both for the Rust CLI and for shell completion of Nix commands.

The problem with the existing implementation was that the build variant names were generated by probing the Torch/tvm-ffi/CUDA/ROCm/XPU versions on the actual package set of a build set. Even though Nix uses lazy evaluation, this still required evaluating enough of the package set associated with each build set to extract this information, which could have a multi-second latency.

Note that we can generate the variant with the build configurations, since they contain the framework and backend versions, plus the kernel configuration to filter the applicable build sets. This change follows this observation to generate the variant names without evaluating the package sets.

With much cheaper evaluation, we can also expose the variants as a list for the Rust CLI.


nix-builder/versions.nix
```diff
@@ -90,7 +90,7 @@
       "aarch64-linux"
     ];
     bundleBuild = true;
-    tvmFfi = true;
+    tvmFfiVersion = "0.1";
```
sayakpaul, line 93: Hardcoding it seems weird.

danieldk: We are doing the same for Torch versions, we'll need it in the future for supporting two tvm-ffi features (e.g. 0.1.x and 0.2.x) similar to our two release PyTorch policy.

PR #420, opened 2026-03-31
kernel-builder: add variants support

danieldk:
`list-variants` now lists all supported build variants of a kernel. The `devshell`, `testshell`, and `build` commands now accept the `--variant` option to select a different variant than the default.

Also fix the `build` subcommand, which should call `nix build` and not `nix run`.

### DIFF
--- a/kernel-builder/src/list_variants.rs
+++ b/kernel-builder/src/list_variants.rs
@@ -8,7 +8,7 @@ use crate::util::check_or_infer_kernel_dir;
 /// Get the list of variants for a flake.
 pub fn variants(flake: &Flake) -> Result<Vec<String>> {
     let nix = Nix::new().json(true);
-    let output = nix.output(NixSubcommand::Eval {
+    let output = nix.run_capture_stdout(NixSubcommand::Eval {
         flake,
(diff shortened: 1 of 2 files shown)


kernel-builder/src/develop.rs
```diff
@@ -4,20 +4,44 @@ use std::path::PathBuf;
 
 use eyre::Result;
 
+use crate::list_variants::arch_variants;
 use crate::nix::{Flake, Nix, NixSubcommand};
 use crate::util::check_or_infer_kernel_dir;
 
+/// Validate a variant against the arch variants list.
+fn validate_arch_variant(flake: &Flake, variant: &str) -> Result<()> {
```
sayakpaul, line 12: Does the backend need to be considered as well?

kernel-builder/src/main.rs
```diff
@@ -69,6 +72,10 @@ enum Commands {
         #[arg(value_name = "KERNEL_DIR")]
         kernel_dir: Option<PathBuf>,
 
+        /// Build a specific variant.
```
sayakpaul, line 75: Super nice!

PR #421, opened 2026-03-31
feat: add rust kernels library for loading kernels

drbh:
This PR adds a new client library for loading hf kernels in rust. This allow tvmffi based kernels to be called from rust and optionally integrates with candle for a better tensor ux.

## Example usage with candle

```rust
use candle_core::{Device, Tensor};
use kernels::Result;
use kernels::candle::CallKernel;

fn main() -> Result<()> {
    let activation = kernels::candle::get_kernel("drbh/relu-tvm", 1)?;
    let device = activation.device()?;
    println!("Backend: {}", activation.backend());

    let x = Tensor::new(&[-1.0f32, 2.0, -3.0, 4.0, -0.5, 0.0, 1.5, -2.5], &device)?;
    let y = Tensor::zeros_like(&x)?;
    activation.call("relu", &[&y, &x])?;

    let result = y.to_vec1::<f32>()?;
    let expected = Tensor::new(&*x.to_vec1::<f32>()?, &Device::Cpu)?
        .relu()?
        .to_vec1::<f32>()?;

    println!("Input:    {:?}", x.to_vec1::<f32>()?);
    println!("TVM FFI:  {result:?}");
    println!("Candle:   {expected:?}");
    assert_eq!(result, expected);
    println!("OK");
    Ok(())
}
```

repo with candle and non candle examples https://github.com/drbh/hf-kernels-rust


kernels-rs/src/backend.rs
```diff
@@ -0,0 +1,101 @@
...
+    let rest = stdout.split("CUDA Version:").nth(1)?;
+    Some(rest.split_whitespace().next()?.to_string())
+}
+
+fn cuda_version_from_nvcc() -> Option<String> {
+    let output = Command::new("nvcc").arg("--version").output().ok()?;
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let after = stdout.split("release ").nth(1)?;
+    Some(after.split(',').next()?.trim().to_string())
+}
```
danieldk, line 98: This may not be the same as the library that a framework is compiled against and dynamically loads. Also, `nvidia-smi` gives the driver library version, not the CUDA runtime version. We need to get it from cudart, e.g. see:

https://github.com/huggingface/kernels/blob/8ed7bb48817507c00f8c7d9c0ac8857d2e1e5a6f/kernels/src/kernels/backends.py#L254

libloading seems to be the most widely used library for dlopen:

https://github.com/nagisa/rust_libloading/

kernels-rs/src/candle.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub fn candle_device(self) -> Result<Device> {
+        match self {
+            BackendKind::Cpu => Ok(Device::Cpu),
+            #[cfg(feature = "candle-cuda")]
+            BackendKind::Cuda => Device::new_cuda(0).map_err(Into::into),
+            #[cfg(not(feature = "candle-cuda"))]
+            BackendKind::Cuda => Ok(Device::Cpu),
+            BackendKind::Xpu => Ok(Device::Cpu),
+        }
+    }
```
danieldk: I think this can be `TryFrom<BackendDevice> for Device`. Not 100% sure if it works with the coherency rules, since it's a different mod in the same crate. But I think it should.

kernels-rs/src/candle.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    pub fn candle_supported(self) -> Self {
+        match self {
+            #[cfg(feature = "candle-cuda")]
+            BackendKind::Cuda => BackendKind::Cuda,
+            #[cfg(not(feature = "candle-cuda"))]
+            BackendKind::Cuda => BackendKind::Cpu,
+            other => other,
+        }
+    }
```
danieldk: The function name is not very descriptive, maybe `to_candle_supported`?

kernels-rs/src/candle.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+impl From<&Device> for BackendKind {
+    fn from(device: &Device) -> Self {
+        match device {
+            Device::Cpu => BackendKind::Cpu,
+            #[cfg(feature = "candle-cuda")]
+            Device::Cuda(_) => BackendKind::Cuda,
+            #[allow(unreachable_patterns)]
+            _ => BackendKind::Cpu,
```
danieldk: I think it would be better to explicitly enumerate the other variants here, so that we can rely on exhaustiveness checking when other variants get added?

Also it seems that as it is, if Candle returns a device type that we don't support, it would result in `Cpu`, which results in kernels that are not compatible with the device type?

kernels-rs/src/candle.rs
```diff
@@ -0,0 +1,224 @@
...
+        lanes: 1,
+    })
+}
+
+fn cpu_storage_data_ptr(cpu: &CpuStorage, offset: usize) -> Result<*mut c_void> {
+    macro_rules! ptr {
+        ($v:expr) => {
+            Ok(unsafe { $v.as_ptr().add(offset) as *mut c_void })
+        };
+    }
```
danieldk: Remove, make explicit.

kernels-rs/src/candle.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    // SyncOnDrop records a stream event; the pointer stays valid as long
+    // as the caller holds the storage read-guard.
+    macro_rules! ptr {
+        ($slice:expr) => {{
+            let view = $slice.slice(offset..);
+            let (device_ptr, _sync) = view.device_ptr(&stream);
+            Ok(device_ptr as *mut c_void)
+        }};
+    }
```
danieldk: I think rather than a macro, this could be a trait + impl? At least I think with a generic type it should work with one implementation for all cases?

kernels-rs/src/candle.rs
```diff
@@ -0,0 +1,224 @@
...
+impl KernelModule {
+    pub fn device(&self) -> Result<Device> {
+        self.backend().kind().candle_device()
+    }
+}
+
+// Tensors are passed to the kernel as DLPack pointers directly into
+// candle's storage - no copies for contiguous tensors.
+pub trait CallKernel {
+    fn call(&self, func_name: &str, args: &[&Tensor]) -> Result<()>;
```
danieldk: What if there are non-tensor argument, e.g. option bools, epsilon floats, etc.?

kernels-rs/src/candle.rs
```diff
@@ -0,0 +1,224 @@
...
+                    dtype: dtype_to_dl(contiguous[i].dtype())?,
+                })
+            })
+            .collect::<Result<_>>()?;
+
+        let device_type = match kind {
+            BackendKind::Cpu => tvm_ffi::DL_CPU,
+            BackendKind::Cuda => tvm_ffi::DL_CUDA,
+            BackendKind::Xpu => tvm_ffi::DL_ONEAPI,
+        };
```
danieldk: Seems like this could use a `From` implementation outside the function?

drbh: good point, updated in latest

kernels-rs/src/candle.rs
drbh: good catch I originally only tested with a kernel that expected tensors. updated to handle multiple types in the latest changes

kernels-rs/src/candle.rs
drbh: updated to remove the macros (comment above) and explored a trait but ended up settling on a generic function like

```rust
fn cuda_slice_data_ptr<T>(
    slice: &cudarc::driver::CudaSlice<T>,
    stream: &cudarc::driver::CudaStream,
    offset: usize,
) -> Result<*mut c_void> {
```

this helped make the cpu and cuda function follow a similar functional pattern

note: the cpu path uses a generic function like

```rust
fn cpu_slice_data_ptr<T>(slice: &[T], offset: usize) -> Result<*mut c_void> {
```

happy to explore another approach if you see any issues with this! thanks!

kernels-rs/src/candle.rs
drbh: removed and opt'ed to add `*_slice_data_ptr` functions for both cpu and cuda to avoid the macros in both places.

kernels-rs/src/candle.rs
drbh: sound good to me, updated in latest

kernels-rs/src/candle.rs
drbh: this is much nicer, thanks for the suggestion! updated in latest

kernels-rs/src/backend.rs
drbh: good catch, I've updated to prefer querying via `cudaRuntimeGetVersion` from `cudart` in the latest changes. I've tested locally and am not running into any issues - however I'm not 100% sure if we need more logic to search for the cudart like `ctypes.util.find_library("cudart")` does if its not in the default location

kernels-rs/src/candle.rs
drbh: agreed thats a much better approach. I've updated to enumerate the `Device`s in latest and throw errors if the device is not supported by the currently tvmffi impl

PR #422, opened 2026-04-01
feat: better reporting of the variant being requested.

sayakpaul:
https://huggingface.slack.com/archives/C090JN2P8NB/p1774964954954269

### DIFF
--- a/kernels/src/kernels/cli/__init__.py
+++ b/kernels/src/kernels/cli/__init__.py
@@ -65,7 +65,8 @@ def main():
     versions_parser.set_defaults(func=kernel_versions)
 
     upload_parser = subparsers.add_parser(
-        "upload", help="(Deprecated) Upload kernels to the Hub. Use `kernel-builder upload` instead."
+        "upload",
+        help="(Deprecated) Upload kernels to the Hub. Use `kernel-builder upload` instead.",
     )
     upload_parser.add_argument(
         "kernel_dir",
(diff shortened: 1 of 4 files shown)


kernels/src/kernels/cli/__init__.py
```diff
@@ -65,7 +65,8 @@ def main():
     versions_parser.set_defaults(func=kernel_versions)
 
     upload_parser = subparsers.add_parser(
-        "upload", help="(Deprecated) Upload kernels to the Hub. Use `kernel-builder upload` instead."
+        "upload",
+        help="(Deprecated) Upload kernels to the Hub. Use `kernel-builder upload` instead.",
```
sayakpaul, line 69: Result of `make style`

kernels/src/kernels/variants.py
```diff
@@ -232,6 +229,59 @@ def resolve_variants(
...
+    os: str,
+    torch_version: Version | None,
+    torch_cxx11_abi: bool | None,
+    tvm_ffi_version: Version | None,
+) -> str:
+    """Return a human-readable description of the requested variant given explicit system parameters."""
+    parts = []
+
+    if torch_version is not None:
+        parts.append(f"torch{torch_version.major}{torch_version.minor}")
```
danieldk, line 262: I don't think we should go back to manual variant creation, it is flaky and breaks things.

(E.g. here it breaks the deprecation of the `cxx` tag.)

sayakpaul:
@danieldk I guess this PR can be closed since you wanted to tackle it through description of why a variant couldn't be loaded?

PR #423, opened 2026-04-01
kernels: add a function `system_variants`

danieldk:
This function lists all the variants that are available on the current system.

### DIFF
--- a/kernels/tests/test_variants.py
+++ b/kernels/tests/test_variants.py
@@ -11,67 +11,43 @@ from kernels.variants import (
     system_variants,
 )
 
-VARIANT_STRINGS = [
-    "torch25-cxx98-cu118-aarch64-linux",
-    "torch25-cxx98-cpu-x86_64-linux",
-    "torch29-cpu-aarch64-darwin",
-    "torch29-cxx11-cpu-aarch64-linux",
-    "torch29-cxx11-cpu-x86_64-linux",
-    "torch29-cxx11-cu126-aarch64-linux",
-    "torch29-cxx11-cu126-x86_64-linux",
-    "torch29-cxx11-cu128-aarch64-linux",
-    "torch29-cxx11-cu128-x86_64-linux",
-    "torch29-cxx11-cu130-aarch64-linux",
-    "torch29-cxx11-cu130-x86_64-linux",
-    "torch29-cxx11-rocm63-x86_64-linux",
-    "torch29-cxx11-rocm64-x86_64-linux",
-    "torch29-cxx11-xpu20252-x86_64-linux",
-    "torch29-cpu-aarch64-linux",
-    "torch29-cpu-x86_64-linux",
-    "torch29-cu126-aarch64-linux",
-    "torch29-cu126-x86_64-linux",
-    "torch29-cu128-aarch64-linux",
-    "torch29-cu128-x86_64-linux",
-    "torch29-cu130-aarch64-linux",
-    "torch29-cu130-x86_64-linux",
-    "torch29-rocm63-x86_64-linux",
-    "torch29-rocm64-x86_64-linux",
-    "torch29-xpu20252-x86_64-linux",
-    "torch29-metal-aarch64-darwin",
-    "torch210-cpu-aarch64-darwin",
-    "torch210-cu128-x86_64-windows",
-    "torch210-cxx11-cpu-aarch64-linux",
-    "torch210-cxx11-cpu-x86_64-linux",
-    "torch210-cxx11-cu126-aarch64-linux",
-    "torch210-cxx11-cu126-x86_64-linux",
-    "torch210-cxx11-cu128-aarch64-linux",
-    "torch210-cxx11-cu128-x86_64-linux",
-    "torch210-cxx11-cu130-aarch64-linux",
-    "torch210-cxx11-cu130-x86_64-linux",
-    "torch210-cxx11-rocm70-x86_64-linux",
-    "torch210-cxx11-rocm71-x86_64-linux",
-    "torch210-cxx11-xpu20253-x86_64-linux",
-    "torch210-cpu-aarch64-linux",
-    "torch210-cpu-x86_64-linux",
-    "torch210-cu126-aarch64-linux",
-    "torch210-cu126-x86_64-linux",
-    "torch210-cu128-aarch64-linux",
-    "torch210-cu128-x86_64-linux",
-    "torch210-cu130-aarch64-linux",
-    "torch210-cu130-x86_64-linux",
-    "torch210-rocm70-x86_64-linux",
-    "torch210-rocm71-x86_64-linux",
-    "torch210-xpu20253-x86_64-linux",
-    "torch210-metal-aarch64-darwin",
-    "torch210-xpu20253-x86_64-windows",
-    "tvm-ffi01-cpu-x86_64-linux",
-    "tvm-ffi01-cu126-x86_64-linux",
-    "tvm-ffi01-cu128-x86_64-linux",
-    "tvm-ffi01-cu130-x86_64-linux",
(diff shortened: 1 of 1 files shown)


kernels/src/kernels/variants.py
```diff
@@ -28,32 +30,77 @@
 
 @dataclass(unsafe_hash=True)
 class Torch:
-    _VARIANT_REGEX: ClassVar[re.Pattern] = re.compile(r"torch(\d+?)(\d+)")
+    """Versioned Torch framework (arch variants)."""
 
-    version: Version | None
+    _VARIANT_REGEX: ClassVar[re.Pattern] = re.compile(
+        r"torch(\d+?)(\d+)(?:-(cxx11|cxx98))?"
```
danieldk: Prepare for variants without an ABI tag (see extra variants in tests).

sayakpaul: Here I think one liner explainer comment would be nice. Bits around `(cxx11|cxx98)` aren't particularly clear I think.

kernels/src/kernels/variants.py
```diff
@@ -28,32 +30,77 @@
...
+
+            torch_version = parse(torch.__version__)
+            torch_version = Version(f"{torch_version.major}.{torch_version.minor}")
+
+            os_ = platform.system().lower()
+            if os_ == "linux":
+                cxx11_abi = torch.compiled_with_cxx11_abi()
+                return [
+                    Torch(version=torch_version, cxx11_abi=cxx11_abi),
+                    Torch(version=torch_version, cxx11_abi=None),
```
sayakpaul, line 65: Curious why include `Torch(version=torch_version, cxx11_abi=None)` here as well?

kernels/src/kernels/variants.py
```diff
@@ -121,37 +194,92 @@ def parse(s: str) -> "Noarch":
...
+    @staticmethod
+    def possible_variants() -> list["ArchVariant"]:
+        frameworks: list[Torch | TvmFfi] = (
+            Torch.possible_variants() + TvmFfi.possible_variants()
+        )
+        archs = Arch.possible_variants()
+        return [
+            ArchVariant(framework=fw, arch=arch)
+            for fw, arch in itertools.product(frameworks, archs)
+        ]
```
sayakpaul, line 222: This is so neat! As declarative as it gets.

kernels/src/kernels/variants.py
```diff
@@ -121,37 +194,92 @@ def parse(s: str) -> "Noarch":
...
+        else:
+            framework_str = parts[0]
+            arch_parts = parts[1:]
+        return ArchVariant(
+            framework=Torch.parse(framework_str), arch=Arch.parse(arch_parts)
+        )
+    elif parts[0] == "tvm" and len(parts) >= 2 and parts[1].startswith("ffi"):
+        return ArchVariant(
+            framework=TvmFfi.parse(f"tvm-{parts[1]}"), arch=Arch.parse(parts[2:])
+        )
```
sayakpaul, line 290: This conditional logic feels a bit rigid, but I guess it's purely heuristics-driven and isn't meant to change? If it changes then there's something wrong.

kernels/src/kernels/variants.py
```diff
@@ -121,37 +194,92 @@ def parse(s: str) -> "Noarch":
...
-            framework = Torch.parse(parts[0])
-            arch = Noarch.parse("-".join(parts[1:]))
-        elif parts[0].startswith("torch"):
-            framework = Torch.parse(parts[0])
-            arch = Arch.parse(parts[1:])
-        elif parts[0] == "tvm" and parts[1].startswith("ffi"):
-            framework = TvmFfi.parse(f"tvm-{parts[1]}")
-            arch = Arch.parse(parts[2:])
-        else:
-            raise ValueError(f"Unknown framework in variant string: {variant_str!r}")
```
sayakpaul, line 152: So, we're removing `parse()` from the variant dataclasses and keeping a central `parse_variant()` utility?

kernels/src/kernels/variants.py
```diff
@@ -121,37 +194,92 @@ def parse(s: str) -> "Noarch":
...
+        ArchVariant.possible_variants() + NoarchVariant.possible_variants()
+    )
+    return _sort_variants(result)
+
+
+def parse_variant(variant_str: str) -> Variant:
+    """Parse a variant string into an ArchVariant or NoarchVariant."""
+    parts = variant_str.split("-")
+
+    if parts[0] == "torch":
```
sayakpaul, line 272: Should there be a constraint on the length of `parts` in case of `NoarchVariant` because `ArchVariant` will also have `parts[0] == "torch"`?

kernels/tests/test_variants.py
```diff
@@ -3,7 +3,13 @@
...
-from kernels.variants import Variant, _resolve_variant_for_system, get_variants
+from kernels.variants import (
+    _resolve_variant_for_system,
+    get_variants,
+    parse_variant,
+    resolve_variants,
+    system_variants,
+)
 
 VARIANT_STRINGS = [
```
sayakpaul: It could be nice to programatically construct this list against a predefined set of `framework` (+ versions), `platform`, `arch`, `backend`, etc.

kernels/src/kernels/variants.py
danieldk: Torch used to have C++98 and C++11 ABI builds, but has now fully switched to the C++11 ABI.  So the ABI tag is now only there for historical reasons. They will be removed from Linux kernel builds in the future, but in order to do so, we have to support it in `kernels` for a while, so that once we flip that switch, `kernels` can deal with build variants without an ABI tag.

danieldk: I am adding a comment to the code to clarify this.

kernels/src/kernels/variants.py
danieldk: Yeah, I'm not a super big fan either. The alternative would be failure-driven parsing, where we try to parse as noarch first, if that fails as torch arch and if that fails as tvm-ffi. But that has other downsides I think.

kernels/src/kernels/variants.py
danieldk: Yeah, since there are now two variant classes (`ArchVariant` and `NoArchVariant`) and `Variant` changed to being a union.

If we had Rust enums, we could do this in a nice way 😁.

danieldk: I'm going to try to make it like a Rust enum.

danieldk: Made everything messier, so I'll leave it as-is for now.

kernels/src/kernels/variants.py
danieldk: But in the arch variant, parts[0] only starts with `torch`, since it also has the version number.

kernels/tests/test_variants.py
danieldk: Done.

kernels/src/kernels/variants.py
danieldk: Added a comment.

kernels/src/kernels/variants.py
sayakpaul: Yeah aligned.

kernels/src/kernels/variants.py
sayakpaul: Ouch. Thanks for trying out!

kernels/tests/test_variants.py
```diff
@@ -3,47 +3,51 @@
...
-
+from kernels.variants import (
+    _resolve_variant_for_system,
+    get_variants,
+    parse_variant,
+    resolve_variants,
+    system_variants,
+)
+
+VARIANT_STRINGS = (
```
sayakpaul, line 14: Ouf so neat!

PR #425, opened 2026-04-02
Update nixpkgs and other flake dependencies

danieldk:


nix-builder/pkgs/python-modules/torch/binary/generic.nix
```diff
@@ -125,13 +125,11 @@ let
   };
 
 in
-buildPythonPackage {
+buildPythonPackage.override { stdenv = effectiveStdenv; } {
```
danieldk, line 128: Using `stdenv` in the attrset passed to `buildPythonPackage` is now deprecated in nixpkgs.

nix-builder/pkgs/python-modules/hooks/python-wheel-deps-check-hook.sh
```diff
@@ -15,7 +15,7 @@ pythonWheelDepsCheckHook() {
     echo "Finished executing pythonWheelDepsCheck"
 }
 
-if [ -z "${dontCheckRuntimeDeps-}" ]; then
+if [ -z "${dontCheckWheelDeps-}" ]; then
```
danieldk, line 18: Forgot to rename this when adding this hook to our repo, but the name now clashes with the same variable in the original hook in nixpkgs.

nix-builder/pkgs/rocm-packages/default.nix
```diff
@@ -9,8 +9,12 @@
...
 
 let
-  fixedPoint = final: { inherit callPackage lib packageMetadata; };
-  composed = lib.composeManyExtensions [
+  inherit (lib.fixedPoints) extends composeManyExtensions;
+
+  fixedPoint = final: {
+    inherit lib packageMetadata;
+  };
+  composed = composeManyExtensions [
```
danieldk, line 17: https://huggingface.slack.com/archives/C090JN2P8NB/p1775027030268379

PR #428, opened 2026-04-07
feat: `get_loaded_kernels()`

cbensimon:
New `get_loaded_kernels()` API

Enables:
- `get_kernel` (internal) caching based on repo infos (done in this PR)
- kernels packaging (PyTorch ahead-of-time compilation context)

Example "kernels-included package" -> https://huggingface.co/cbensimon/FLUX.2-klein-4B-sm90-cu128-glibc235-r52/tree/main/package

### DIFF
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -119,8 +119,9 @@ def _import_from_path(module_name: str, variant_path: Path, _repo_infos: RepoInf
         raise ImportError(f"Cannot load module {module_name} from spec")
     sys.modules[module_name] = module
     spec.loader.exec_module(module)  # type: ignore
-    _loaded_kernels[module.op._namespace] = LoadedKernel(
-        op_namespace=module.op._namespace,
+    op_namespace = sys.modules[f"{module_name}._ops"].ops.name
+    _loaded_kernels[op_namespace] = LoadedKernel(
+        op_namespace=op_namespace,
         variant_path=variant_path,
         package_name=package_name,
         module_name=module_name,
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -36,7 +36,7 @@ KNOWN_BACKENDS = {"cpu", "cuda", "metal", "neuron", "rocm", "xpu", "npu"}
 
 class RepoInfos(NamedTuple):
     repo_id: str
-    revision: str | None
+    revision: str
     backend: str | None
 
 
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -12,6 +12,7 @@ import sys
 from importlib.metadata import Distribution
 from pathlib import Path
 from types import ModuleType
+from typing import NamedTuple
 
 from huggingface_hub import HfApi, constants
 
@@ -31,6 +32,30 @@ from kernels.variants import (
 )
 
 KNOWN_BACKENDS = {"cpu", "cuda", "metal", "neuron", "rocm", "xpu", "npu"}
+
+
+class RepoInfos(NamedTuple):
+    repo_id: str
+    revision: str | None
+    version: int | str | None
+    backend: str | None
+
+
+class LoadedKernel(NamedTuple):
+    op_namespace: str
+    variant_path: Path
+    package_name: str
+    module_name: str
+    repo_infos: RepoInfos | None
+
+
+_loaded_kernels: dict[str, LoadedKernel] = {}
+
+
+def get_loaded_kernels() -> dict[str, LoadedKernel]:
+    """Returns the `op_namespace -> LoadedKernel` mapping
+    """
+    return _loaded_kernels.copy()
 
 
 def _get_cache_dir() -> str | None:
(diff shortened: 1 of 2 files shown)


kernels/src/kernels/utils.py
```diff
@@ -33,6 +34,29 @@
...
+    package_name: str
+    module_name: str
+    torch_namespace: str | None
+    repo_infos: RepoInfos | None
+
+
+_loaded_kernels: dict[str, LoadedKernel] = {}
+
+
+def get_loaded_kernels() -> dict[str, LoadedKernel]:
```
danieldk: I know this is still a TODO, but I am curious what the use case of this function is, since it maps internal module names. I don't think we want to expose this to the user?

danieldk: If the loaded kernels need to be externally probed, maybe we can give the kernel a more predictable name in `sys.modules` (e.g. a kernels-specific prefix)? In that way the module table could be probed and it would be more resistant against e.g. someone unloading modules.

cbensimon: > I am curious what the use case of this function is, since it maps internal module names. I don't think we want to expose this to the user?

We actually don't need module names for the use-case I am envisioning (kernels packaging in the context of PyTorch ahead-of-time compilation) so we could totally simply return `list(_loaded_kernels.values())` instead of the mapping. Keying on module_name for the internal registry still feels quite natural to me but let me know what you think

(Internal Slack thread where we are also discussing this for reference: https://huggingface.slack.com/archives/C090JN2P8NB/p1775752974980369)

kernels/src/kernels/utils.py
```diff
@@ -279,10 +320,20 @@ def get_kernel(
         return get_local_kernel(override, package_name_from_repo_id(repo_id))
 
     revision = select_revision_or_version(repo_id, revision=revision, version=version)
+    repo_infos = RepoInfos(
+        repo_id=repo_id,
+        revision=revision,
+        backend=backend,
+    )
+    if not reload:
+        for loaded_kernel in get_loaded_kernels():
```
danieldk: Doing the lookup here makes the cached loading only work in `get_kernel` and not all other paths that import it. Maybe it makes more sense to do this in `_import_from_path` for now? We could always add a second cache keyed on the repository information if we'd like to skip network access as well in a future PR.

kernels/src/kernels/utils.py
```diff
@@ -239,6 +276,7 @@ def get_kernel(
     version: int | str | None = None,
     backend: str | None = None,
     user_agent: str | dict | None = None,
+    reload: bool = False,
```
danieldk: I don't think we should add this argument. We try to do the right/most sensible thing. I think a reload arg should only be added when using cached loads turns out to be a problem.

The smaller our API surface, the easier it is to maintain.

kernels/src/kernels/utils.py
```diff
@@ -93,6 +119,17 @@ def _import_from_path(module_name: str, variant_path: Path) -> ModuleType:
         raise ImportError(f"Cannot load module {module_name} from spec")
     sys.modules[module_name] = module
     spec.loader.exec_module(module)  # type: ignore
+    op_namespace: str | None = None
+    for so_path in file_path.parent.iterdir():
+        if so_path.is_file() and so_path.name.endswith('.so'):
+            op_namespace = so_path.name.split('.')[0]
```
danieldk: This is fragile. E.g. it doesn't work for no-arch kernels and IIRC we also have a different dylib extension name on macOS/Windows.

If we want a kernel to reveal its ops name, I think we should have a standardized API for it in kernels themselves. I think externally probing these things without standardization will break with various kernels (like it does here with noarch). The interface of kernels themselves is how downstream code should interact with kernels. `kernels` is only a client to fetch + load kernels, not to expose/probe kernel internals (if we can avoid it).

kernels/src/kernels/utils.py
```diff
@@ -93,6 +119,17 @@ def _import_from_path(module_name: str, variant_path: Path) -> ModuleType:
...
+    _loaded_kernels[module_name] = LoadedKernel(
+        variant_path=variant_path,
```
danieldk: This is easy to get from the kernel, so I don't think it should in `LoadedKernel`: `Path(k.__file__).parent`

kernels/src/kernels/utils.py
```diff
@@ -93,6 +119,17 @@ def _import_from_path(module_name: str, variant_path: Path) -> ModuleType:
...
+        package_name=package_name,
+        module_name=module_name,
+        op_namespace=op_namespace,
```
danieldk: As suggested above, we should probably define some kernel interface for this one.

kernels/src/kernels/utils.py
cbensimon: Ok as you prefer. When profiling `get_kernel` I noticed 95% of the time is spent in `install_kernel` that's why I thought it was the thing we wanted to cache. I only tried with flash-attn-3 though. Do some kernels take a long time to (re-)load (the actual loading part in `_import_from_path` when called a second time)?

kernels/src/kernels/utils.py
cbensimon: Removing it

kernels/src/kernels/utils.py
cbensimon: Ok I see. How would you envision such a standardized API for the kernels to expose their op namespace? (metadata.json could be a good candidate given that it's auto-generated, which I'm not sure about)

kernels/src/kernels/utils.py
danieldk: I haven't done any benchmarking, so I'm not sure.

danieldk (approved):
Looks great!

I think we could add this new function to the docs in a follow-up PR, but we could do that after we possibly add the metadata as well.

PR #435, opened 2026-04-09
feat: resolve repo type and fetch accordingly

drbh:
This PR adds support for fetching from kernel repo types and falls back to fetching from model repo types if the kernel repo does not exist

### DIFF
--- a/kernels/src/kernels/_versions.py
+++ b/kernels/src/kernels/_versions.py
@@ -107,15 +107,13 @@ def select_revision_or_version(
     *,
     revision: str | None,
     version: int | str | None,
-) -> tuple[str, str]:
+) -> tuple[str, str | None]:
     """Select a revision, returning (revision, repo_type)."""
     if revision is not None and version is not None:
         raise ValueError("Only one of `revision` or `version` must be specified.")
 
-    from kernels.utils import _resolve_repo_type
-
     if revision is not None:
-        return revision, _resolve_repo_type(repo_id)
+        return revision, None
     elif version is not None:
         ref, repo_type = resolve_version_spec_as_ref(repo_id, version)
         return ref.target_commit, repo_type
@@ -127,4 +125,4 @@ def select_revision_or_version(
         stacklevel=2,
     )
 
-    return "main", _resolve_repo_type(repo_id)
+    return "main", None
--- a/kernels/src/kernels/_versions.py
+++ b/kernels/src/kernels/_versions.py
@@ -8,20 +8,11 @@ from packaging.version import InvalidVersion, Version
 logger = logging.getLogger(__name__)
 
 
-def _get_available_versions(
-    repo_id: str,
-) -> tuple[dict[int, GitRefInfo], str]:
-    """Get kernel versions that are available in the repository.
+def _get_available_versions(repo_id: str) -> dict[int, GitRefInfo]:
+    """Get kernel versions that are available in the repository."""
+    from kernels.utils import _get_hf_api
 
-    Tries ``"kernel"`` repo type first. If the repository is not found,
-    falls back to ``"model"``.
-
-    Returns a tuple of (versions, repo_type)."""
-    from kernels.utils import _get_hf_api, _resolve_repo_type
-
-    repo_type = _resolve_repo_type(repo_id)
-
-    refs = _get_hf_api().list_repo_refs(repo_id=repo_id, repo_type=repo_type)
+    refs = _get_hf_api().list_repo_refs(repo_id=repo_id, repo_type="kernel")
 
     versions = {}
     for branch in refs.branches:
@@ -32,7 +23,7 @@ def _get_available_versions(
         except ValueError:
             continue
 
-    return versions, repo_type
+    return versions
 
 
 def _get_available_versions_old(repo_id: str) -> dict[Version, GitRefInfo]:
@@ -41,11 +32,10 @@ def _get_available_versions_old(repo_id: str) -> dict[Version, GitRefInfo]:
 
     This is for the old tag-based versioning scheme.
     """
-    from kernels.utils import _get_hf_api, _resolve_repo_type
+    from kernels.utils import _get_hf_api
 
-    repo_type = _resolve_repo_type(repo_id)
     versions = {}
-    for tag in _get_hf_api().list_repo_refs(repo_id, repo_type=repo_type).tags:
+    for tag in _get_hf_api().list_repo_refs(repo_id, repo_type="kernel").tags:
         if not tag.name.startswith("v"):
             continue
         try:
@@ -56,18 +46,15 @@ def _get_available_versions_old(repo_id: str) -> dict[Version, GitRefInfo]:
     return versions
 
 
-def resolve_version_spec_as_ref(
-    repo_id: str, version_spec: int | str
-) -> tuple[GitRefInfo, str]:
+def resolve_version_spec_as_ref(repo_id: str, version_spec: int | str) -> GitRefInfo:
     """
     Get the ref for a kernel with the given version spec.
+
     The version specifier can be any valid Python version specifier:
     https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers
-
-    Returns a tuple of (ref, repo_type).
     """
     if isinstance(version_spec, int):
-        versions, repo_type = _get_available_versions(repo_id)
+        versions = _get_available_versions(repo_id)
 
         ref = versions.get(version_spec, None)
         if ref is None:
@@ -84,7 +71,7 @@ def resolve_version_spec_as_ref(
                 latest_version,
             )
 
-        return ref, repo_type
+        return ref
     else:
         warnings.warn(
             """Version specifiers are deprecated, support will be removed in a future `kernels` version.
@@ -99,7 +86,7 @@ def resolve_version_spec_as_ref(
                 f"No version of `{repo_id}` satisfies requirement: {version_spec}"
             )
 
-        return versions_old[accepted_versions[-1]], "model"
+        return versions_old[accepted_versions[-1]]
 
 
 def select_revision_or_version(
@@ -107,16 +94,14 @@ def select_revision_or_version(
     *,
     revision: str | None,
     version: int | str | None,
-) -> tuple[str, str | None]:
-    """Select a revision, returning (revision, repo_type)."""
+) -> str:
     if revision is not None and version is not None:
         raise ValueError("Only one of `revision` or `version` must be specified.")
 
     if revision is not None:
-        return revision, None
+        return revision
     elif version is not None:
-        ref, repo_type = resolve_version_spec_as_ref(repo_id, version)
-        return ref.target_commit, repo_type
+        return resolve_version_spec_as_ref(repo_id, version).target_commit
 
     warnings.warn(
         "Future versions of `kernels` (>=0.15) will require specifying a kernel version or revision. "
@@ -125,4 +110,4 @@ def select_revision_or_version(
         stacklevel=2,
     )
 
-    return "main", None
+    return "main"
--- a/kernels/src/kernels/_versions.py
+++ b/kernels/src/kernels/_versions.py
@@ -8,12 +8,23 @@ from packaging.version import InvalidVersion, Version
 logger = logging.getLogger(__name__)
 
 
-def _get_available_versions(repo_id: str) -> dict[int, GitRefInfo]:
-    """Get kernel versions that are available in the repository."""
-    from kernels.utils import _get_hf_api
+def _get_available_versions(
+    repo_id: str,
+) -> tuple[dict[int, GitRefInfo], str]:
+    """Get kernel versions that are available in the repository.
+
+    Tries ``"kernel"`` repo type first. If the repository is not found,
+    falls back to ``"model"``.
+
+    Returns a tuple of (versions, repo_type)."""
+    from kernels.utils import _get_hf_api, _resolve_repo_type
+
+    repo_type = _resolve_repo_type(repo_id)
+
+    refs = _get_hf_api().list_repo_refs(repo_id=repo_id, repo_type=repo_type)
 
     versions = {}
-    for branch in _get_hf_api().list_repo_refs(repo_id).branches:
+    for branch in refs.branches:
         if not branch.name.startswith("v"):
             continue
         try:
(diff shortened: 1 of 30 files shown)


kernels/src/kernels/utils.py
```diff
@@ -576,6 +607,36 @@ def _platform() -> str:
...
+            warnings.warn(
+                f"Repository '{repo_id}' uses deprecated repo type 'model'; "
+                "use 'kernel' instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            repo_type = "model"
+
+        except RepositoryNotFoundError:
+            raise
```
Wauplin:
```suggestion
        api.repo_info(repo_id=repo_id, repo_type="model")
        warnings.warn(
            f"Repository '{repo_id}' uses deprecated repo type 'model'; "
            "use 'kernel' instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        repo_type = "model"
```

(nit) doing a try/except/raise without additional logic is the same as not doing try/except at all

kernels/src/kernels/lockfile.py
```diff
@@ -38,32 +39,48 @@ def get_kernel_locks(repo_id: str, version_spec: int | str) -> KernelLock:
...
         )
 
-    if r.siblings is None:
-        raise ValueError(
-            f"Cannot get sibling information for {repo_id} for tag {tag_for_newest.name}"
-        )
+    # Kernel-type repos don't expose siblings in repo_info, fall back
+    # to list_repo_tree (mirroring huggingface_hub's snapshot_download).
+    siblings = getattr(r, "siblings", None)
+    if siblings is None or len(siblings) == 0:
```
Wauplin: I would advice to always do the `list_repo_tree` call and avoid the `files_metadata=True` above. On repos with many files (e.g. kernels repos with many builds) I have found the `.siblings` attribute to be inaccurate in some cases. It's a known issue server-side but can't do much about it (better to not rely on it)

Wauplin (commented):
Thanks for the ping! I've mostly checked the `huggingface_hub` integration

kernels/src/kernels/utils.py
drbh: good catch! cleaned up in latest commit

kernels/src/kernels/lockfile.py
drbh: oh interesting - thanks for the tip! I've updated to default to list_repo_tree in the latest changes

kernels/src/kernels/cli/upload.py
```diff
@@ -124,7 +124,7 @@ def upload_kernels_dir(
 
     is_new_branch = False
     if branch is not None:
-        is_new_branch = not _branch_exists(api, repo_id, branch)
+        is_new_branch = not _branch_exists(api, repo_id, branch, repo_type="model")
```
danieldk: Fixed to `model` because `kernels upload` is deprecated?

kernels/src/kernels/lockfile.py
```diff
@@ -38,32 +39,43 @@ def get_kernel_locks(repo_id: str, version_spec: int | str) -> KernelLock:
...
     )
     if r.sha is None:
         raise ValueError(
             f"Cannot get commit SHA for repo {repo_id} for tag {tag_for_newest.name}"
         )
 
-    if r.siblings is None:
-        raise ValueError(
-            f"Cannot get sibling information for {repo_id} for tag {tag_for_newest.name}"
+    siblings = [
```
danieldk, line 65: Why can r.siblings not be used? Maybe worth adding a comment here?

danieldk (commented):
Great PR! Added two comments.

My main worry is that some public functions (`install_kernel`, `get_variants`, may have missed some) get a public `repo_type` argument, which we'll have to remove again if we switch fully to kernel-repos only. Is there some way we can get rid of them? (e.g. by doing automatic detection inside the functions)

kernels/src/kernels/lockfile.py
drbh: siblings cant be used anymore since that attribute is only supported on model repo types. internal convo here https://huggingface.slack.com/archives/C090JN2P8NB/p1775657277734999?thread_ts=1775656845.729939&cid=C090JN2P8NB

note this is similar to how hfh handles this too

kernels/src/kernels/cli/upload.py
drbh: yea exactly, this is because the python upload only support uploading model repo prior to being deprecated, so now all of the call explicitly use `repo_type="model"` instead of updating the deprecated path for kernel types. happy to change if it makes more sense to upload to kernel repos, but intentionally avoided adding the new functionality for now.

sayakpaul: Yeah no point updating deprecated functionality.

kernels/src/kernels/_versions.py
```diff
@@ -30,10 +41,11 @@ def _get_available_versions_old(repo_id: str) -> dict[Version, GitRefInfo]:
 
     This is for the old tag-based versioning scheme.
     """
-    from kernels.utils import _get_hf_api
+    from kernels.utils import _get_hf_api, _resolve_repo_type
 
+    repo_type = _resolve_repo_type(repo_id)
     versions = {}
-    for tag in _get_hf_api().list_repo_refs(repo_id).tags:
+    for tag in _get_hf_api().list_repo_refs(repo_id, repo_type=repo_type).tags:
```
sayakpaul: (nit): Similar to the above function, do we want?

```py
tags = _get_hf_api().list_repo_refs(repo_id, repo_type=repo_type).tags
````

File: kernels/src/kernels/_versions.py [hunk 6]
```diff
@@ -44,15 +56,19 @@ def _get_available_versions_old(repo_id: str) -> dict[Version, GitRefInfo]:
...
-
+    Get the ref for a kernel with the given version spec.
     The version specifier can be any valid Python version specifier:
     https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers
+
+    Returns a tuple of (ref, repo_type).
     """
     if isinstance(version_spec, int):
-        versions = _get_available_versions(repo_id)
+        versions, repo_type = _get_available_versions(repo_id)
```
sayakpaul: `_get_available_versions()` returning `versions` as well as `repo_type` looks confusing to me. Isn't it better to just call `_resolve_repo_type(repo_id)` to determine the `repo_type`?

kernels/src/kernels/utils.py
```diff
@@ -576,6 +601,35 @@ def _platform() -> str:
...
+
+    api.repo_info(repo_id=repo_id, repo_type="model", revision=revision)
+    # TODO: add back the deprecation warning when we remove support for the "model" repo type.
+    # we should also add a removal date in the warning message for a final migration date.
+    # warnings.warn(
+    #     f"Repository '{repo_id}' uses deprecated repo type 'model'; "
+    #     "use 'kernel' instead.",
+    #     DeprecationWarning,
+    #     stacklevel=2,
+    # )
```
sayakpaul: Should we not add it right now? If so, we could add a deprecation date (maybe two release cycles) and instructions to do it the proper way (re-uploading the repo as a `kernel` type)?

IIUC, after the model repos are removed, the above LOC will definitely fail. Deprecation shouldn't mean complete removal; it should indicate that removal is coming and change is needed. 

Or am I misunderstanding something?

sayakpaul (commented):
Thanks for the PR!

Main comment is using `_resolve_repo_type()` for fetching the repo type inline instead of calling it as a part of other functions such as `select_revision_or_version`. This feels more idiomatic to me.

Also, I think we should add the deprecation now. If we do so, we could also add a test to make sure deprecation warnings are raised accordingly.

kernels/src/kernels/cli/upload.py
```diff
@@ -70,7 +68,6 @@ def _upload_build_dir(
             repo_id=repo_id,
             operations=list(chunk),
             revision=revision,
-            repo_type="model",
```
Wauplin, line 73: no `repo_type="kernel"` here? (same above)

kernels/tests/test_basic.py
```diff
@@ -257,7 +260,7 @@ def test_local_overrides(monkeypatch, local_kernel_path):
 
     # Ensure that we are testing with a non-existing kernel, so that we know
     # that the kernel must be local.
-    with pytest.raises(RepositoryNotFoundError):
+    with pytest.raises(HfHubHTTPError):
```
Wauplin, line 263: `RepositoryNotFoundError` was more specific. Did it change? Or it is that `RepositoryNotFoundError` is not raised for kernel repos (and in such a case we should fix it on `huggingface_hub` side)

kernels/tests/test_kernel_locking.py
```diff
@@ -75,6 +75,7 @@ def forward(self) -> str:
         assert version() == "1"
 
 
+@pytest.mark.skip(reason="Tags are not supported on kernel repos")
```
Wauplin, line 78: > "Tags are not supported on kernel repos"

Is this something that you need resolved on `huggingface_hub` side? If yes I can work on a PR, if not maybe just remove the test/feature if it doesn't exist anymore

kernels/tests/test_layer.py
```diff
@@ -1128,6 +1128,7 @@ def test_kernel_modes_cross_fallback():
         assert linear.n_calls == 2
 
 
+@pytest.mark.skip(reason="Tags are not supported on kernel repos")
```
Wauplin: same questions here for tags and RepositoryNotFoundError

Wauplin (commented):
Made a new pass on the PR and dropping `"model"` indeed made the implementation simpler ^^

I've left some comments, mostly to know if you need new things to be updated on the `huggingface_hub` side or not. 

(also a nit, the PR title and description are outdated now)

kernels/src/kernels/cli/upload.py
drbh: no `repo_type="kernel"` in this case since we plan to fully remove this upload file and all of the upload logic on the python cli in a following PR. 

I removed `repo_type="model"` but realize that was unneeded - however in this case I don't think is technically changes any of the logic since if I understand correctly the default functionality is to push to model repos.

however as mentioned this will all be removed in the next PR I open since we are cleaning up deprecated paths for the v0.14.0 release

kernels/tests/test_basic.py
drbh: oh thanks good catch, `RepositoryNotFoundError` is a better error but I think throw kernel repos throw a 401 instead of a 404 when a repo cannot be found (example in this [error](https://github.com/huggingface/kernels/actions/runs/24454980712/job/71453213659)). We may also need a update on the server side to improve the error. 

not sure if we should block this PR with that update, maybe it can be a small followup PR? what do you think?

kernels/tests/test_kernel_locking.py
drbh: I'm not sure we need any changes on the `huggingface_hub` side, this test is skipped because it tests the backward compatible path where we used tags instead of branches.

this test and the `_old` locking path is also going to be removed in a follow up PR as this path is deprecated an we are removing support in v0.14.0

kernels/tests/test_layer.py
drbh: tried to use `RepositoryNotFoundError` but ran into the 401 issue noted above. I've reverted back to `HfHubHTTPError` so CI passes but also this test will be removed in a follow up PR

thanks for pointing these things out!

kernels/src/kernels/utils.py
```diff
@@ -132,6 +132,7 @@ def _import_from_path(
 def install_kernel(
     repo_id: str,
     revision: str,
+    *,
```
danieldk, line 134: This is an API break, because what used to be positional arguments are now not anymore. *If* we want to break the API (I think there is something to be said for it), we should put the asterisk after `repo_id`, because it removes the potential confusion of two string arguments.

kernels/src/kernels/utils.py
```diff
@@ -235,14 +238,17 @@ def _find_kernel_in_repo_path(
 def install_kernel_all_variants(
     repo_id: str,
     revision: str,
+    *,
```
danieldk, line 240: Same here.

kernels/src/kernels/utils.py
```diff
@@ -318,7 +324,10 @@ def get_kernel(
         backend=backend,
     )
     package_name, variant_path = install_kernel(
-        repo_id, revision=revision, backend=backend, user_agent=user_agent
+        repo_id,
+        revision,
```
danieldk: Not using a kwarg for revision seems like a regression. Also see the other comments about kwargs.

kernels/src/kernels/cli/upload.py
Wauplin: > I removed repo_type="model" but realize that was unneeded - however in this case I don't think is technically changes any of the logic since if I understand correctly the default functionality is to push to model repos.

yes indeed, "model" is the default repo type

> no repo_type="kernel" in this case since we plan to fully remove this upload file and all of the upload logic on the python cli in a following PR.

Got it!

kernels/tests/test_basic.py
Wauplin: >  I think throw kernel repos throw a 401 instead of a 404 when a repo cannot be found

That depends whether user is authenticated or not. It's not specific to kernels. You can test that by opening 
- https://huggingface.co/api/models/huggingface/doesnt_exist
- https://huggingface.co/api/kernels/huggingface/doesnt_exist

in your browser. In normal mode you get a 404 and in incognito (i.e. logged out) you get a 401. This is because repo might exist and be private (so unauthorized). The theory behind it is debatable, but low prio as we could argue for both 401 and 404 .

So here I would advice to either make authenticated calls or catch `HfHubHTTPError`, it's fine

kernels/tests/test_kernel_locking.py
Wauplin: ok great!

kernels/src/kernels/utils.py
drbh: update in latest to enforce kwargs and moved `*` after `repo_id`

kernels/src/kernels/utils.py
drbh: same as above

kernels/src/kernels/utils.py
drbh: thanks for pointing out, updated as mentioned above

danieldk (approved):

kernels/tests/test_basic.py
drbh: ah okay that makes sense thanks for the info! will keep `HfHubHTTPError` for now

PR #438, opened 2026-04-09
kernels: remove unused `init`/`generate-readme` subcommands

danieldk:
`init` was never provided in a release, so we do not need the deprecation error. `generate-readme` was not documented and I don't think it was ever used by anybody.


drbh (approved):
LGTM

PR #447, opened 2026-04-13
feat: default to uploading kernel repo type

drbh:
this PR defaults to uploading to kernel repo type kernels


drbh:
Note: this PR also fixes a latent test bug introduced in https://github.com/huggingface/kernels/pull/322. 

The upload tests were creating an invalid build variant layout (`torch-universal` without `metadata.json`), but that path was not actually being exercised in CI because the workflow step had a broken matrix condition.

danieldk (approved):

PR #456, opened 2026-04-15
move skills to builder.

### LINKED ISSUE #442: Move kernels skills out of the kernels package and transfer to builder

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/442

Main change:

`kernels/src/kernels/cli/skills.py` -> `kernel-builder/scripts/install_skills.py` (standalone script with its own argparse)

Otherwise, we'd need a CLI implemented in Rust. I think that would be overkill for now, for something that's experimental?

The new usage is `python kernel-builder/scripts/install_skills.py --claude`.


kernel-builder/src/main.rs
```diff
@@ -215,6 +223,40 @@ enum Commands {
     },
 }
 
+#[derive(Debug, Subcommand)]
+enum SkillsCommands {
+    /// Install a kernels skill for an AI assistant.
+    Add {
+        /// Skill ID to install (default: cuda-kernels). Use `rocm-kernels` for ROCm-focused kernels.
+        #[arg(long, default_value = skills::DEFAULT_SKILL_ID)]
+        skill: String,
```
danieldk: We can make this an enum and let clap reject invalid options: https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#enumerated-values

kernel-builder/src/skills.rs
```diff
@@ -0,0 +1,163 @@
+use std::fs;
+use std::path::PathBuf;
+
+use eyre::{Context, Result};
+
+pub const DEFAULT_SKILL_ID: &str = "cuda-kernels";
+pub const SUPPORTED_SKILL_IDS: &[&str] = &["cuda-kernels", "rocm-kernels"];
```
danieldk: Not needed, see above.

kernel-builder/src/skills.rs
```diff
@@ -0,0 +1,163 @@
...
+    claude: PathBuf,
+    opencode: PathBuf,
+}
+
+fn global_targets() -> Result<Targets> {
+    let home = dirs::home_dir().ok_or_else(|| eyre::eyre!("Cannot determine home directory"))?;
+    Ok(Targets {
+        codex: home.join(".codex/skills"),
+        claude: home.join(".claude/skills"),
+        opencode: home.join(".config/opencode/skills"),
```
danieldk: I think this should use https://docs.rs/dirs/latest/dirs/fn.config_dir.html

kernel-builder/src/skills.rs
```diff
@@ -0,0 +1,163 @@
...
+        .collect();
+    Ok(entries)
+}
+
+fn download_file(skill_id: &str, rel_path: &str) -> Result<String> {
+    let url = format!("{}/{rel_path}", github_raw_base(skill_id));
+    download(&url)
+}
+
+fn remove_existing(path: &PathBuf) -> Result<()> {
```
danieldk:
```suggestion
fn remove_existing(path: impl AsRef<Path>) -> Result<()> {
    let path = path.as_ref();
```

kernel-builder/src/skills.rs
```diff
@@ -0,0 +1,163 @@
...
+        fs::remove_file(path).wrap_err_with(|| format!("Cannot remove {}", path.display()))?;
+    } else if path.is_dir() {
+        fs::remove_dir_all(path)
+            .wrap_err_with(|| format!("Cannot remove directory {}", path.display()))?;
+    }
+    Ok(())
+}
+
+fn install_to(target: &PathBuf, force: bool, skill_id: &str) -> Result<PathBuf> {
+    let target = fs::canonicalize(target).unwrap_or_else(|_| target.clone());
```
danieldk:
```suggestion
fn install_to(target: impl AsRef<Path>, force: bool, skill_id: &str) -> Result<PathBuf> {
    let target = fs::canonicalize(target.as_ref()).unwrap_or_else(|_| target.clone());
```

Also, the latter part seems wrong? I think we want to fail if the path cannot be canonicalized?

kernel-builder/src/skills.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    if dest.exists() {
+        if !force {
+            eyre::bail!(
+                "Skill already exists at {}.\nRe-run with --force to overwrite.",
+                dest.display()
+            );
+        }
+        remove_existing(&dest)?;
+    }
```
danieldk, line 114: If this happens, the target has already been partially manipulated (see removals, and dir creation above).

The `FileSet` API was added to `kernel-builder` for exactly this reason. You can write a file set as an atomic operation/transaction. If a file exists already, the whole set of changes is not applied.

kernel-builder/src/skills.rs
```diff
@@ -0,0 +1,163 @@
...
+    Ok(dest)
+}
+
+pub fn add_skill(
+    skill_id: &str,
+    claude: bool,
+    codex: bool,
+    opencode: bool,
+    global: bool,
+    dest: Option<PathBuf>,
```
danieldk, line 135: `PathBuf` is fine here, since this is the top-level function.

kernel-builder/src/skills.rs
```diff
@@ -0,0 +1,163 @@
...
+    force: bool,
+) -> Result<()> {
+    if !SUPPORTED_SKILL_IDS.contains(&skill_id) {
+        let supported = SUPPORTED_SKILL_IDS.join(", ");
+        eyre::bail!("Unsupported skill '{skill_id}'. Supported skills: {supported}");
+    }
```
danieldk: See above, exclude through clap parsing.

kernel-builder/Cargo.toml
```diff
@@ -28,6 +28,8 @@ serde-value = "0.7"
 thiserror = "1"
 toml = "0.8"
 url = { version = "2", features = ["serde"] }
+ureq = "3"
```
danieldk: We already have `reqwest` in the dependency tree (for huggingface-hub). Maybe using `reqwest` here as well is better than another dependency (+ transitive dependencies) for the same functionality.

danieldk (commented):
Looks great! Added a bunch of comments.

kernel-builder/src/skills.rs
```diff
@@ -0,0 +1,182 @@
...
+use clap::ValueEnum;
+use eyre::{Context, Result};
+
+use crate::pyproject::FileSet;
+
+const GITHUB_RAW_BASE_TEMPLATE: &str =
+    "https://raw.githubusercontent.com/huggingface/kernels/main/kernel-builder/skills";
+
+#[derive(Clone, Debug, Default, ValueEnum)]
+pub enum SkillId {
```
danieldk, line 13: Minor nit: just `Skill` seems like a better name?

danieldk (approved):
Looks great, added one minor nit, feel free to implement or ignore.

PR #462, opened 2026-04-15
Prefer ruff formatting

drbh:
this pr update the repo to prefer using `ruff` formatting over `black` and `isort`. additionally it adds `griffe` to add a check for any breaking changes to public functions


drbh:
I think the failing test is unrelated and introduced in https://github.com/huggingface/kernels/pull/467?

.github/workflows/lint.yml
```diff
@@ -6,46 +6,39 @@ jobs:
...
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
-      - name: Run ruff
+      - name: Run ruff check
         uses: astral-sh/ruff-action@4919ec5cf1f49eff0871dbcea0da843445b837e6  # v3
-        src: >-
-          kernels
+        with:
+          src: kernels
+          version: "0.15.10"
```
sayakpaul, line 13: Do we always want to run with latest `ruff`?

.github/workflows/lint.yml
```diff
@@ -6,46 +6,39 @@ jobs:
...
+        uses: astral-sh/ruff-action@4919ec5cf1f49eff0871dbcea0da843445b837e6  # v3
+        with:
+          src: kernels
+          version: "0.15.10"
+          args: format --check
 
-  black:
-    name: Run black check
+  griffe:
+    name: Check API compatibility
```
sayakpaul, line 22: Ou jazzy. I didn't know this!

kernels/pyproject.toml
```diff
@@ -85,4 +82,6 @@ line-length = 119
 # Ignored rules:
 # "E501" -> line length violation
 lint.ignore = ["E501"]
-lint.select = ["E", "F", "W"]
+lint.select = ["E", "F", "I", "W"]
```
sayakpaul, line 87: Do we want to also instruct ruff to take care of sorting the imports, remove unused imports, etc. if not already done?

sayakpaul (commented):
Just one major comment.

.github/workflows/lint.yml
danieldk: I think pinning the ruff version is not a bad idea, since there will be subtle formatting differences between versions and then if CI uses the latest and a local venv or Nix env uses a different version, then there will always be mismatches.

kernels/pyproject.toml
danieldk: It looks like `I` enables import formatting, resorting, and it is run as part of linting:

https://github.com/astral-sh/ruff/issues/8926#issuecomment-1834048218

Since we are running `ruff check --fix`, it should be good I think.

sayakpaul (approved):

PR #464, opened 2026-04-16
kernel-builder: detect card in `result`, `build`, or target dir

danieldk:
Before this change, the card would only be detected in the `build` directory. However, we often only have a `result` directory or specify the build directory directly.

### DIFF
--- a/kernel-builder/src/upload.rs
+++ b/kernel-builder/src/upload.rs
@@ -764,7 +764,7 @@ mod tests {
     }
 
     #[test]
-    fn test_discover_build_file_in_kernel_dir() {
+    fn test_discover_build_file_in_fully_specified_build_dir() {
         let temp_dir = tempfile::tempdir().unwrap();
         let kernel_dir = temp_dir.path();
 
--- a/kernel-builder/src/upload.rs
+++ b/kernel-builder/src/upload.rs
@@ -293,13 +293,13 @@ fn collect_benchmark_commit_ops(
 
 /// Collect README commit operation: upload build/CARD.md as README.md.
 fn collect_readme_commit_ops(kernel_dir: &Path, operations: &mut Vec<CommitOperation>) {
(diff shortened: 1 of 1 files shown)


kernel-builder/src/upload.rs
```diff
@@ -356,6 +356,35 @@ fn collect_build_commit_ops(
...
 
+fn discover_build_file(
+    kernel_dir: impl AsRef<Path>,
+    filename: impl AsRef<Path>,
+) -> Result<PathBuf> {
+    let kernel_dir = kernel_dir.as_ref();
+    let filename = filename.as_ref();
+
+    let candidates = [
+        kernel_dir.join("result").join(filename),
```
sayakpaul, line 367: But is it a valid build in case we don't have the `build` dir and hence, no point doing the system card?

kernel-builder/src/upload.rs
```diff
@@ -698,4 +727,58 @@ mod tests {
...
+        let temp_dir = tempfile::tempdir().unwrap();
+        let kernel_dir = temp_dir.path();
+
+        let target = kernel_dir.join("CARD.md");
+        fs::write(&target, "# Test card").unwrap();
+
+        let found = discover_build_file(kernel_dir, "CARD.md").unwrap();
+        assert_eq!(found, target);
+    }
+
```
sayakpaul, line 777: Should there be a test where a target dir is explicitly specified?

sayakpaul (dismissed):
Just one question.

kernel-builder/src/upload.rs
danieldk: Yes, normally when we do `kernel-builder build`, we only have the `result` symlink. `build` only gets created with `kernel-builder build-and-copy`.

kernel-builder/src/upload.rs
danieldk: That's actually this one, the test name does not clearly communicate that, let me update the name.

PR #210, opened 2026-01-14
add utilities to generate template repo cards

sayakpaul:
As discussed [internally](https://huggingface.slack.com/archives/C090JN2P8NB/p1767506604945619), this PR adds utilities to easily generate template repository cards for the kernels built with `kernels` and `kernel-builder`.

For now, the repository card contains basic info, which IMO, is sufficient for a v1. But of course, feedback is welcome.

If we run 

```
python example_usage.py --kernels_dir /fsx/sayak/kernels-community/activation --card_path activation.md
```

We should expect to get an `activation.md`, looking like so:

<details>
<summary>Unfold</summary>

---
library_name: kernels
license: apache-2.0
---

This is the repository card of {repo_id} that has been pushed on the Hub. It was built to be used with the [`kernels` library](https://github.com/huggingface/kernels). This card was automatically generated.

## How to use

```python
# make sure `kernels` is installed: `pip install -U kernels`
from kernels import get_kernel

kernel_module = get_kernel("REPO_ID") # <- change the ID if needed
silu_and_mul = kernel_module.silu_and_mul

silu_and_mul(...)
```

## Benchmarks

[TODO: provide benchmarks if available]

## Code source

[TODO: provide original code source and other relevant citations if available]

## Notes

[TODO: provide additional notes about this kernel if needed]

</details>

On the Hub, this would preview much better, of course. You can preview [here](https://huggingface.co/sayakpaul/example-kernels-repocard).

## TODOs

- [ ] Tests
- [x] Docs
- [x] `upload-card` CLI command
- [x] Remove the `example_usage.py` script
- [ ] Discuss how this should be integrated in the build workflow (or any other workflow for that matter)

## From Slack

> We could have an upload-card subcommand (or even make it part of upload) that uploads the card to main on the Hub.

I agree to this. I will work on it in this PR later.

@danieldk @drbh the above are some TODOs that came to mind, but the list isn't limited to that. Would love an initial review of this draft and address anything.

PR #470, opened 2026-04-16
Move card filling to Nix

danieldk:
This ensures that filled cards are in `result` as well (not just `build`) and keeps people that use Nix directly (i.e. me :-)) happy.


examples/kernels/cutlass-gemm-tvm-ffi/build.toml
```diff
@@ -5,6 +5,9 @@ backends = [
     "xpu",
 ]
 
+[general.hub]
+repo-id = "kernels-test/cutlass-gemm-tvm-ffi"
```
sayakpaul, line 9: I am guessing the token is configured to write to `kernels-test` too?

examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
```diff
@@ -0,0 +1,48 @@
+---
+library_name: kernels
```
sayakpaul, line 2: Auto-generated? 👀

drbh (dismissed):
lgtm!

question, this seems like this change will make the CARD.md non optional since if I'm reading correctly `fill-card` will throw if it cant find the template or if theres no repo-id? enforcing a card seems reasonable but just want to understand the change fully

sayakpaul (commented):
Should we add a test too?

danieldk:
> question, this seems like this change will make the CARD.md non optional since if I'm reading correctly `fill-card` will throw if it cant find the template or if theres no repo-id? enforcing a card seems reasonable but just want to understand the change fully

Great catch. I first used a hook where I also checked the presence of `CARD.md`, but I forgot to port it to the current bundle post-install.

examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
danieldk: Not sure? This comes from the template that `init` uses:

https://github.com/huggingface/kernels/blob/main/kernel-builder/src/init/templates/CARD.md?plain=1

I have no idea where it comes from, maybe @drbh ?

examples/kernels/cutlass-gemm-tvm-ffi/build.toml
danieldk: We don't actually upload these. I just needed to give them an identifier. We could also consider changing these to `example/<name>`, though that org seems to exist:

https://huggingface.co/Example

examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
drbh: yea, added because I was using the output from the previous [`ModelCardData`](https://github.com/huggingface/huggingface_hub/blob/370dde27b740912f9ec46f047c9a1583860315d3/src/huggingface_hub/repocard_data.py#L282) output which included the [library name field](https://github.com/huggingface/kernels/pull/210/changes#diff-c2cbcaf996f388539865041a4790cecbaceb5500517708e910f2fa742a40bb73R59)

that being said its not fully clear if we need it since its really a model loading concept... maybe we can remove (no strong preference)

danieldk: I'll leave that to a separate PR. No preference either.

drbh (approved):
👌

.github/workflows/build_kernel.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       group: ${{ matrix.runner }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25
```
sayakpaul, line 27: Good catch!

.github/workflows/nix_checks.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: nix fmt -- --ci
+      - name: Nix checks
+        run: nix build .\#checks.x86_64-linux.default
+      - name: Test kernel card generation
+        run: |
+          (
+            cd examples/kernels/silu-and-mul
+            nix build -L .
+            test -e result/CARD.md
+          )
```
sayakpaul, line 37: Hmm, do we need to trigger an entire build to check for the card stuff? Seems like an overkill to me 👀

sayakpaul (commented):
Just one comment concerning test as it seems like we're building first then validate card content.

.github/workflows/nix_checks.yml
sayakpaul: Or I guess it's simple and doesn't cost much?

danieldk: Yeah, it's a noarch kernel, so cheap. We need to do a bundle build to generate the card.

sayakpaul (approved):
Thanks for iterating!

.github/workflows/build_kernel.yaml
H1235515648: 1.hsk hhhs 😂

Issue #465, opened 2026-04-16
Include capabilities for noarch kernels in build.toml and export to metadata


sayakpaul:

sayakpaul:
This can be closed as we have added support for this. Cc: @danieldk

PR #471, opened 2026-04-16
Revamp metadata writing and add kernel id

danieldk:
Changes in kernel-builder:

* Strongly type the kernel identifier within kernel-builder.
* Write the kernel identifier to the kernel metadata.
* Make noarch kernels include the backend in the identifier as well.
* Fix tvm-ffi build-time metadata update to include the arches.
* Rename build-time metadata update script to be more general.
* Run the build-time metadata update script for non-CUDA/ROCm backends. This is not strictly necessary now, but allows us to add other metadata bits in the future.
* Add a build hook that validates the metadata after build using the Rust-side data structures. This ensures that build-time changes are valid and schema-conforming.

Changes in kernels:

* Support the kernel id metadata.
* Use the kernel id (when available) as the module identifier in `sys.modules`, avoiding the path hash.


docs/source/kernel-requirements.md
```diff
@@ -36,16 +36,51 @@ must be available for that combination.
...
+- `id` (`str`, required): a unique identifier for the kernel. This
+  identifier must also be a valid Python module name. If the kernel
+  registers Torch ops, they must be registered as `torch.ops.<id>`
+- `version` (`int`, required): the kernel version number.
+- `backend` (`dict`, required): information about the compute backend that
+  this build variant supports.
+- `python-depends` (`list[str]`, optional): list of Python dependencies
+  from a curated set of Python dependencies.
+
+Example `metadata.json`:
```
sayakpaul, line 51: I think should also contain the `id` and the other new keys?

docs/source/kernel-requirements.md
```diff
@@ -36,16 +36,51 @@ must be available for that combination.
...
 {
   "python-depends": ["einops"],
   "version": 1
+}
+```
+
+The `metadata.json` file is generated automatically by `kernel-builder`,
+so creating this file is only required for kernels that are built with
+a different tool. For such kernels, we strongly recommend generating
+`metadata.json` using the [`kernels-data`](https://github.com/huggingface/kernels/tree/main/kernels-data)
```
sayakpaul: Very cool! I think we should also have some guidance on how to do it?

docs/source/kernel-requirements.md
```diff
@@ -36,16 +36,51 @@ must be available for that combination.
...
+{
+  # ...
+  "backend": {
+    "type": "cuda",
+    "archs": ["7.0", "7.2", "7.5", "8.0", "8.6", "8.7", "8.9", "9.0+PTX"]
+  }
+}
 ```
 
+The backend `type` must be one of `cpu`, `cuda`, `metal`, `neuron`, `rocm`,
```
sayakpaul: CANN as well?

Wonder if we could have something like 

```py
from kernels import get_all_supported_backends

print(get_all_supported_backends())
```

No strong opinions.

kernel-builder/src/pyproject/templates/torch/noarch/_ops.py
```diff
@@ -1,8 +1,34 @@
 import torch
-ops = torch.ops.{{ ops_name }}
 
-def add_op_namespace_prefix(op_name: str):
+def get_backend() -> str:
```
sayakpaul, line 3: Is CANN missing here?

kernel-builder/src/pyproject/templates/torch/noarch/_ops.py
```diff
@@ -1,8 +1,34 @@
...
+def _find_ops_name() -> str:
+    kernel_name = "{{ kernel_name }}"
+    unique_id = "{{ kernel_unique_id }}"
+    backend = get_backend()
+    return f"_{kernel_name}_{backend}_{unique_id}"
+
+
+_OPS_NAME = _find_ops_name()
+
+ops = getattr(torch.ops, _OPS_NAME)
```
sayakpaul, line 32: Help me unpack this a little. `_find_ops_name()` seems to have placeholders that are not fully resolved at this point. Same for `add_op_namespace_prefix()`

kernel-builder/src/pyproject/templates/torch/add_build_metadata.py
```diff
@@ -16,20 +16,19 @@ def main():
         "destination",
         help="Path to write the output metadata JSON file to.",
     )
-    parser.add_argument(
```
sayakpaul, line 19: Are there no additional instances of `args.backend` in the module?

kernels/src/kernels/metadata.py
```diff
@@ -1,40 +1,40 @@
...
-        return Metadata(version=None, python_depends=[])
-
     @staticmethod
     def load_from_variant(variant_path: Path) -> "Metadata":
         metadata_path = variant_path / "metadata.json"
         if metadata_path.exists():
             with open(metadata_path, "r") as f:
                 metadata_dict = json.load(f)
+                if (kernel_id := metadata_dict.get("id", None)) is None:
+                    warnings.warn(
```
sayakpaul, line 23: Should we provide guidance to users on how they should do to have this metadata?

kernels/src/kernels/metadata.py
```diff
@@ -1,40 +1,40 @@
...
+                        stacklevel=2,
+                    )
                 return Metadata(
+                    id=kernel_id,
                     python_depends=metadata_dict.get("python-depends", []),
                     version=metadata_dict.get("version", None),
                 )
 
-        return Metadata(version=None, python_depends=[])
+        warnings.warn(
```
sayakpaul, line 36: Same.

kernels/src/kernels/metadata.py
```diff
@@ -1,40 +1,40 @@
...
 
 @strict
 @dataclass
 class Metadata:
+    id: str | None
     python_depends: list[str]
     version: int | None
 
-    @staticmethod
-    def load_from_build_toml(build_toml_path: Path) -> "Metadata":
```
sayakpaul, line 17: Ah so we are doing explicit metadata.json hence not needed?

sayakpaul (commented):
Thanks!

I left some questions. Does it affect https://github.com/huggingface/kernels/issues/465?

Also, we should probably add some tests as well?

docs/source/kernel-requirements.md
danieldk: I think we should keep kernels self-sufficient, so that e.g. local loading without `kernels` also works.

kernel-builder/src/pyproject/templates/torch/noarch/_ops.py
danieldk: CANN is not supported by kernel-builder, only by kernels. However neuron has limited support, so I added it to this function.

kernel-builder/src/pyproject/templates/torch/noarch/_ops.py
danieldk: These are filled by `kernel-builder`:

https://github.com/huggingface/kernels/blob/42b958b7535175d68b0ebf7ebb6ef8ed5231268d/kernel-builder/src/pyproject/torch/noarch.rs#L48

kernels/src/kernels/metadata.py
danieldk: This function was not referenced anywhere anymore, so I removed it.

kernels/src/kernels/metadata.py
danieldk: Yeah, that would be much nicer, done!

kernels/src/kernels/metadata.py
danieldk: Done

kernel-builder/src/pyproject/templates/torch/add_build_metadata.py
danieldk: No, this option originated from when the script added the backend as well, but we later moved writing the backend to kernel-builder itself (we could do that since we are generating a `metadata-{backend}.json` for every supported backend, which is copied to `metadata.json` during the build).

docs/source/kernel-requirements.md
danieldk: I removed this remark for now. We should add something with instructions once we have something that is more complete and can write the metadata. Though to be honest, I'd encourage people to just use kernel-builder.

docs/source/kernel-requirements.md
danieldk: Added.

kernel-builder/src/pyproject/templates/torch/noarch/_ops.py
sayakpaul: Oh cool. Perhaps a small explainer comment could helpful for future reference and it's not immediately clear where this will be used. No strong opinions.

kernels/src/kernels/utils.py
```diff
@@ -105,21 +108,25 @@ def _import_from_path(module_name: str, variant_path: Path, _repo_infos: RepoInf
     # it would also be used for other imports. So, we make a module name that
     # depends on the path for it to be unique using the hex-encoded hash of
     # the path.
-    package_name = module_name
-    path_hash = "{:x}".format(ctypes.c_size_t(hash(file_path)).value)
-    module_name = f"{module_name}_{path_hash}"
-    spec = importlib.util.spec_from_file_location(module_name, file_path)
+    if metadata.id is None:
```
sayakpaul, line 111: Should a warning be thrown in line with the previous warning?

```
f"Kernel loaded from `{variant_path}` does not have metadata,"
```

Or is it a fair assumption that `LoadedKernel` can't always guarantee `metadata.id`?

sayakpaul (approved):
Looks great! I think it'd be good to have the tests for `kernels` to include some tests for the changes introduced?

kernels/src/kernels/utils.py
danieldk: The warning will already by emitted by: https://github.com/huggingface/kernels/blob/793d303c4e72b87329f2ab748536e5b1d75159c2/kernels/src/kernels/metadata.py#L23

PR #473, opened 2026-04-16
Remove unneeded backward compat

drbh:
this PR removes items that will not be supported in future kernels versions

- [X] python based upload path
- [X] version specifier's locked and fetching

note: this PR is currently branched from https://github.com/huggingface/kernels/pull/462 but will be rebased once the formatting changes are merged

another note: this PR is expected to fail the `Lints / Check API compatibility` workflow since we are intentionally removing support for backward compatibility of some features


.github/workflows/lint.yml
```diff
@@ -6,46 +6,39 @@ jobs:
...
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
-      - name: Run ruff
+      - name: Run ruff check
         uses: astral-sh/ruff-action@4919ec5cf1f49eff0871dbcea0da843445b837e6  # v3
-        src: >-
-          kernels
+        with:
+          src: kernels
+          version: "0.15.10"
```
danieldk: Maybe we can have some version variables at the top of the file (for non-workflows)? Ruff below uses the same version.

.github/workflows/lint.yml
```diff
@@ -6,46 +6,39 @@ jobs:
...
-
-      - name: Install black
-        run: uv pip install black
-
-      - name: Check formatting
-        run: |
-          uv run black --check kernels
+          python-version: "3.12"
+      - name: Check for breaking changes
+        run: uvx griffe check kernels --search kernels/src -a main
```
danieldk: Do we need to lock the `griffe` version too?

danieldk (commented):
Started reviewing and added two small comments, but I think it's better to review once the styling PR is merged and this is rebased. Hard to distinguish the formatting changes from the rest.

.github/workflows/lint.yml
drbh: sounds good, I've added vars for the version at the top of the file in the latest changes

.github/workflows/lint.yml
drbh: good idea, added a version for griffe at the top of the file too

danieldk (approved):
Yay, end of the version specifiers!

Issue #396, opened 2026-03-24
Wrap `kernels-data` data structures with PyO3 and use in `kernels`


danieldk:
This is to ensure that we use exactly the same parsers in kernel-builder and kernels. Post-0.13.0.

danieldk:
Fixed by #471 .

PR #475, opened 2026-04-17
[core] package kernels-data as a python wrapper.

sayakpaul:
Part of https://github.com/huggingface/kernels/issues/396. The subsequent PR will address validation in the client part of `kernels`.

~Could we configure the access management on PyPI similar to the release workflow of ABI check?~ ✅


kernels-data/src/config/name.rs
```diff
@@ -4,7 +4,7 @@ use regex::Regex;
 use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
 
 /// A validated kernel name matching `^[a-z][-a-z0-9]*[a-z0-9]$`.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
```
sayakpaul, line 7: pyo3's `#[pyclass(..., hash)]` requires the Rust struct to implement `std::hash::Hash` — pyo3 delegates Python's `__hash__` to the derived Rust hash.

Any preferred alternative?

kernels-data/bindings/python/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    Cpu,
+    #[pyo3(name = "CUDA")]
+    Cuda,
+    #[pyo3(name = "METAL")]
+    Metal,
+    #[pyo3(name = "NEURON")]
+    Neuron,
+    #[pyo3(name = "ROCM")]
+    Rocm,
+    #[pyo3(name = "XPU")]
```
danieldk, line 87: I'd mirror the casing that we are currently using in Python-land for the Python names (`name = ...`): https://github.com/huggingface/kernels/blob/178616a80f2d644f3affafa8cbdc9355eb9a250d/kernels/src/kernels/backends.py#L32

I think we also need `CANN`.

kernels-data/bindings/python/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+#[pymethods]
+impl PyMetadata {
+    /// Load `metadata.json` from a build variant directory.
+    ///
+    /// Returns `None` if the file does not exist in the given directory.
+    /// Raises `ValueError` if the file exists but cannot be parsed.
+    #[staticmethod]
+    fn load_from_variant(variant_path: PathBuf) -> PyResult<Option<Self>> {
```
danieldk: I'd instead do `load`, which takes the full path to `metadata.json` and let the caller construct the path.

danieldk: Also not 100% sure if PyO3 has this conversion, but it might be worth trying


```suggestion
    fn load_from_variant(variant_path: impl AsRef<Path>) -> PyResult<Option<Self>> {
```

or  `variant_path: &Path`.

kernels-data/bindings/python/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    #[getter]
+    fn version(&self) -> Option<usize> {
+        self.version
+    }
+
+    #[getter]
+    fn license(&self) -> Option<&str> {
+        self.license.as_deref()
```
danieldk:
```suggestion
        self.license.as_ref()
```

is the correct semantics, we are not dereferencing here.

kernels-data/bindings/python/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    #[getter]
+    fn upstream(&self) -> Option<&str> {
+        self.upstream.as_deref()
```
danieldk: See above.

kernels-data/bindings/python/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    #[getter]
+    fn python_depends(&self) -> Vec<String> {
+        self.python_depends.clone()
+    }
```
danieldk:
```suggestion
    fn python_depends(&self) -> &[String] {
        &self.python_depends
    }
```

prevents double conversion (if PyPI supports this).

kernels-data/bindings/python/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/// Parse a kernel `metadata.json` file.
+///
+/// Raises `ValueError` on any I/O or parse error.
+#[pyfunction(name = "parse_metadata")]
+fn py_parse_metadata(path: PathBuf) -> PyResult<PyMetadata> {
+    parse_metadata(&path)
+        .map(Into::into)
+        .map_err(|err| PyValueError::new_err(format!("{err:#}")))
```
danieldk: Why standalone?

kernels-data/bindings/python/src/lib.rs
sayakpaul: Yeah not supported. Kept PathBuf, which is what pyo3 officially supports for path   
arguments.

sayakpaul:
@danieldk thanks for the review. Your comments should have been addressed now. I have also squeezed in a mypy check in the CI for `kernels-data`.

danieldk (approved):
Looks great!

PR #481, opened 2026-04-17
kernels-data: allow unknown fields in `Metadata`

danieldk:
Now that we will use `kernels-data` inside `kernels`, we have to allow
unknown fields in metadata. Otherwise the following scenario will break:

1. We publish kernels 0.14.
2. We add a field `foo` to the metadata in kernel-builder 0.15.
3. `kernels` 0.14 will not load metadata from kernel-builder 0.15,
   since it contains an unknown (for 0.14) field `foo`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: style kernel-builder-cli-docs quality
 
 
-export check_dirs := kernels/src kernels/tests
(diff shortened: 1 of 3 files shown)


drbh (dismissed):
approving but we'll need to update the test case

drbh (approved):

PR #484, opened 2026-04-18
layer: make _CACHED_LAYER context-local to match _KERNEL_MAPPING

GitGlimpse895:
`_CACHED_LAYER` is a plain module-level `dict`, but `_KERNEL_MAPPING` is a
`ContextVar`. This mismatch means that when `use_kernel_mapping()` is used
concurrently (e.g. multiple threads or async tasks with different mappings),
the layer cache is incorrectly shared across contexts — a repo resolved in
one context can be returned for a different repo selected in another context.

Fix by moving `_CACHED_LAYER` into a `ContextVar` in `globals.py`, alongside
`_KERNEL_MAPPING`. `_get_layer_memoize` is updated to read/write from the
context-local dict. Since `use_kernel_mapping.__enter__` already deep-copies
the mapping context, the layer cache is automatically scoped and invalidated
per-context with no additional logic required.

Changes:
- `globals.py`: add `_CACHED_LAYER` as a `ContextVar`
- `layer.py`: remove module-level `_CACHED_LAYER` dict; update
  `_get_layer_memoize` to use the context-local instance

### DIFF
--- a/kernels/src/kernels/layer/globals.py
+++ b/kernels/src/kernels/layer/globals.py
@@ -1,8 +1,15 @@
 import os
 from contextvars import ContextVar
+from typing import TYPE_CHECKING, Type
 
 from .repos import DeviceRepos
+
+if TYPE_CHECKING:
+    from torch import nn
+    from .repos import RepositoryProtocol
 
 _DISABLE_KERNEL_MAPPING: bool = bool(int(os.environ.get("DISABLE_KERNEL_MAPPING", "0")))
 
 _KERNEL_MAPPING: ContextVar[dict[str, dict[str, DeviceRepos]]] = ContextVar("_KERNEL_MAPPING", default={})
+
+_CACHED_LAYER: ContextVar[dict["RepositoryProtocol", Type["nn.Module"]]] = ContextVar("_CACHED_LAYER", default={})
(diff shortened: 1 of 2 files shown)


kernels/src/kernels/layer/globals.py
```diff
@@ -1,8 +1,15 @@
 import os
 from contextvars import ContextVar
+from typing import TYPE_CHECKING, Type
 
 from .repos import DeviceRepos
 
+if TYPE_CHECKING:
+    from torch import nn
+    from .repos import RepositoryProtocol
```
danieldk, line 9: Any reason to put the `RepositoryProtocol` inside the `TYPE_CHECKING` conditional?

danieldk (commented):
Added a comment on an import and a larger comment below.

danieldk:
I still need to be convinced of this part:

> `_CACHED_LAYER` is a plain module-level `dict`, but `_KERNEL_MAPPING` is a
> `ContextVar`. This mismatch means that when `use_kernel_mapping()` is used
> concurrently (e.g. multiple threads or async tasks with different mappings),
> the layer cache is incorrectly shared across contexts — a repo resolved in
> one context can be returned for a different repo selected in another context.

I am not sure what mismatch would occur, the cache key is a repository. How can a repo resolved in one context be a different repo in another context? For example:

```
_CACHED_LAYER[LayerRepository(repo_id="foo", layer_name="bar", version=1)]
```

should always map to exactly the same module, regardless of context. It would be helpful to add a test that would fail without this PR.

GitGlimpse895:
You're right, the cache key being content-addressed means there's no real collision risk. Closing this — thanks for the review!

PR #487, opened 2026-04-18
fix/rocm properties export

GitGlimpse895:
`ROCMProperties` was fully implemented in `device.py` with its own docstring and example, but was never wired up end-to-end:

- `Device.properties` was typed as `CUDAProperties | None`, silently rejecting `ROCMProperties` at the type-system level.
- `Device.__post_init__` validated `CUDAProperties` placement but had no symmetric guard for `ROCMProperties`, allowing invalid combinations like `Device(type="cuda", properties=ROCMProperties(...))` to pass silently.
- `ROCMProperties` was never exported from `kernels/layer/__init__.py` or the top-level `kernels/__init__.py`, making `from kernels import ROCMProperties` fail despite being shown in the class docstring example.

This PR fixes all three gaps across three files, completing the ROCm capability-range kernel selection feature that was already wired up in `_ROCMRepos`.


GitGlimpse895:
Hey @sayakpaul — wanted to get your eyes on this if you have a moment! This PR wires up `ROCMProperties` end-to-end — fixes the type annotation on `Device.properties`, adds a symmetric validation guard in `__post_init__`, and exports `ROCMProperties` from both `kernels/layer/__init__.py` and the top-level `kernels/__init__.py`. It's a pretty contained fix, but curious what you think — does this approach look right to you, or would you handle the validation guard differently?

danieldk (approved):
Looks good, thanks! Sorry for the delay.

PR #493, opened 2026-04-20
kernels-data: skip serializing archs when `None`

danieldk:


danieldk:
Merging this hotfix to unblock myself :).

Issue #312, opened 2026-03-05
all kernels segfaults on load/import on Python 3.14-nogil


Qubitium:
Env: 

Intel Xeon Granite Rapids
Ubuntu 24.04

Python 3.14-nogil (freethreading enabled)

```
pip show torch transformers kernels
Name: torch
Version: 2.10.0+cpu

Name: transformers
Version: 5.3.0

Name: kernels
Version: 0.12.2
```

@danieldk  @MekkCyber  

I have no idea what's go. Was the kernel incorrectly compiled (possiblity 1), thread race in the importer (strange but possibility 2). 

Run below:

```OMP_NUM_THREADS=32 MKL_NUM_THREADS=32 OPENBLAS_NUM_THREADS=32 NUMEXPR_NUM_THREADS=32 python scripts/repro_hf_kernel_import_crash.py --mode both```

Reproducing Script:

```py
#!/usr/bin/env python3
"""
Standalone reproducer for HF kernels import crashes.

This script intentionally does not import gptqmodel.
It only uses torch + kernels package APIs.
"""

from __future__ import annotations

import argparse
import faulthandler
import os
import platform
import sys
import sysconfig
import traceback
from pathlib import Path


def _print_runtime_env() -> None:
    print("=== Runtime ===", flush=True)
    print(f"python: {sys.version}", flush=True)
    print(f"platform: {platform.platform()}", flush=True)
    print(f"machine: {platform.machine()}", flush=True)
    print(f"Py_GIL_DISABLED: {sysconfig.get_config_var('Py_GIL_DISABLED')}", flush=True)
    if hasattr(sys, "_is_gil_enabled"):
        print(f"sys._is_gil_enabled(): {sys._is_gil_enabled()}", flush=True)

    for key in (
        "PYTHON_GIL",
        "OMP_NUM_THREADS",
        "MKL_NUM_THREADS",
        "OPENBLAS_NUM_THREADS",
        "NUMEXPR_NUM_THREADS",
    ):
        print(f"{key}={os.getenv(key)}", flush=True)

    try:
        import torch

        print(f"torch: {torch.__version__}", flush=True)
        print(f"torch.compiled_with_cxx11_abi(): {torch.compiled_with_cxx11_abi()}", flush=True)
    except Exception as exc:
        print(f"torch import failed: {exc!r}", flush=True)

    try:
        import kernels

        print(f"kernels package: {kernels.__file__}", flush=True)
    except Exception as exc:
[...]

PR #494, opened 2026-04-21
Do not load arch kernels on free-threaded Python

danieldk:
Free-threaded Python interpreters do not support ABI3 before Python 3.15, so there is no way to extensions that use a stable ABI. So to avoid segfaulting, we disable support for arch kernels on such interpreters.

Also see #312.


docs/source/installation.md
```diff
@@ -17,3 +17,8 @@ or if you want the latest version from the `main` branch:
 ```bash
 pip install "kernels[benchmark] @ git+https://github.com/huggingface/kernels#subdirectory=kernels"
 ```
+
+**Note:** we strongly recommend not using a free-threaded Python build yet.
+These builds are not only experimental, but do not support the stable ABI
+on Python versions before 3.15. Kernels are compiled with the stable ABI
+to support a wide range of Python versions.
```
sayakpaul:
```suggestion
> [!IMPORTANT] 
> We strongly recommend not using a free-threaded Python build yet.
These builds are not only experimental, but do not support the stable ABI
on Python versions before 3.15. Kernels are compiled with the stable ABI
to support a wide range of Python versions.
```

kernels/tests/test_basic.py
```diff
@@ -152,9 +152,9 @@ def test_version_old():
 
 def test_version():
     kernel = get_kernel("kernels-test/versions", version=1)
-    assert kernel.version() == "1"
+    assert kernel.version() == 1
```
sayakpaul, line 111: I thought these were already fixed in my previous PRs.

sayakpaul (approved):
Thanks! How about we also add a test to check if warning is properly raised?

sayakpaul (dismissed):

sayakpaul (approved):
I would just add a simple test but other than that, this is good!

PR #495, opened 2026-04-21
Build Python 3.13t + 3.14t wheels of PyO3 packages

danieldk:
Also set the version to 0.14.0.dev1 to trigger new uploads later.


.github/workflows/kernels_data_python_release.yaml
```diff
@@ -45,26 +45,33 @@ jobs:
...
-        uses: PyO3/maturin-action@v1
+      - name: Build free-threaded wheels (3.13t)
+        uses: PyO3/maturin-action@e83996d129638aa358a18fbd1dfb82f0b0fb5d3b # v1.51.0
+        with:
+          target: ${{ matrix.platform.target }}
+          args: --release --out dist --manifest-path kernels-data/bindings/python/Cargo.toml -i python3.13t
+          sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}
+          manylinux: auto
+      - name: Build free-threaded wheels (3.14t)
+        uses: PyO3/maturin-action@e83996d129638aa358a18fbd1dfb82f0b0fb5d3b # v1.51.0
```
sayakpaul, line 67: Nice. Could have been a matrix for 3.13t and 3.14t but no strong opinions.

sayakpaul (approved):
Nice one!

.github/workflows/kernels_data_python_release.yaml
danieldk: Yeah, once it's more we could change it in a matrix. I kept it like this, because if we need to let maturin generate the workflow again in the future, the diff will be smaller, making it easier to cherry-pick just the parts that changed.

PR #496, opened 2026-04-21
[feat] Make utils release

sayakpaul:
So that we can run `make pre-release` and then after release, run `make bump-dev` and save on some Claude calls :)


sayakpaul:
Failing tests seem unrelated to this PR?

scripts/_version_common.py
```diff
@@ -0,0 +1,69 @@
...
+    REPO_ROOT / "kernels-data" / "Cargo.toml",
+    REPO_ROOT / "kernels-data" / "bindings" / "python" / "Cargo.toml",
+    REPO_ROOT / "kernel-builder" / "Cargo.toml",
+    REPO_ROOT / "kernel-abi-check" / "kernel-abi-check" / "Cargo.toml",
+    REPO_ROOT / "kernel-abi-check" / "bindings" / "python" / "Cargo.toml",
+]
+
+# Anchored to the start of a line so inline dep entries like
+# `clap = { version = "4", ... }` are never matched.
+VERSION_LINE_RE = re.compile(r'^version\s*=\s*"([^"]+)"', re.MULTILINE)
```
danieldk: I would parse it with `tomlib` and get the version from the parsed representation. regexps are fragile.

scripts/_version_common.py
```diff
@@ -0,0 +1,69 @@
...
+    original = path.read_text()
+    match = VERSION_LINE_RE.search(original)
+    if match is None:
+        raise SystemExit(f"Could not find a top-level `version = \"...\"` line in {path}.")
+
+    old_version = match.group(1)
+    if old_version == new_version:
+        return None
+
+    updated = VERSION_LINE_RE.sub(f'version = "{new_version}"', original, count=1)
```
danieldk: Modify in the parsed TOML and serialize again.

scripts/bump_to_dev.py
```diff
@@ -0,0 +1,106 @@
...
+
+# Accept plain release versions only (e.g. 0.13.0, 1.2, 1.2.3). Anything with
+# a pre-, post-, or dev-release suffix is rejected: this tool is run *after*
+# a release to kick off the next dev cycle, so the starting point must be a
+# clean release version.
+RELEASE_VERSION_RE = re.compile(r"^(\d+)\.(\d+)(?:\.\d+)?$")
+
+
+def next_dev_versions(current: str) -> tuple[str, str]:
+    match = RELEASE_VERSION_RE.match(current.strip())
```
danieldk: Better have the function take

https://packaging.pypa.io/en/stable/version.html#packaging.version.Version

and do the work on a nice, parsed representation.

scripts/pre_release.py
```diff
@@ -0,0 +1,105 @@
...
+    replace_top_level_version,
+)
+
+# Accept PEP 440 dev versions like `0.14.0.dev0` (the form kernels/pyproject.toml
+# uses). Cargo.toml files use `0.14.0-dev0`, but we never read from them — we
+# only write to them, using the release string we derive from pyproject.
+DEV_VERSION_RE = re.compile(r"^(\d+)\.(\d+)(?:\.(\d+))?\.dev\d+$")
+
+
+def next_release_version(current: str) -> str:
```
danieldk: Pass a `Version` and get the bits from there. Can also be serialized back to a string.

Makefile
```diff
@@ -24,3 +24,25 @@ kernel-builder-cli-docs:
...
 	ruff format --check ${check_dirs}
 	ruff check ${check_dirs}
+
+# Bump every version site to the next dev release based on the currently
+# installed `kernels` package version (e.g. installed 0.13.0 -> 0.14.0.dev0).
+# Refreshes Cargo.lock and kernels/uv.lock so all sites stay consistent.
+bump-dev:
+	python scripts/bump_to_dev.py
+	cargo check --workspace
+	cd kernels && uv lock
```
danieldk, line 37: Just to verify (docs seem to agree): this only updates the lock for `kernels` right?

danieldk: If we add `uv` to the set of tools that are used, it should also be added to `nativeBuildInputs` of `flake.nix`.

scripts/_version_common.py
sayakpaul: Existing `tomli` or `tomlilib` doesn't seem to have a serialization module. So, what should we prefer in that case? Use something like `tomlkit`?

sayakpaul:
@danieldk should be good now. PTAL.

docs/source/api/kernels.md
```diff
@@ -26,4 +26,4 @@
 
 ### get_locked_kernel
 
-[[autodoc]] kernels.get_locked_kernel
+[[autodoc]] kernels.get_locked_kernel
```
sayakpaul, line 29: I don't understand why it's here.

scripts/_version_common.py
```diff
@@ -0,0 +1,95 @@
...
+
+def replace_top_level_version(path: Path, new_version: str, *, dry_run: bool) -> str | None:
+    old_version = _parsed_version(path)
+    if old_version == new_version:
+        return None
+
+    # tomllib/tomli are parse-only and we don't want a write-capable TOML dep, so
+    # we rewrite the version line ourselves. Anchoring to ``^`` plus the escaped
+    # old value (just parsed) makes this an unambiguous single match — inline dep
+    # entries like `clap = { version = "4", ... }` can't collide.
```
danieldk: Sorry, I completely forgot about this. We are already including `tomlkit` as a dependency:

https://github.com/huggingface/kernels/blob/1e09d03ee4c30995c2f35762fb9e1e61670eb608/kernels/pyproject.toml#L17

It can write and, best thing, it preserves the layout of the original TOML file:

https://tomlkit.readthedocs.io/en/latest/

I forget why we originally added it as a dependency (we were probably writing TOML files at some point in `kernels`).

sayakpaul:
@danieldk incorporated `tomlkit` for writing.

danieldk (approved):
Really nice!

PR #498, opened 2026-04-21
Add `make pin-actions` target to pin all GitHub actions

danieldk:
And also pin all the actions :).


sayakpaul (dismissed):
Wow! Today I learned `pinact`.

PR #499, opened 2026-04-22
kernels: use `Metadata` from `kernels-data`

danieldk:
This changeset combines a few related changes:

- Use `Metadata` from `kernels-data` to ensure using the same metadata data structures in kernel-builder and kernels.
- Make metadata mandatory and make several metadata fields mandatory, namely `id`, `version`, `license` the new `name` field.
- Update the `build.toml` format to v4, making the required fields mandatory as well as the `torch-noarch` section for torch-noarch kernels.
- Remove support for the very old v1 and v2 `build.toml` formats.
- In `LoadedKernel`, add metadata and remove `module_name` (can be queried from the metadata).
- Rename `RepoInfos` to `RepoInfo`.


nix-builder/versions.nix
```diff
@@ -1,14 +1,4 @@
 [
```
danieldk, line 1: @adarshxs for sglang, it's better to change their flake to add `torchVersions` with this variant. The old variant is causing some issues in automatic version selection for devshells, etc.

adarshxs: got it will do

drbh (approved):
LGTM

PR #500, opened 2026-04-22
feat: include layer usage examples in the card.

### LINKED ISSUE #486: Card generation incorrectly shows `layers` as a function
`layers` is a module with kernel layers (if present), however, card generation currently adds `layer` as a function:

Image

This also leads to incorrect usage fragments. If `layers` is present, we should make a separate section _Available layers` with the names of the layers. We should also not use `layers` in the usage example.

sayakpaul:
Fix https://github.com/huggingface/kernels/issues/486


danieldk (approved):
🔥🔥🔥

Looks great!

Issue #436, opened 2026-04-09
Consider removing variant enumeration again and instead explain why each available are not compatible


danieldk:
The issue with variant enumeration is that it is easy to get incomplete enumerations (which is one of the reasons we removed it from the core kernel logic). The (purely) informative enumeration #423 again suffers from this issue, because it does not take into account that we also support falling back to older CUDA versions. Similar issues may rise when we support compatibility with older ABI versions (e.g. Torch stable ABI?).

I think that we should instead focus on the possibility to provide an explanation why existing variants are rejected. This does not only suffer from the incomplete enumeration issue (and maintaining both compatibility checking and enumeration code paths), but is also more informative, because it makes it clear _why* none of the available variants can be used (which could guide the user towards e.g. using a different Torch version).

ArjunSrivastava1:
hmm, looks to be a rather interesting discussion/issue i've stumbled upon
maybe i can help add a few perspectives in some ways?

what ur trying to describe is of a profiling to be done for the kernels correct? so as a diagnostic can be prepared for the rejected variants at the time or before running

such a system exists in languages such as golang( gopprof) and rust - exceptions and details before the code works, the exceptions in rust are after all, famous for being verbose as they are: gopprof also makes entire diagrams and deploys them on a site too - its more used for prof current code and the functions in the code itself for analysis and optimisation work - but it still works on what the user already has got

as such, the question comes down to this for anything that will be made - is it more feasible to run the diagnostic module thats gonna be returning the reasons of rejections(assume one is made) at runtime? or compile time? which is easier to complete with what we already have?

i say inspo can be taken from rusts system as it will be easier to leverage runtime compatibility and we are just, working with rejections of existing variants

sayakpaul:
I don't think there's any need for profiling. We have thorough metadata, which is queryable through our Python API, which we can use for validation purposes.

ArjunSrivastava1:
good, thats one issue down that we dont have to look further into - validations and reasons can come from there then

step one: use the metadata and api like u said for validation
step two: ???
step three: have a working thing that gives us reasons for rejections for existing variants or why they were incompatible

now, how do we go forward then? theres also quite a bit of latency involved in for querying via python api too, i suppose its not a concern for query to be sluggish in anyway then

that leaves us with the step of putting together the data with a framework that lists out validations and stuff for a more informative format that can be used

sayakpaul:
The latency would be negligible.

ArjunSrivastava1:
[...]

PR #504, opened 2026-04-24
Improved error message for unavailable variant

theely:
If a kernel is not available, the raised exception now also includes the name of the required variant.
This change should help users to identify what variant they need for their system.

### DIFF
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -201,11 +201,11 @@ def install_kernel(
     package_name = package_name_from_repo_id(repo_id)
 
     variants = get_variants(api, repo_id=repo_id, revision=revision)
-    variant = resolve_variant(variants, backend)
-
-    if variant is None:
+    try:
+        variant = resolve_variant(variants, backend)
+    except FileNotFoundError as e:
         raise FileNotFoundError(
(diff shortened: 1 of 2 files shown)


danieldk (commented):
Thanks for contributing to kernels!

Enumeration of variants is always going to be incomplete, see: https://github.com/huggingface/kernels/pull/422 E.g., in the PR

```
        missing_variant.filename = f"torch{torch_version.major}{torch_version.minor}-{'cxx11' if torch_cxx11_abi else 'cxx98'}-cu{selected_backend.version.major}{selected_backend.version.minor}-{cpu}-{os}"
```

Misses: tvm-ffi, the variant without the ABI tag (which is also supported), backwards CUDA compatibility (e.g. if the system has CUDA 12.8, CUDA 12.1 to 12.7 are also supported), etc. In this case it would also report the a variant for `cu` if the backend is ROCm, XPU, NPU, CPU, Metal, etc.

We discussed this a while back and came to the conclusion that we are instead going to add support for 'explaining', where for every available variant it's possible to get/list the reason why it is not selected.

sayakpaul:
Being tracked in https://github.com/huggingface/kernels/issues/436

danieldk:
Fixed by https://github.com/huggingface/kernels/pull/561

PR #505, opened 2026-04-24
Add support for Torch 2.12

danieldk:


nix-builder/pkgs/python-modules/nvidia-cutlass-dsl-libs/default.nix
sayakpaul, line 78: Why is this being surfaced now? 👀

nix-builder/pkgs/python-modules/nvidia-cutlass-dsl/default.nix
sayakpaul, line 16: What is happening here?

nix-builder/pkgs/python-modules/torch/binary/torch-versions-hash.json
```diff
@@ -150,5 +150,73 @@
         "version": "2.11.0"
       }
     }
+  },
+  "2.12": {
```
sayakpaul, line 154: TODO: update.

nix-builder/pkgs/python-modules/triton/default.nix
```diff
@@ -0,0 +1,26 @@
+{
```
sayakpaul, line 1: Really cool!

sayakpaul (commented):
Very very cool work on getting the Triton CUDA binary wheels in the build.

nix-builder/pkgs/python-modules/nvidia-cutlass-dsl-libs/default.nix
danieldk: We didn't do cache builds for macOS since the cutlass-dsl update. We cannot really do cache builds in CI, because the macOS CI runners are impure and we do not want to push output paths from impure builds to the binary cache.

drbh (dismissed):
lgtm!

drbh (dismissed):

PR #506, opened 2026-04-26
add tvm-ffi metal 

### LINKED ISSUE #352: tvm-ffi Metal support

ved1beta:
fixes #352

## Summary
- Adds noarch variants for the tvm-ffi framework (`tvm-ffi-metal`, `tvm-ffi-cuda`, `tvm-ffi-cpu`, `tvm-ffi-rocm`, `tvm-ffi-xpu`, `tvm-ffi-universal`)
- Introduces `TvmFfiNoarch` mirroring `TorchNoarch`;
- Extends `parse_variant` and `BUILD_VARIANT_REGEX`
- Updates `_sort_variants` so within noarch, Torch is preferred over tvm-ffi 

### tests 
Torch beats tvm-ffi when both are present
noarch fallback when no arch variant matches
arch variant beats noarch on the actual hardware

### DIFF
--- a/kernels/src/kernels/variants.py
+++ b/kernels/src/kernels/variants.py
@@ -23,7 +23,12 @@ from kernels.backends import (
 )
 from kernels.compat import has_torch, has_tvm_ffi
 
-BUILD_VARIANT_REGEX = re.compile(r"^(torch\d+\d+|torch-(cpu|cuda|metal|neuron|rocm|xpu)|tvm-ffi\d+\d+)")
+BUILD_VARIANT_REGEX = re.compile(
+    r"^(torch\d+\d+"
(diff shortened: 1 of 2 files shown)


danieldk:
We already have a branch for tvm-ffi and there is also support for kernel-builder in it. However Metal support will require upstream tvm-ffi changes, because there is currently no way for tvm-ffi to pass `MTLBuffer`s (only pointers to the raw data), which is needed for other Metal APIs (as well as no way to get other compute context).

For this reason we are postponing tvm-ffi Metal support until these issue have been fixed upstream.

ved1beta:
ahh thanks for the response , will try to put a pr upstream :saluting_face:

PR #507, opened 2026-04-27
Update `CARD.md` to include install tip when benchmarking

alvarobartt:
## Description

This PR updates the `CARD.md` content that goes into the Hugging Face Hub for the benchmarking section to include the installation instruction with `pip` including the `benchmark` extra.

Note that this might be opinionated, so feel free to close if not relevant, I just found that useful when running the benchmarks as I had to `pip install kernels` -> `kernels benchmark ...` -> `Error: Missing dependencies for benchmark: tabulate\nInstall with: pip install 'kernels[benchmark]'` -> `pip install "kernels[benchmark]"`.

Thanks in advance! cc @sayakpaul @danieldk

### DIFF
--- a/examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
+++ b/examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
@@ -42,7 +42,7 @@ Function list not available.
 ## Benchmarks
 {% if has_benchmark %}
 
-Benchmarking script is available for this kernel. Run `kernels benchmark {{ repo_id }}`.
+Benchmarking script is available for this kernel. Install as `pip install -U "kernels[benchmark]"` and run `kernels benchmark {{ repo_id }}`.
 {% else %}
 
 No benchmark available yet.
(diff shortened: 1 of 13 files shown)


sayakpaul:
Oh it should actually be included in the card template so that we don't have to manually update the cards. The template: https://github.com/huggingface/kernels/blob/main/kernel-builder/src/init/templates/CARD.md

sayakpaul:
Cc: @drbh

danieldk (approved):

danieldk:
> Oh it should actually be included in the card template so that we don't have to manually update the cards. The template: https://github.com/huggingface/kernels/blob/main/kernel-builder/src/init/templates/CARD.md

One of the changes is in the init template, but requires regen of the card templates in the kernel repos.

alvarobartt:
> One of the changes is in the init template, but requires regen of the card templates in the kernel repos.

Sure, I can just create a quick script to do so on the kernels at https://huggingface.co/kernels-community/kernels programmatically opening PRs in there?

PR #509, opened 2026-04-29
Add issue and pr templates

drbh:
This PR adds issue and pr templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,55 @@
+name: Bug Report
+description: Report a bug or unexpected behavior
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,55 @@
...
+    attributes:
+      label: Environment
+      description: Please provide version info.
+      placeholder: |
+        - kernels version:
+        - Python version:
+        - PyTorch version:
+        - CUDA version:
+        - GPU model:
+        - OS:
```
danieldk, line 46: Maybe if it concerns the builder also adding the output of `nix-shell -p nix-info --run "nix-info -m"`

danieldk (dismissed):
Looks great, added one small idea for the bug report type.

.github/ISSUE_TEMPLATE/bug_report.yml
drbh: oh good point, added in latest change

PR #511, opened 2026-04-29
feat: prefer using hf hub library

drbh:
This PR updates the kernel-builder to use the hf-hub repo based on the https://github.com/huggingface/hf-hub/pull/150 branch that ports the huggingface_hub_rust functionality to the existing hf-hub library

this PR will be updated to use the crate based distribution when the upstream is merged and published (will move out of draft at that point)


drbh:
optimistically merging

PR #512, opened 2026-04-29
feat: add trusted orgs and flag

drbh:
This PR adds a `trust_remote_code` and hardcoded list of trusted orgs

### DIFF
--- a/kernels/src/kernels/__init__.py
+++ b/kernels/src/kernels/__init__.py
@@ -24,6 +24,7 @@ from kernels.layer import (
     use_kernel_mapping,
 )
 from kernels.utils import (
+    TRUSTED_KERNEL_ORGS,
     LoadedKernel,
     RepoInfo,
     get_kernel,
@@ -52,6 +53,7 @@ __all__ = [
     "Metadata",
     "Mode",
     "RepoInfo",
+    "TRUSTED_KERNEL_ORGS",
     "get_kernel",
     "get_loaded_kernels",
     "get_local_kernel",
--- a/kernels/src/kernels/__init__.py
+++ b/kernels/src/kernels/__init__.py
@@ -24,7 +24,6 @@ from kernels.layer import (
     use_kernel_mapping,
 )
 from kernels.utils import (
-    TRUSTED_KERNEL_ORGS,
     LoadedKernel,
     RepoInfo,
     get_kernel,
@@ -53,7 +52,6 @@ __all__ = [
     "Metadata",
     "Mode",
     "RepoInfo",
-    "TRUSTED_KERNEL_ORGS",
     "get_kernel",
     "get_loaded_kernels",
     "get_local_kernel",
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -31,6 +31,8 @@ from kernels.variants import (
 
 KNOWN_BACKENDS = {"cpu", "cuda", "metal", "neuron", "rocm", "xpu", "npu"}
 
+_ALWAYS_TRUSTED_ORGS = {"kernels-community", "kernels-staging", "kernels-test", "sglang"}
+
 
 def _check_trust_remote_code(repo_id: str, trust_remote_code: bool | list[str]) -> None:
     """Check whether a kernel repository is trusted.
@@ -59,6 +61,10 @@ def _check_trust_remote_code(repo_id: str, trust_remote_code: bool | list[str]) 
             "to bypass trust checks.",
             stacklevel=3,
         )
+
+    org = repo_id.split("/", 1)[0]
+    if org in _ALWAYS_TRUSTED_ORGS:
+        return
 
     api = _get_hf_api()
     try:
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -69,12 +69,11 @@ def _check_trust_remote_code(repo_id: str, trust_remote_code: bool | list[str]) 
     api = _get_hf_api()
     try:
         info = api.repo_info(repo_id, repo_type="kernel")
-    except Exception:
+    except Exception as e:
         raise ValueError(
-            f"Kernel repository '{repo_id}' is not from a trusted publisher. "
-            f"Could not verify publisher trust status. "
-            f"Set trust_remote_code=True to allow loading kernels from untrusted sources."
-        )
+            f"Could not verify publisher trust status for kernel repository '{repo_id}'. "
+            "Set trust_remote_code=True to allow loading kernels from untrusted sources."
+        ) from e
 
     if not getattr(info, "trustedPublisher", False):
         raise ValueError(
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -31,22 +31,20 @@ from kernels.variants import (
 
 KNOWN_BACKENDS = {"cpu", "cuda", "metal", "neuron", "rocm", "xpu", "npu"}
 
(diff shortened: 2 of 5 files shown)


kernels/src/kernels/utils.py
```diff
@@ -31,6 +31,45 @@
 
 KNOWN_BACKENDS = {"cpu", "cuda", "metal", "neuron", "rocm", "xpu", "npu"}
 
+TRUSTED_KERNEL_ORGS = {"kernels-community", "kernels-test", "sglang"}
```
sayakpaul:
```suggestion
TRUSTED_KERNEL_ORGS = {"kernels-community", "kernels-staging", "kernels-test", "sglang"}
```

sayakpaul (dismissed):

kernels/src/kernels/__init__.py
```diff
@@ -24,6 +24,7 @@
     use_kernel_mapping,
 )
 from kernels.utils import (
+    TRUSTED_KERNEL_ORGS,
```
danieldk: Do we want to expose this? Seems like an internal. Also, what it means would break once we switch to signing identities.

kernels/tests/test_basic.py
```diff
@@ -233,6 +233,44 @@ def test_neuron():
...
+
+    This only checks that the trust check passes — it may still fail
+    for other reasons (e.g. network, missing variant).
+    """
+    # kernels-community is in TRUSTED_KERNEL_ORGS, so trust check passes.
+    # The call may fail later (no CUDA, etc.), but not with a trust error.
+    try:
+        get_kernel("kernels-community/relu", version=1)
+    except ValueError as e:
+        assert "not from a trusted organisation" not in str(e)
```
danieldk:
```suggestion
    get_kernel("kernels-community/relu", version=1)
```

I don't think we should make other errors disappear.

danieldk: Though I think the test is a bit superfluous, since most other tests test this already.

kernels/tests/test_basic.py
```diff
@@ -233,6 +233,44 @@ def test_neuron():
...
+
+
+def test_trust_remote_code_flag_allows_untrusted():
+    """trust_remote_code=True should bypass the org check."""
+    # This will likely fail for network/variant reasons, but NOT with trust error.
+    try:
+        get_kernel("some-random-org/some-kernel", version=1, trust_remote_code=True)
```
danieldk: This leads to remote compromise of our CI if someone creates `some-random-org`.

We should make an org that is owned by us and put a noarch kernel in it.

kernels/src/kernels/utils.py
julien-c: please rely on the hub api for this, so it's consistent with the UI:

image

julien-c: (cc @krampstudio)

sayakpaul:
```
{"kernels-community", "kernels-staging", "kernels-test", "sglang"}
```

Can we ensure the above list is consistent with the Hub?

kernels/tests/test_basic.py
drbh: good point, updated to point at personal kernel repos and uploaded a placeholder

kernels/src/kernels/utils.py
drbh: thanks for the feedback, i've updated `_check_trust_remote_code` to check the repo_info response for the trustedPublisher attribute.

full logic is:
1. if `trust_remote_code=True`, skip all checks
2. check the hardcoded list (just "kernels-test" for now)
3. fetch the repo_info and check attribute

krampstudio: > ```
> {"kernels-community", "kernels-staging", "kernels-test", "sglang"}
> ```
> 
> Can we ensure the above list is consistent with the Hub?

Here is the list configured on the Hub:

Screenshot 2026-05-02 at 15 27 58

kernels/src/kernels/utils.py
```diff
@@ -31,6 +31,57 @@
...
+
+    api = _get_hf_api()
+    try:
+        info = api.repo_info(repo_id, repo_type="kernel")
+    except Exception:
+        raise ValueError(
+            f"Kernel repository '{repo_id}' is not from a trusted publisher. "
+            f"Could not verify publisher trust status. "
+            f"Set trust_remote_code=True to allow loading kernels from untrusted sources."
+        )
```
danieldk: I would remove the first line. It could be from a trusted publisher, but the request is failing for some other reason. I think the original exception should also be propagated, since it is useful to see the underlying issue. So:


```suggestion
    except Exception as e:
        raise ValueError(
            f"Could not verify publisher trust status for kernel repository '{repo_id}'. "
            "Set trust_remote_code=True to allow loading kernels from untrusted sources."
        ) from e
```

kernels/src/kernels/utils.py
```diff
@@ -31,6 +31,57 @@
...
+        warnings.warn(
+            "Signing identity verification is not yet implemented. "
+            "The provided signing identities will be ignored and the "
+            "kernel will be treated as untrusted. Use trust_remote_code=True "
+            "to bypass trust checks.",
+            stacklevel=3,
+        )
+
+    org = repo_id.split("/", 1)[0]
+    if org in _ALWAYS_TRUSTED_ORGS:
```
danieldk, line 66: Do we need this, since `kernels-test` is already a trusted org. That way, this would get properly tested during tests as well.

kernels/tests/test_basic.py
```diff
@@ -233,6 +233,22 @@ def test_neuron():
     torch.testing.assert_close(relu.relu(x), x.relu())
 
 
+def test_trust_remote_code_blocks_untrusted_org():
+    """Kernels from untrusted orgs should be rejected by default."""
+    with pytest.raises(ValueError, match=r"not from a trusted publisher"):
+        get_kernel("drbh/not-a-trused-org-kernel", version=1)
```
danieldk: We should make an extra org, something like `kernels-test-untrusted` and add the `kernels-test` maintainers to that org.

kernels/tests/test_basic.py
```diff
@@ -233,6 +233,22 @@ def test_neuron():
...
+
+
+def test_trust_remote_code_allows_trusted_org():
+    """Kernels from trusted orgs should not raise a trust error."""
+    get_kernel("kernels-community/relu", version=1)
+
+
+def test_trust_remote_code_flag_allows_untrusted():
+    """trust_remote_code=True should bypass the org check."""
+    get_kernel("drbh/ci-test-kernel", version=1, trust_remote_code=True)
```
danieldk: Same here.

kernels/src/kernels/utils.py
drbh: nope this is unneeded now (just ran into some CI issues without this special case earlier). it looks like `kernels-test` are returning the trustedPublisher attribute now after some spot checks. 

removed in latest

kernels/tests/test_basic.py
drbh: good point! just created and added you as an admin https://huggingface.co/kernels-test-untrusted

kernels/tests/test_basic.py
drbh: updated this too

kernels/src/kernels/utils.py
drbh: ahh the issue is now we surface a untrusted publisher if the repo info cannot be fetched since the check depends on the repo info. need to revisit with a better solution

danieldk (approved):

PR #516, opened 2026-05-04
kernel-builder: fix failed uploads trying to remove directories

danieldk:
When trying to collect candidates for stale entries, we included directories as well. This causes upload failures, since directories are not Git objects and the API would reject these.


drbh (approved):
LGTM, thanks for fixing! 

this issue was introduced when moving to the hf hub library in https://github.com/huggingface/kernels/pull/511 (specifically the api change from list_files to list_tree resulted in dirs being included in the list)

PR #519, opened 2026-05-04
fix: remove existing test repo before upload

drbh:
This PR deletes the existing test repo from staging before upload so we force repo creation each time, this should also fix issues where staging repos have refs removed but keep the repo which causes upload to fail on the `/refs` call

### DIFF
--- a/.github/workflows/test_e2e.yaml
+++ b/.github/workflows/test_e2e.yaml
@@ -48,14 +48,22 @@ jobs:
       # Remove the existing test repo if it exists so we test repo creation in the upload step
       - name: Delete existing test repo
         run: |
-          pip install huggingface_hub
-          python -c "
+          cat > /tmp/delete_repos.py << 'PYEOF'
           from huggingface_hub import HfApi
-          api = HfApi(endpoint='${{ env.HF_ENDPOINT }}', token='${{ secrets.HF_STAGING_TOKEN }}')
-          for repo_type in ['kernel', 'model']:
-              api.delete_repo('${{ env.E2E_REPO_ID }}', repo_type=repo_type, missing_ok=True)
-              print(f'Deleted {repo_type} repo (or it did not exist)')
-          "
+          import os
+
+          api = HfApi(
+              endpoint=os.environ["HF_ENDPOINT"],
+              token=os.environ["HF_STAGING_TOKEN"],
+          )
+          repo_id = os.environ["E2E_REPO_ID"]
+          for repo_type in ["kernel", "model"]:
+              api.delete_repo(repo_id, repo_type=repo_type, missing_ok=True)
+              print(f"Deleted {repo_type} repo (or it did not exist)")
+          PYEOF
+          nix-shell -p python3Packages.huggingface-hub --run "python /tmp/delete_repos.py"
+        env:
+          HF_STAGING_TOKEN: ${{ secrets.HF_STAGING_TOKEN }}
 
       - name: Init kernel project
         run: |
--- a/.github/workflows/test_e2e.yaml
+++ b/.github/workflows/test_e2e.yaml
@@ -148,7 +148,7 @@ jobs:
           import torch
           from kernels import get_kernel
 
-          kernel = get_kernel('${{ env.E2E_REPO_ID }}', revision='${{ env.E2E_BRANCH }}')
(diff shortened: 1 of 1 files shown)


.github/workflows/test_e2e.yaml
```diff
@@ -20,7 +20,7 @@ env:
   E2E_BRANCH: e2e-${{ github.event.pull_request.number || github.run_id }}-${{ github.run_attempt }}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
sayakpaul, line 23: Could you explain this change?

.github/workflows/test_e2e.yaml
```diff
@@ -45,6 +45,23 @@ jobs:
         env:
           USER: runner
 
+      # Remove the existing test repo if it exists so we test repo creation in the upload step
+      - name: Delete existing test repo
+        run: |
+          cat > /tmp/delete_repos.py << 'PYEOF'
+          from huggingface_hub import HfApi
```
sayakpaul, line 52: Can't we just use `delete_repo()` without having to initialize the API?

.github/workflows/test_e2e.yaml
```diff
@@ -45,6 +45,23 @@ jobs:
         env:
           USER: runner
 
+      # Remove the existing test repo if it exists so we test repo creation in the upload step
+      - name: Delete existing test repo
+        run: |
+          cat > /tmp/delete_repos.py << 'PYEOF'
```
sayakpaul, line 51: Why not keep it under `scripts` or `workflows` directory?

.github/workflows/test_e2e.yaml
```diff
@@ -121,6 +138,8 @@ jobs:
           CUDA_TAG=$(echo "$VARIANT" | grep -oP 'cu\d+')
           echo "Installing torch matching variant $VARIANT (CUDA tag: $CUDA_TAG)"
           uv sync --all-extras --dev
+          # E2E validates this checkout, so use the local kernels-data binding instead of the released wheel from uv.lock.
```
sayakpaul, line 141: Why is this needed?

sayakpaul (commented):
Thanks! Left some comments.

.github/workflows/test_e2e.yaml
drbh: this change updates the `concurrency` group to be scoped to this workflow instead of the workflow + the pr.

this way the actions should only run one time per workflow, and not per pr, so in the case multiple PR's are open, it should only run this workflow one at a time to avoid race conditions. 

since this change adds a step to delete the staging repo, if the workflows ran in parallel then it could hypothetically have a race condition if one action deleted the repo after the other action uploaded and before it downloaded in the `get_kernel` call. 

so this change should guard against that race condition

.github/workflows/test_e2e.yaml
drbh: oh I wasn't aware that was possible? can you share a snippet we can use to delete a repo in a more concise way? I was under the impression we'd need to create the api object pointing to the staging url. 

Happy to make changes to a better approach!

.github/workflows/test_e2e.yaml
drbh: we can move it if you think thats a better location - however since the snippet is so small and only really expected to be used in this case I thought it made sense to inline.

lmk what you think!

.github/workflows/test_e2e.yaml
drbh: this was added to ensure the E2E test is using the latest kernel-data in the repo, instead of the version from the lock file. 

I ran into an issue with a difference in the api used to load the metadata in a previous action run https://github.com/huggingface/kernels/actions/runs/25754879907/job/75640986832#step:5:36

and by reinstalling the kernels-data library after syncing, it make sure to use the latest changes (and not the one pinned in the uv.lock)

.github/workflows/test_e2e.yaml
sayakpaul: https://github.com/huggingface/diffusers/blob/6abf75263a09a3e7a62458f544ce2fac28568fe2/tests/models/test_modeling_common.py#L2046

Docs: https://huggingface.co/docs/huggingface_hub/v1.12.1/en/package_reference/hf_api#huggingface_hub.HfApi.delete_repo

sayakpaul (approved):
Thanks for the answers!

PR #522, opened 2026-05-06
kernel-builder: raise C++ standard level to C++20

danieldk:
## What does this PR do?

Raise C++ version to C++ 20.

## Motivation

Torch 2.12 on macOS requires at least C++20.


adarshxs (approved):
LGTM

PR #525, opened 2026-05-06
nix-builder: support building CI per framework

danieldk:
Support building CI per framework to support per-backend CI in kernels-community.


drbh (approved):
lgtm!

Issue #538, opened 2026-05-11
Update to check trusted repos from hub


drbh:

danieldk:
I think this issue can be closed?

PR #539, opened 2026-05-12
feat: prefer using hub api to check if publisher is trusted

drbh:
This PR prefers checking if a publisher is trusted from the hub api response

depends on: https://github.com/huggingface-internal/moon-landing/pull/18061
related to: https://github.com/huggingface/kernels/issues/538

CI may need to be rerun after the latest backend deployment

### DIFF
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -66,17 +66,16 @@ def _check_trust_remote_code(repo_id: str, trust_remote_code: bool | list[str]) 
     try:
         info = _get_hf_api().get_organization_overview(publisher)
     except Exception:
-        failure_reason = "could not verify publisher trust status"
-    else:
-        if getattr(info, "trustedKernelPublisher", False) is True:
-            # If the response indicates that the publisher is trusted we allow the
-            # kernel to be loaded.
-            return
-        # otherwise we fail with a message stating that the publisher is not trusted.
-        failure_reason = "is not from a trusted publisher"
+        raise ValueError(
+            f"Kernel repository '{repo_id}' could not verify publisher trust status. "
+            "Set trust_remote_code=True to allow loading kernels from untrusted sources."
+        )
+
+    if getattr(info, "trustedKernelPublisher", False):
+        return
 
     raise ValueError(
-        f"Kernel repository '{repo_id}' {failure_reason}. "
+        f"Kernel repository '{repo_id}' is not from a trusted publisher. "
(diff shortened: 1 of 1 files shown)


wadeKeith (commented):
Withdrawn; this review was too cursory.

kernels/src/kernels/utils.py
```diff
@@ -62,33 +61,25 @@ def _check_trust_remote_code(repo_id: str, trust_remote_code: bool | list[str])
...
-    if org in _ALWAYS_TRUSTED_ORGS:
-        return
+    publisher = repo_id.split("/", 1)[0]
+
+    try:
+        info = _get_hf_api().get_organization_overview(publisher)
+    except Exception:
+        failure_reason = "could not verify publisher trust status"
+    else:
+        if getattr(info, "trustedKernelPublisher", False) is True:
```
danieldk:
```suggestion
        if getattr(info, "trustedKernelPublisher", False):
```

kernels/src/kernels/utils.py
```diff
@@ -62,33 +61,25 @@ def _check_trust_remote_code(repo_id: str, trust_remote_code: bool | list[str])
...
 
-    org = repo_id.split("/", 1)[0]
-    if org in _ALWAYS_TRUSTED_ORGS:
-        return
+    publisher = repo_id.split("/", 1)[0]
+
+    try:
+        info = _get_hf_api().get_organization_overview(publisher)
+    except Exception:
+        failure_reason = "could not verify publisher trust status"
```
danieldk: I think it's better to have a raise here as immediately. Raising once we know we can solves future logic bugs when this function becomes more complicated.

drbh: sounds good to me this makes sense. updated in latest

kernels/src/kernels/utils.py
drbh: good catch, updated in latest too

sayakpaul (approved):
Any test needs to be updated?

drbh:
> Any test needs to be updated? 

I don't think so, the `trustedPublishers` returned from the api should match the ones that were previously hardcoded so the tests should pass. 

CI looks green (aside from the E2E test which should be resolved in https://github.com/huggingface/kernels/pull/519) so I think the tests may be good as it

**will leaved unmerged for now, but if this looks good please feel free to merge! Otherwise I'll address any issues and merge tomorrow in my morning

PR #541, opened 2026-05-12
feat: prefer aotrion from release

drbh:
This PR avoids compiling aotrion and prefers using the prebuilt `libaotriton_v2.so` from https://github.com/ROCm/aotriton/releases

example of contents from
https://github.com/ROCm/aotriton/releases/download/0.11b/aotriton-0.11b-manylinux_2_28_x86_64-rocm6.2-shared.tar.gz

```bash
├── include
│   └── aotriton
│       ├── _internal
│       │   ├── aiter_hip_common.h
│       │   ├── flash
│       │   │   └── aiter.h
│       │   ├── kernel_cluster.h
│       │   ├── lazy_tensor_internal.h
│       │   ├── packed_kernel.h
│       │   ├── triton_kernel.h
│       │   └── util.h
│       ├── config.h
│       ├── cpp_tune.h
│       ├── dtypes.h
│       ├── flash.h
│       ├── runtime.h
│       ├── util.h
│       └── v2
│           └── flash.h
└── lib
    ├── cmake
    │   └── aotriton
    │       ├── aotriton-config-version.cmake
    │       ├── aotriton-config.cmake
    │       ├── aotriton-targets-release.cmake
    │       └── aotriton-targets.cmake
    ├── libaotriton_v2.so -> libaotriton_v2.so.0.11.2
    └── libaotriton_v2.so.0.11.2
```

### DIFF
--- a/nix-builder/pkgs/aotriton/generic.nix
+++ b/nix-builder/pkgs/aotriton/generic.nix
@@ -3,7 +3,6 @@
   clr,
   fetchurl,
   lib,
-  python3,
   rocm-core,
   stdenv,
   xz,
@@ -31,8 +30,6 @@ stdenv.mkDerivation {
   };
 
   nativeBuildInputs = [ autoPatchelfHook ];
-  nativeInstallCheckInputs = [ python3 ];
-
   buildInputs = [
     clr
     stdenv.cc.cc.lib
@@ -42,8 +39,6 @@ stdenv.mkDerivation {
   dontConfigure = true;
   dontBuild = true;
   dontStrip = true;
-  doInstallCheck = true;
-
   installPhase = ''
     runHook preInstall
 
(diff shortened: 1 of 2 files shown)


nix-builder/pkgs/aotriton/generic.nix
```diff
@@ -1,137 +1,82 @@
...
+    import os
+
+    ctypes.CDLL(
+        "$out/lib/libaotriton_v2.so",
+        mode=os.RTLD_NOW | os.RTLD_LOCAL,
+    )
+    PY
+
+    runHook postInstallCheck
+  '';
```
danieldk: Hmmm, I don't think this adds much to all the fixup stuff that the build already does.

If we want to test this, it's probably best to make passthru tests, one to verify that Torch loads with aotrition. Second that actually does some computation that triggers aotriton use (but we'd need a ROCm GPU to actually run it).

drbh: thanks for the suggestion, removed the installCheckPhase in latest commit. do you think it makes sense to add the passthru tests somewhere?

danieldk: As long as we cannot run it on the GPU, it does not seem that useful to have the passthru test, since loadability is already checked when ROCm Torch gets built:

```
❯ ldd result/torch210-cxx11-rocm71-x86_64-linux/torch-out/lib/python3.13/site-packages/torch/_C.cpython-313-x86_64-linux-gnu.so | grep ao                                                                                            
        libaotriton_v2.so.0.11.1 => /nix/store/k85zxmmnjfyhji6b5q6y9yhs2i4f1ajq-rocm-merged/lib/libaotriton_v2.so.0.11.1 (0x00007260c4400000)
❯ grep "pythonImportsCheck" nix-builder/pkgs/python-modules/torch/binary/generic.nix                                                                                                                                                 
  pythonImportsCheck = [ "torch" ];
```

danieldk (approved):
Looks great!

PR #544, opened 2026-05-12
feat: enforce that a version must be specified

drbh:
This PR enforces that a version or revision must be specified in all cases

note: this PR also updates the examples cards to render correctly; updates the `__all__`s for some functions, adds a small change for the cli to work on tvm-ffi kernels, and regenerates the cards since there were template placeholders previously. I added these changes to this PR since I updated the card template to include the version and thought the examples should reflect.

### DIFF
--- a/examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
+++ b/examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
@@ -1,25 +1,56 @@
 ---
 library_name: kernels
-license: apache-2.0
----
+{% if license %}license: {{ license }}
+{% endif %}---
 
-This is the repository card of kernels-test/cutlass-gemm-tvm-ffi that has been pushed on the Hub. It was built to be used with the [`kernels` library](https://github.com/huggingface/kernels). This card was automatically generated.
+This is the repository card of {{ repo_id }} that has been pushed on the Hub. It was built to be used with the [`kernels` library](https://github.com/huggingface/kernels). This card was automatically generated.
 
 ## How to use
+{% if functions %}
 
 ```python
 # make sure `kernels` is installed: `pip install -U kernels`
 from kernels import get_kernel
 
-kernel_module = get_kernel("kernels-test/cutlass-gemm-tvm-ffi", version=1)
-cutlass_gemm = kernel_module.cutlass_gemm
+kernel_module = get_kernel("{{ repo_id }}", version={{ version }})
+{{ functions[0] }} = kernel_module.{{ functions[0] }}
 
-cutlass_gemm(...)
+{{ functions[0] }}(...)
 ```
+{% else %}
+
+Usage example not available.
+{% endif %}
 
 ## Available functions
-- `cutlass_gemm`
+{% if functions %}
+{% for func in functions %}
+- `{{ func }}`
+{% endfor %}
+{% else %}
+
+Function list not available.
+{% endif %}
+{% if layers %}
+
+## Available layers
+{% for layer in layers %}
+- `{{ layer }}`
+{% endfor %}
+{% endif %}
 
 ## Benchmarks
+{% if has_benchmark %}
+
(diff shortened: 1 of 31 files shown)


examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
```diff
@@ -1,56 +1,25 @@
 ---
 library_name: kernels
-{% if license %}license: {{ license }}
```
danieldk, line 3: Shouldn't the license be filled from the metadata, I don't think we want to hardcode it?

examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
```diff
@@ -1,56 +1,25 @@
...
 ## How to use
-{% if functions %}
 
 ```python
 # make sure `kernels` is installed: `pip install -U kernels`
 
-kernel_module = get_kernel("{{ repo_id }}")
-{{ functions[0] }} = kernel_module.{{ functions[0] }}
+kernel_module = get_kernel("kernels-test/cutlass-gemm-tvm-ffi", version=1)
```
danieldk: I am confused, this should be done by card filling, right? The card should be filled at upload time, not in the repo already.

examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
```diff
@@ -1,56 +1,25 @@
...
-Function list not available.
-{% endif %}
-{% if layers %}
-
-## Available layers
-{% for layer in layers %}
-- `{{ layer }}`
-{% endfor %}
-{% endif %}
+- `cutlass_gemm`
```
danieldk: Same here. Now the list of functions would never update as a kernel is extended.

kernels/src/kernels/_versions.py
```diff
@@ -59,14 +58,12 @@ def select_revision_or_version(
 
     if revision is not None:
         return revision
-    elif version is not None:
+
+    if version is not None:
```
danieldk: We can keep this an `elif` right? To indicate that this is a continuation of the previous condition.

kernels/src/kernels/utils.py
```diff
@@ -392,6 +393,13 @@ def get_kernel(
         result = activation.relu(out, x)
         ```
     """
+    if revision is None and version is None:
```
danieldk: Why do we need this here? `select_revision_or_version` gets called below. We should have one place for validation, not the same validation in multiple places.

kernels/src/kernels/layer/layer.py
```diff
@@ -71,6 +72,8 @@ def __init__(
     ):
         if revision is not None and version is not None:
             raise ValueError("Either a revision or a version must be specified, not both.")
+        if revision is None and version is None:
+            raise ValueError("Either a revision or a version must be specified.")
```
danieldk, line 77: Here we need the duplication because the fetching is only lazy (for good reasons).

danieldk (commented):
Something seems to have gone wrong for the cards? They should be templates, but this PR makes them filled.

examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
drbh: yea 100% I made the mistake of pushing hydrated cards 🤦‍♂️, fixed in latest changes

examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
drbh: same as above, should be resolved in latest changes

kernels/src/kernels/_versions.py
drbh: yes that makes more sense syntactically, updated to use elif in the latest changes

kernels/src/kernels/utils.py
drbh: good catch, removed in latest

danieldk (approved):
Nice!!!

PR #550, opened 2026-05-14
Update version bumping scripts with the `--major` option

danieldk:
With this change the script supports both major and minor version bumping. For example:

Codebase at `0.10.1.dev0`

```
  (none)          -> 0.10.1
  --major         -> 0.11.0
  --dev           -> 0.10.1.dev1
  --dev --major   -> 0.11.0.dev0
```

Codebase at `0.10.1`:

```
  (none)          -> 0.10.2
  --major         -> 0.11.0
  --dev           -> 0.10.2.dev0
```

These are the typical version bumping workflows within the project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,8 @@ quality:
 	ruff format --check ${check_dirs}
 	ruff check ${check_dirs}
 
-# Bump every version site to the next dev release based on the currently
-# installed `kernels` package version (e.g. installed 0.13.0 -> 0.14.0.dev0).
+# Bump every version site to the next patch dev release
+# (e.g. 0.10.1 -> 0.10.2.dev0, or 0.10.1.dev0 -> 0.10.1.dev1).
 # Refreshes Cargo.lock and kernels/uv.lock so all sites stay consistent.
 bump-dev:
 	python scripts/bump_version.py --dev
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -39,13 +39,35 @@ bump-dev:
 bump-dev-dry-run:
 	python scripts/bump_version.py --dev --dry-run
 
-# Strip the `.dev0` / `-dev0` suffix from every version site in prep for a
-# release (e.g. codebase 0.14.0.dev0 -> 0.14.0). Refreshes Cargo.lock and
-# kernels/uv.lock so all sites stay consistent.
+# Bump every version site to the next minor dev release
+# (e.g. 0.10.1 -> 0.11.0.dev0, or 0.10.1.dev0 -> 0.11.0.dev0).
+# Refreshes Cargo.lock and kernels/uv.lock so all sites stay consistent.
+bump-dev-major:
```
sayakpaul, line 45: Should the new commands be added here too?
https://github.com/huggingface/kernels/blob/6ca4e2f917fa2f0114f0076c99b50258196e63d4/Makefile#L1

sayakpaul (dismissed):

Makefile
danieldk: Thanks, missed that, added!

sayakpaul (approved):

PR #558, opened 2026-05-17
nix-builder: switch to manylinux_2_28, dynamically link libstdc++

danieldk:
Before this change, we were ensuring manylinux_2_28 compliance by:

* Building and linking against glibc 2.27 (there was no nixpkgs revision with 2.28).
* Linking libstdc++ statically.

Linking libstdc++ statically has turned out to be a pain point. If a kernel uses libstdc++ functionality that does some global initialization, the initialization of the dynamically-linked libstdc++ (e.g. through Torch) and the statically linked libstdc++ could conflict. One example is `std::regex`, which indirectly initializes some global code for locale handling.

As an alternative, I explored just linking libstdc++ dynamically, which worked up to some point by avoiding certain C++ features. However gcc 13 moved the initialization of the standard stream objects into the shared library. As a result, library compiled with gcc 13 or later requires a libstdc++ that is newer than manylinux_2_28.

The EL8 (RHEL, AlmaLinux, etc.) gcc toolset used in the manylinux_2_28 solves this by using EL8 libstdc++ and providing a static library that gets linked into a binary for newer C++ features. Since this is part of a fairly large patchset to gcc, it does not seem feasible to reproduce this easily in the Nix gcc derivations. So, instead, we repackage the gcc toolsets from AlmaLinux as Nix derivations in this change. This gives us a toolchain that is as close to the official manylinux_2_28 toolchain as possible.

The packaged toolchains are exposed as a standard nixpkgs stdenv, so that they can be used with other Nix derivations (such as the Torch/tvm-ffi extensions).

One exception is made in reproducing the toolchain and that is that we build glibc 2.28 ourselves. glibc carries the dynamic loader. The dynamic loader in AlmaLinux embeds FHS paths such as `/lib64`, which are not valid in Nix, and lead to linking errors since the library directory of e.g. `libc.so.6` cannot be found. By rebuilding glibc 2.28, we get a dynamic loader with the correct paths into the Nix store.


nix-builder/pkgs/manylinux/glibc_2_28/common.nix
```diff
@@ -0,0 +1,272 @@
+/*
```
danieldk, line 1: Everything in the `glibc_2_28` directory comes from Nix and does not need a review. Changes are:

* Bump the version to 2.28.
* Move `postFixup` outside the condition to remove a stray dependency on glibc 2.27.

nix-builder/pkgs/manylinux/default.nix
```diff
@@ -0,0 +1,40 @@
+{
```
danieldk, line 1: The JSON files in this directory are auto-generated from repomd data and do not need to be reviewed.

docs/source/builder/design-nix-builder.md
```diff
@@ -0,0 +1,106 @@
...
+
+We repackage the AlmaLinux 8 toolsets and libstdc++ as Nix derivations (see
+the `nix-builder/packages/manylinux_2_28` source directory). Then we merge
+various toolset packages to an unwrapped gcc that resembles unwrapped gcc in
+nixpkgs. Finally, we wrap binutils and gcc to combine them into a stdenv.
+
+The stdenv does not reuse glibc from AlmaLinux, since its dynamic loader has
+hardcoded FHS paths (`/lib64` etc.) that are not valid in Nix. Using this
+dynamic loader results in linking errors, since the paths in the dynamic
+loader are used as a last resort (to link glibc libraries). So, instead we
```
sayakpaul, line 46: Could be nice / helpful to point to the LOC/workflow that performs the build.

docs/source/builder/design-nix-builder.md
```diff
@@ -0,0 +1,106 @@
...
+build our own glibc 2.28 package and use that.
+
+## The package set pattern
+
+We repackage various existing package sets as Nix derivations. For instance,
```
sayakpaul, line 52: Could be nice to explain why the repackaging is helpful.

docs/source/builder/design-nix-builder.md
```diff
@@ -0,0 +1,106 @@
...
+packages in intermediate overlays.
+
+The package sets all use a similar list of overlays:
+
+- An initial overlay (`components.nix`) that applies a generic builder
+  to the package set metadata. The metadata typically comes from a Yum/DNF
+  repository that contains RPM packages.The generic builder will extract the
+  RPMs and move binaries, libraries, and headers to the right location. This
+  results in a set of Nix derivations that may or may not build.
+- The next overlay (`overrides.nix`) fixes up derivations created by the
```
sayakpaul: > created by the generic builder that do not build

Reads weird.

docs/source/builder/design-nix-builder.md
```diff
@@ -0,0 +1,106 @@
...
+  generic builder that do not build. Fixing the derivations typically consists
+  of adding missing dependencies and changing embedded FHS paths to Nix store
+  paths.
+- Additional overlays with derivations that combine outputs from previous
+  overlays. One typical example are derivations that construct a full compiler
+  toolchain.
```
sayakpaul: If we have an example, it would be nice to include.

nix-builder/pkgs/manylinux/rhel2nix.py
```diff
@@ -0,0 +1,439 @@
+#!/usr/bin/env python3
+
+import argparse
```
sayakpaul, line 3: When is this supposed to run?

sayakpaul (dismissed):
Just some nits.

nix-builder/pkgs/manylinux/rhel2nix.py
danieldk: Only by hand if we ever want to update the package set (unlikely) or need to add another manylinux version (e.g. AlmaLinux 9 for manylinux_2_34).

docs/source/builder/design-nix-builder.md
danieldk: Added.

docs/source/builder/design-nix-builder.md
danieldk: Added.

docs/source/builder/design-nix-builder.md
danieldk: Fixed.

docs/source/builder/design-nix-builder.md
danieldk: Added.

PR #561, opened 2026-05-18
Remove compatible variant enumeration, add decision trace

### LINKED ISSUE #436: Consider removing variant enumeration again and instead explain why each available are not compatible

danieldk:
We added support for enumerating variants that are compatible with the system. However, this has the issue that it is nearly impossible to make the enumeration exhaustive (e.g. because we have fallback from backend versions) and it is easy to miss variants.

As discussed on Slack, it would be better to instead have an explanation/decision trace on why the available variants were rejected. This PR add such traces and hooks them up in `kernels versions` and some error messages.

Fixes #436.

### DIFF
--- a/kernels/src/kernels/variants.py
+++ b/kernels/src/kernels/variants.py
@@ -1,4 +1,3 @@
-import itertools
 import logging
 import platform
 import re
@@ -17,7 +16,6 @@ from kernels.backends import (
     XPU,
     Backend,
     ROCm,
-    _backend,
     _select_backend,
     parse_backend,
 )
@@ -42,28 +40,6 @@ class Torch:
 
     version: Version
     cxx11_abi: bool | None
-
-    @staticmethod
-    def possible_variants() -> list["Torch"]:
-        if has_torch:
-            import torch
-
-            torch_version = parse(torch.__version__)
-            torch_version = Version(f"{torch_version.major}.{torch_version.minor}")
-
-            os_ = platform.system().lower()
-            if os_ == "linux":
-                cxx11_abi = torch.compiled_with_cxx11_abi()
-                return [
-                    Torch(version=torch_version, cxx11_abi=cxx11_abi),
-                    # We already accept build variants without an ABI tag, so
-                    # that we can remove the tag from builds in the future.
-                    Torch(version=torch_version, cxx11_abi=None),
-                ]
-            else:
-                return [Torch(version=torch_version, cxx11_abi=None)]
-        else:
-            return []
 
     @property
     def variant_str(self) -> str:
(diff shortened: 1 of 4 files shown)


danieldk:
API changes are expected (so the CI fails for that reason), they are not public symbols anyway.

kernels/src/kernels/variants.py
```diff
@@ -411,42 +382,80 @@ def _filter_variants(
...
             if v.arch.backend_name != noarch_backend_name and v.arch.backend_name != "universal":
+                result.append(
+                    VariantRejected(
+                        variant=v,
+                        reason=f"backend ({v.arch.backend_name}) does not match system backend ({noarch_backend_name}) and is not universal",
+                    )
+                )
                 continue
-        result.append(v)
+        result.append(VariantAccepted(variant=v))
```
sayakpaul, line 412: What happens in case a CUDA device isn't available?

kernels/tests/test_variants.py
```diff
@@ -353,7 +351,7 @@ def test_resolve_cuda_no_newer_minor_no_noarch():
 
 def test_resolve_cuda_no_different_major_no_noarch():
     # 11.8 has a different major, so there is no compatible fallback.
-    result = _resolve_variant_for_system(
+    result, trace = _resolve_variant_for_system(
```
sayakpaul, line 383: I think the tests should also assert again `trace` so that we know they emit what's expected.

danieldk: But how do we know what is expected? It depends on what system the tests are run on. I think the most we can do is that `result` is the same set as accepted inside the trace. I'll add that.

kernels/src/kernels/variants.py
danieldk: Currently we do not check device availability at all. If it's a Torch CUDA build, we'll give a CUDA kernel. Otherwise, the caller has to call `get_kernel` with the CPU backend.

kernels/tests/test_variants.py
sayakpaul: Yeah that should suffice.

danieldk: Added the following: the accepted variants in the trace must be equal to the returned variants. The variants in the trace (accepted and rejected) must be the same set as the input.

sayakpaul (approved):
Thanks for this massive user improvement!

PR #562, opened 2026-05-18
ci: kernel flakes are now required to be in a GitHub repo

danieldk:


.github/workflows/test_e2e.yaml
```diff
@@ -89,6 +89,13 @@ jobs:
           cd /tmp/kernels-upload-test
           sed -i 's|github:huggingface/kernels|path:'"$GITHUB_WORKSPACE"'|' flake.nix
 
+      - name: Make flake a Git repo
+        run: |
+          cd /tmp/kernels-upload-test
+          git config --global user.email "bottie@mcbotface.hf.co"
+          git config --global user.name "Botty McBotface"
```
sayakpaul, line 96: Good username.

.github/workflows/test_e2e.yaml
```diff
@@ -89,6 +89,13 @@ jobs:
           cd /tmp/kernels-upload-test
           sed -i 's|github:huggingface/kernels|path:'"$GITHUB_WORKSPACE"'|' flake.nix
 
+      - name: Make flake a Git repo
```
sayakpaul, line 92: Curious why is it needed.

sayakpaul (approved):

.github/workflows/test_e2e.yaml
danieldk: We made it mandatory that a flake is a git repo and has at least one commit, so that we can always use a git shorthash that is used for the kernel identifier (and not completely random material).

PR #563, opened 2026-05-18
Add pre-commit hooks for ruff, nixfmt, and avoiding main commits

danieldk:
We can add Cargo hooks later as well.


sayakpaul (approved):
I love how the pre commit config is written. 10/10

PR #569, opened 2026-05-19
nix-builder: add a hook to detect incorrect op registrations

danieldk:
Also update the documentation to mention that ops should be prefixed with the correct name.


docs/source/builder/writing-kernels.md
```diff
@@ -345,6 +345,47 @@ def mykernel(x: torch.Tensor, out: Optional[torch.Tensor] = None) -> torch.Tenso
     return out
 ```
 
+## Registering Torch operators
+
+You may want to register Torch ops from your kernel's Python code or
+register fake ops for `torch.compile` support. It is important register
```
sayakpaul:
```suggestion
register fake ops for `torch.compile` support. It is important to register
```

docs/source/builder/writing-kernels.md
```diff
@@ -345,6 +345,47 @@ def mykernel(x: torch.Tensor, out: Optional[torch.Tensor] = None) -> torch.Tenso
...
+from ._ops import add_op_namespace_prefix
+
+@torch.library.register_fake(add_op_namespace_prefix("relu_fwd"))
+def relu_fwd_fake(input: torch.Tensor) -> torch.Tensor:
+    return torch.empty_like(input)
+```
+
+As mentioned in the above, the `_ops` module is generated by kernel-builder.
+
+kernel-builder uses a hook to reject incorrect uses Torch op registration
```
sayakpaul:
```suggestion
kernel-builder uses a hook to reject incorrect usage of Torch op registration
```

examples/kernels/silu-and-mul-bad-registration/torch-ext/silu_and_mul_bad_registration/__init__.py
```diff
@@ -0,0 +1,12 @@
+import torch
```
sayakpaul, line 1: Very very good idea to include an example to test bad registration sites!

nix-builder/pkgs/torch-ops-check/torch-ops-check-hook.py
```diff
@@ -0,0 +1,90 @@
+import argparse
+import ast
+import sys
+from pathlib import Path
+
+TORCH_LIBRARY_DECORATORS = frozenset(
```
sayakpaul, line 6: For now, this is fine (as you also mentioned internally).

sayakpaul (commented):
Thanks. Just some nits. 

But I don't see `examples/kernels/silu-and-mul-bad-registration` being used anywhere. Did we mean to include some checks against it in the CI?

danieldk:
> But I don't see `examples/kernels/silu-and-mul-bad-registration` being used anywhere. Did we mean to include some checks against it in the CI?

https://github.com/huggingface/kernels/pull/569/changes#diff-a9fac6739b3ed28b5afba5be10a701a956fc148443963e2216204f634f2ec6c4

nix-builder/lib/checks.nix
```diff
@@ -11,11 +13,23 @@ let
     inherit buildSets;
     path = ../../examples/kernels/relu-torch-bounds;
   };
+
+  badRegistrationCheck = testers.testBuildFailure' {
+    drv = runCommand "check-bad-registration" { buildInputs = [ python3 ]; } ''
```
danieldk: Let me try though if I can make this fully use the hook (and not just the script).

danieldk: Uses the full hook by testing whether building the kernel passes.

sayakpaul (approved):
So cool!

PR #570, opened 2026-05-19
nix-builder: remove Torch 2.10 support

danieldk:


drbh (approved):
nice looks good to me

PR #572, opened 2026-05-20
Add source to metadata

drbh:
This PR adds a new optional `source` attribute to the metadata struct, this `source` is different than `upstream` since it is intended to point to the kernel builder compliant source related to the kernel. This distinction is important since the `upstream` may not originally be formatted in a way that the kernel-builder can compile - yet we want to be able to indicate the original code, and the code used to compile the kernel.

Note this PR also adds a step to validate the `upstream` and the `source` fields to be git urls


docs/source/builder/writing-kernels.md
```diff
@@ -190,6 +190,11 @@ The following sections enumerate all supported options for `build.toml`.
   branch named `v<version>`.
 - `backends` (required): a list of supported backends. Must be one or
   more of `cpu`, `cuda`, `metal`, `rocm`, or `xpu`.
+- `upstream`: URL of the original upstream repository where the kernel
+  source code comes from.
+- `source`: URL of the kernel-builder formatted source repository. This
+  repository must contain a `build.toml` and `flake.nix` so that it can be
+  pulled and built with the kernel builder.
```
danieldk: In both cases, I think we should say something about the format. Should this be a URL that can be passed directly to `git`? What about Git over SSH?

docs/source/kernel-requirements.md
```diff
@@ -71,6 +71,10 @@ metadata. Currently the following top-level keys are supported:
 - `version` (`int`, required): the kernel version number.
 - `license` (`str`, required): the kernel license in. Refer to the
   list of [supported license identifiers](https://huggingface.co/docs/hub/repositories-licenses).
+- `upstream` (`str`, optional): URL of the original upstream repository
+  where the kernel source code comes from.
+- `source` (`str`, optional): URL of the kernel-builder formatted source
+  repository (must contain `build.toml` and `flake.nix`).
```
danieldk: See above.

docs/source/builder/writing-kernels.md
danieldk: Now saw GitUrl below, so I think good to mention here that these should be Git-compatible (passable to `git clone`).

danieldk (commented):
Nice!!! Added some comments on the docs.

sayakpaul:
I like the idea. It would be cool to ship!

danieldk (approved):

PR #573, opened 2026-05-20
feat: prefer unique e2e ci repo rather than branch

drbh:
This PR restructures the e2e workflow to write to a unique repo id per run id and run attempt rather than a fixed repo id. This has the benefit of avoiding conflicts of writing to the same repo, and removes the need to add a concurrency group to handle possible race condition. These changes ensure that a each ci job will create a new repo and push, test it works, then clean up the repo at the end of the workflow.


sayakpaul (dismissed):
More reliable indeed.

danieldk (dismissed):
Nice! Seems to have surfaced an issue.

PR #575, opened 2026-05-21
Support the Torch stable ABI

danieldk:
This change adds an option `stable-abi` to `build.toml`. When setting                                                                                                                                                             
                                                                                                                                                                                                                                      
```toml
[torch]
stable-abi = "2.11"
```                                                                                                                                                                                                                               
                                                                                                                                                                                                                                      
The extension will be built targeting the Torch 2.11 stable ABI. In this case, the variant naming pattern will also change from `torch<version>` to `torch-stable-abi<abi-version>`.

This PR also adds support for this new type of variant to `kernels` and the Nix builder. In the case of the Nix builder, we always use the latest Torch version available for a backend + backend version combination to build the ABI-stable kernel.

This change also removes the auto-update of v3 `build.toml` to v4 `build.toml` format. The auto-conversion can give very weird errors when there is an issue with the file. For instance, when the `stable-abi` option is malformed, the error message will claim that there is no `torch` section (due to how serde deserializes these struct enums). The `update-build` subcommand is still available.

### DIFF
--- a/docs/source/builder/writing-kernels.md
+++ b/docs/source/builder/writing-kernels.md
@@ -227,6 +227,12 @@ options:
   non-compliant kernels if the version range does not correspond to the [required variants](build-variants.md).
 - `minver` (optional): only build for this Torch version and later. Use cautiously, since this option produces
   non-compliant kernels if the version range does not correspond to the [required variants](build-variants.md).
+- `stable-abi` (**experimental**): when set to a Torch version (e.g.
+  `"2.11"`), the kernel is built using the Torch stable ABI. This
+  requires that the kernel itself only
+  [stable ABI headers](https://docs.pytorch.org/docs/2.12/notes/libtorch_stable_abi.html).
+  For an example, see the [`relu-torch-stable-abi`](https://github.com/huggingface/kernels/tree/main/examples/kernels/relu-torch-stable-abi)
+  example kernel.
 
 ### `kernel.<name>`
 
@@ -276,7 +282,6 @@ are available:
 
 - `cxx-flags`: a list of additional flags to be passed to the C++
   compiler.
-
 
 ## Torch bindings
 
--- a/docs/source/builder/writing-kernels.md
+++ b/docs/source/builder/writing-kernels.md
@@ -229,7 +229,7 @@ options:
   non-compliant kernels if the version range does not correspond to the [required variants](build-variants.md).
 - `stable-abi` (**experimental**): when set to a Torch version (e.g.
   `"2.11"`), the kernel is built using the Torch stable ABI. This
-  requires that the kernel itself only
+  requires that the kernel itself only use
   [stable ABI headers](https://docs.pytorch.org/docs/2.12/notes/libtorch_stable_abi.html).
   For an example, see the [`relu-torch-stable-abi`](https://github.com/huggingface/kernels/tree/main/examples/kernels/relu-torch-stable-abi)
   example kernel.
--- a/kernel-builder/src/pyproject/templates/torch/build-variants.cmake
+++ b/kernel-builder/src/pyproject/templates/torch/build-variants.cmake
@@ -1,30 +1,51 @@
 # Generate a standardized build variant name following the pattern:
-# torch<VERSION>-[cxx11-]<COMPUTE>-<ARCH>-<OS>
+# torch<VERSION>-<COMPUTE>-<ARCH>-<OS>
+# or, when compiled against the Torch stable ABI:
+# torch-stable-abi<VERSION>-<COMPUTE>-<ARCH>-<OS>
 #
 # Arguments:
 #   OUT_BUILD_NAME - Output variable name
-#   TORCH_VERSION - PyTorch version (e.g., "2.7.1")
+#   TORCH_VERSION - PyTorch version (e.g., "2.7.1"); ignored when TORCH_STABLE_ABI is set
 #   COMPUTE_FRAMEWORK - One of: cuda, rocm, metal, xpu, cpu
 #   COMPUTE_VERSION - Version of compute framework (e.g., "12.4" for CUDA, "6.0" for ROCm)
 #                     Optional for CPU-only builds (pass empty string or omit)
-# Example output: torch271-cxx11-cu124-x86_64-linux (Linux)
-#                 torch271-cu124-x86_64-windows (Windows)
-#                 torch271-metal-aarch64-darwin (macOS)
+# Optional keyword arguments:
+#   TORCH_STABLE_ABI - Stable ABI version the extension was compiled against (e.g., "2.11");
+#                      when set, TORCH_VERSION is ignored and the prefix becomes
+#                      torch-stable-abi<VERSION> (e.g., "2.11" -> "torch-stable-abi211")
+# Example output: torch27-cu124-x86_64-linux (Linux)
+#                 torch27-cu124-x86_64-windows (Windows)
+#                 torch27-metal-aarch64-darwin (macOS)
+#                 torch-stable-abi211-cu124-x86_64-linux (Linux, stable ABI)
 #
 function(generate_build_name OUT_BUILD_NAME TORCH_VERSION COMPUTE_FRAMEWORK COMPUTE_VERSION)
-    # Flatten version by removing dots and padding to 2 components
-    string(REPLACE "." ";" VERSION_LIST "${TORCH_VERSION}")
-    list(LENGTH VERSION_LIST VERSION_COMPONENTS)
+    cmake_parse_arguments(ARG "" "TORCH_STABLE_ABI" "" ${ARGN})
 
-    # Pad to at least 2 components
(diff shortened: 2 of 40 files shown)


sayakpaul:
> For instance, when the stable-abi option is malformed, the error message will claim that there is no torch section (due to how serde deserializes these struct enums). 

Is there sufficient heuristics we can reliably follow to catch it on the builder's end? For example, can the `stable-abi` field in `build.toml` ever be outside of the "x.x" format?

docs/source/builder/writing-kernels.md
```diff
@@ -227,6 +227,12 @@ options:
   non-compliant kernels if the version range does not correspond to the [required variants](build-variants.md).
 - `minver` (optional): only build for this Torch version and later. Use cautiously, since this option produces
   non-compliant kernels if the version range does not correspond to the [required variants](build-variants.md).
+- `stable-abi` (**experimental**): when set to a Torch version (e.g.
+  `"2.11"`), the kernel is built using the Torch stable ABI. This
+  requires that the kernel itself only
```
sayakpaul:
```suggestion
  requires that the kernel itself only use
```

Otherwise, reads a bit weird.

kernel-builder/src/pyproject/templates/torch/build-variants.cmake
```diff
@@ -1,30 +1,51 @@
...
+#   TORCH_VERSION - PyTorch version (e.g., "2.7.1"); ignored when TORCH_STABLE_ABI is set
 #   COMPUTE_FRAMEWORK - One of: cuda, rocm, metal, xpu, cpu
 #   COMPUTE_VERSION - Version of compute framework (e.g., "12.4" for CUDA, "6.0" for ROCm)
 #                     Optional for CPU-only builds (pass empty string or omit)
-# Example output: torch271-cxx11-cu124-x86_64-linux (Linux)
-#                 torch271-cu124-x86_64-windows (Windows)
-#                 torch271-metal-aarch64-darwin (macOS)
+# Optional keyword arguments:
+#   TORCH_STABLE_ABI - Stable ABI version the extension was compiled against (e.g., "2.11");
+#                      when set, TORCH_VERSION is ignored and the prefix becomes
```
sayakpaul, line 14: Are we also logging some kind of warning to the user that this is getting ignored? Not too strongly opinionated about it, though.

kernel-builder/src/pyproject/templates/torch/preamble.cmake
```diff
@@ -65,11 +65,26 @@ if (TORCH_VERSION VERSION_GREATER {{ torch_maxver }})
 endif()
 {% endif %}
 
+{% if stable_abi %}
+if (TORCH_VERSION VERSION_LESS {{ stable_abi }})
+  message(FATAL_ERROR "Torch version ${TORCH_VERSION} is less than the stable ABI "
+    "version {{ stable_abi }}. Cannot build with stable ABI targeting a newer Torch.")
```
sayakpaul:
```suggestion
    "version {{ stable_abi }}. Cannot build with stable ABI targeting a newer version of Torch.")
```

kernel-builder/src/util.rs
```diff
@@ -49,18 +44,32 @@ pub(crate) fn check_or_infer_target_dir(
     }
 }
 
-pub(crate) fn parse_and_validate(kernel_dir: impl AsRef<Path>) -> Result<BuildCompat> {
```
sayakpaul, line 52: A bit unintuitive as to why we should have two different functions here. Can they be folded into a single function? If it's preferential, I'll understand that.

kernels-data/src/config/v4.rs
```diff
@@ -78,7 +78,7 @@ pub struct Hub {
 }
 
 #[derive(Debug, Deserialize, Clone, Serialize)]
-#[serde(deny_unknown_fields)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
```
sayakpaul, line 81: And this is for?

kernels/src/kernels/variants.py
```diff
@@ -24,8 +24,6 @@
 )
 from kernels.compat import has_torch, has_tvm_ffi
 
-BUILD_VARIANT_REGEX = re.compile(r"^(torch\d+\d+|torch-(cpu|cuda|metal|neuron|rocm|xpu)|tvm-ffi\d+\d+)")
```
sayakpaul, line 27: Unused variable?

kernels/src/kernels/variants.py
```diff
@@ -65,6 +63,32 @@ def parse(s: str) -> "Torch":
         return Torch(version=version, cxx11_abi=cxx11_abi)
 
 
+@dataclass(unsafe_hash=True)
+class TorchStableAbi:
```
sayakpaul, line 67: All the hardwork done in the past to streamline our approach towards handling variants is really now starting to show up. Thanks for all of that!

kernels/tests/test_variants.py
```diff
@@ -48,6 +48,26 @@
     ]
 )
 
+STABLE_ABI_VARIANT_STRINGS = [
```
sayakpaul, line 51: Very nice tests!

nix-builder/lib/variants/tvm-ffi.nix
```diff
@@ -17,7 +17,9 @@ let
       "xpu${flattenVersion (lib.versions.majorMinor buildConfig.xpuVersion)}"
     else
       "cpu";
+  arch = "tvm-ffi${flattenVersion (lib.versions.majorMinor buildConfig.tvmFfiVersion)}-${computeString}-${buildConfig.system}";
```
sayakpaul, line 20: On the surface, this seems like an unrelated change to me. But is this being done to uniformize how `kernelVariant` is leveraged in the torch builder?

sayakpaul (commented):
Very nice. I left some comments. 

Major comments:

* Seems like `kernels-data` is missing tests for the new changes?
* I am okay keeping the non Stable ABI and Stable ABI path separated but is there any merit in trying to club the two?

kernels-data/src/config/v4.rs
danieldk: So that the option name in `build.toml` is `stable-abi` and not `stable_abi`.

kernels/src/kernels/variants.py
danieldk: Yep.

nix-builder/lib/variants/tvm-ffi.nix
danieldk: Yes, this is to that both Torch and tvm-ffi have a `kernelVariant` function.

kernel-builder/src/pyproject/templates/torch/build-variants.cmake
danieldk: We are the user :). I don't really think this warrants a warning, since there is no way to fix it. We could require that the caller does not provide `TORCH_VERSION` in this case, but that makes things more complicated for the caller (more conditional logic).

danieldk:
> > For instance, when the stable-abi option is malformed, the error message will claim that there is no torch section (due to how serde deserializes these struct enums).
> 
> Is there sufficient heuristics we can reliably follow to catch it on the builder's end? For example, can the `stable-abi` field in `build.toml` ever be outside of the "x.x" format?

No and that is validated through the `Version` type. Serde is really good at doing validation and providing diagnostics already. The issue is that if you have something like:

```
struct Foo {
  some_field: MyEnum,
}

enum Bar {
  OptionA { ... },
  OptionB { ... },
}
```

That when `some_field` is deserialized, serde tries to deserialize it as the first of `OptionA` and `OptionB` it can be deserialized as (when there is no enum tagging). But if neither works, it is hard to give proper diagnostics, because it's not clear whether the user meant `OptionA` or `OptionB` (even more so in our case, where they are very similar).

I did this when we went from `v1` -> `v2`, because it's a convenient way of falling back when the serialization fails. We could add some logic that tries to deserialize as the newest version (and give diagnostics based on that) when all fail. But I think `v3` -> `v4` is quite messy, because the versions are so similar (in contrast to prior versions).

I think in the future we might want to switch to something like Rust's edition system (maybe with v5?) where we put a version in `general` and then we always know unambiguously how to deserialize the format.

kernel-builder/src/util.rs
danieldk: They serialize to different types and have different return types, so I'm not 100% sure how to put them in one function.

sayakpaul:
> I think in the future we might want to switch to something like Rust's edition system (maybe with v5?) where we put a version in general and then we always know unambiguously how to deserialize the format.

If that makes v5 -> v6 upgrades (and the subsequent ones) easier, sure.

sayakpaul (approved):
Let's also fix the lint errors before merging.

danieldk:
> Let's also fix the lint errors before merging.

The lints are just API compat due to the removal of `BUILD_VARIANT_REGEX`.

PR #579, opened 2026-05-22
kernel-builder: improve GPU arch handling

danieldk:
Add a bunch of improvements to the GPU arch handling code:

- Completely remove `arch.nix`. This file was originally used to for compiling Torch and for determining supported archs list. However, we repackage Torch binaries and we use our own arch list.
- Completely separate CUDA and HIP CMake code generation. This is cleaner and fixes an issue where when compiling with ROCm, the CUDA code would also get called, since it was not guarded.
- Improve per-kernel GPU arch reporting.


kernel-builder/src/pyproject/templates/kernel.cmake
```diff
@@ -32,97 +31,116 @@ function(cuda_kernel_component SRC_VAR)
...
-            cuda_archs_loose_intersection(_KERNEL_ARCHS "${KERNEL_CUDA_CAPABILITIES}" "${CUDA_ARCHS}")
-        else()
-            set(_KERNEL_ARCHS "${CUDA_KERNEL_ARCHS}")
+    # Determine CUDA architectures
+    if(KERNEL_CUDA_CAPABILITIES)
+        cuda_archs_loose_intersection(_KERNEL_ARCHS "${KERNEL_CUDA_CAPABILITIES}" "${CUDA_ARCHS}")
+    else()
+        set(_KERNEL_ARCHS "${CUDA_KERNEL_ARCHS}")
+    endif()
+    message(STATUS "CUDA kernel: ${KERNEL_NAME}, capabilities: ${_KERNEL_ARCHS}")
```
danieldk, line 40: This is the only real change here, the rest is just moving the code out of the conditional block due to the CUDA/HIP split.

kernel-builder/src/pyproject/templates/kernel.cmake
```diff
@@ -32,97 +31,116 @@ function(cuda_kernel_component SRC_VAR)
...
-        set(_TMP_SRC ${${SRC_VAR}})
-        list(APPEND _TMP_SRC ${_KERNEL_SRC})
-        set(${SRC_VAR} ${_TMP_SRC} PARENT_SCOPE)
+    # Determine ROCm architectures
+    if(KERNEL_ROCM_ARCHS)
+        hip_archs_loose_intersection(_KERNEL_ARCHS "${KERNEL_ROCM_ARCHS}" "${ROCM_ARCHS}")
+    else()
+        set(_KERNEL_ARCHS "${ROCM_ARCHS}")
     endif()
+    message(STATUS "ROCm kernel: ${KERNEL_NAME}, archs: ${_KERNEL_ARCHS}")
```
danieldk, line 121: This is the only real change here, the rest is just moving the code out of the conditional block due to the CUDA/HIP split.

kernel-builder/src/pyproject/templates/utils.cmake
```diff
@@ -500,53 +500,6 @@ function(cuda_remove_ptx_suffixes OUT_CUDA_ARCHS CUDA_ARCHS)
...
-
-#
-# Override the GPU architectures detected by cmake/torch and filter them by
-# `GPU_SUPPORTED_ARCHES`. Sets the final set of architectures in
-# `GPU_ARCHES`. This only applies to the HIP language since for CUDA we set
-# the architectures on a per file basis.
-#
-# Note: this is defined as a macro since it updates `CMAKE_CUDA_FLAGS`.
-#
-macro(override_gpu_arches GPU_ARCHES GPU_LANG GPU_SUPPORTED_ARCHES)
```
danieldk, line 513: Unused now, remove.

kernel-builder/src/pyproject/templates/kernel-component/hip.cmake
```diff
@@ -0,0 +1,10 @@
+if(GPU_LANG STREQUAL "HIP")
```
sayakpaul, line 1: Nice separation.

kernel-builder/src/pyproject/templates/kernel.cmake
sayakpaul: Would it make sense to split the CUDA and HIP functions into their own scripts and use them here or too much moving around?

kernel-builder/src/pyproject/kernel.rs
```diff
@@ -83,46 +86,67 @@ fn render_kernel_component_cuda(
...
-        Kernel::Rocm {
-            rocm_archs,
-            hip_flags,
-            ..
-        } => (
-            None,
-            rocm_archs.as_deref(),
-            None,
-            hip_flags.as_deref(),
-            None,
```
sayakpaul, line 108: Very nice. Way less confusing.

nix-builder/lib/extension/torch/arch.nix
```diff
@@ -224,10 +224,9 @@ stdenv.mkDerivation (prevAttrs: {
   env =
     lib.optionalAttrs cudaSupport {
       CUDAToolkit_ROOT = "${lib.getDev cudaPackages.cuda_nvcc}";
-      TORCH_CUDA_ARCH_LIST = lib.concatStringsSep ";" torch.cudaCapabilities;
     }
     // lib.optionalAttrs rocmSupport {
-      PYTORCH_ROCM_ARCH = lib.concatStringsSep ";" torch.rocmArchs;
+      #PYTORCH_ROCM_ARCH = lib.concatStringsSep ";" torch.rocmArchs;
```
sayakpaul: Did we plan to remove?

sayakpaul (commented):
Awesome, left some questions.

kernel-builder/src/pyproject/templates/kernel.cmake
danieldk: I considered that in the big CMake refactor a few months back, but in the end we need all those functions anyway (since the kernel may be multi-backend), so I decided to put them together and do the variable substitution in `cuda.cmake`, `cpu.cmake`, `xpu.cmake`, etc.

nix-builder/lib/extension/torch/arch.nix
danieldk: Ah yes, forgot. Removed now.

sayakpaul (approved):

PR #580, opened 2026-05-22
feat: make kernel loading work when offline mode ie enabled.

### LINKED ISSUE #553: Offline kernel loading fails despite cached snapshot
### Description

There seems to be an issue with how kernels are loaded when `HF_HUB_OFFLINE` is set to 1 even though a snapshot is available locally in the HF cache.

The issue appears to come from `install_kernel()` in kernels/utils.py:

In [utils.py](https://github.com/huggingface/kernels/blob/97c6d1e9e51dee3a7bdd96abb493f10046501531/kernels/src/kernels/utils.py), `get_variants()` calls `api.list_repo_tree(...)`, which seems to attempts to access the Hub. In offline mode this raises, even though the required files are already available locally under the cached snapshot.

Shouldn't the kernels be loadable locally with `get_variants_local()`?

### Steps to reproduce

With `HF_HUB_OFFLINE` = 1

```python
get_kernel("kernels-community/finegrained-fp8", revision="b0b199bf716f33c14bb7a3ea9ca86b87e5635906")
```

### Expected behavior

Should load from local

### Environment

kernels: 0.13.0
python: 3.13.7
pytorch: 2.8.0+cu126
GPU: H200
linux: Red Hat Enterprise Linux release 8.7
cuda: 13.0

### Additional context

```
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    get_kernel("kernels-community/finegrained-fp8", revision="b0b199bf716f33c14bb7a3ea9ca86b87e5635906")
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/kernels/utils.py", line 282, in get_kernel
    package_name, variant_path = install_kernel(
                                 ~~~~~~~~~~~~~~^
        repo_id, revision=revision, backend=backend, user_agent=user_agent
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/kernels/utils.py", line 137, in install_kernel
    variants = get_variants(api, repo_id=repo_id, revision=revision)
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/kernels/variants.py", line 300, in get_variants
    item.path.split("/")[-1] for item in tree if isinstance(item, RepoFolder)
                                         ^^^^
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/huggingface_hub/hf_api.py", line 3836, in list_repo_tree
    for path_info in paginate(path=tree_url, headers=headers, params={"recursive": recursive, "expand": expand}):
                     ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/huggingface_hub/utils/_pagination.py", line 35, in paginate
    r = session.get(path, params=params, headers=headers)
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/httpx/_client.py", line 1053, in get
    return self.request(
           ~~~~~~~~~~~~^
        "GET",
        ^^^^^^
    ...<7 lines>...
        extensions=extensions,
        ^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/httpx/_client.py", line 825, in request
    return self.send(request, auth=auth, follow_redirects=follow_redirects)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/httpx/_client.py", line 914, in send
    response = self._send_handling_auth(
        request,
    ...<2 lines>...
        history=[],
    )
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/httpx/_client.py", line 942, in _send_handling_auth
    response = self._send_handling_redirects(
        request,
        follow_redirects=follow_redirects,
        history=history,
    )
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/httpx/_client.py", line 977, in _send_handling_redirects
    hook(request)
    ~~~~^^^^^^^^^
  File "/home2020/home/lilpa/edoyen/miniconda/envs/gflcg/lib/python3.13/site-packages/huggingface_hub/utils/_http.py", line 224, in hf_request_event_hook
    raise OfflineModeIsEnabled(
        f"Cannot reach {request.url}: offline mode is enabled. To disable it, please unset the `HF_HUB_OFFLINE` environment variable."
    )
huggingface_hub.errors.OfflineModeIsEnabled: Cannot reach https://huggingface.co/api/models/kernels-community/finegrained-fp8/tree/b0b199bf716f33c14bb7a3ea9ca86b87e5635906/build?recursive=false&expand=false: offline mode is enabled. To disable it, please unset the `HF_HUB_OFFLINE` environment variable.
```

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/553


kernels/tests/test_basic.py
```diff
@@ -243,6 +245,67 @@ def test_trust_remote_code_flag_allows_untrusted():
     get_kernel("kernels-test-untrusted/ci-test-kernel", version=1, trust_remote_code=True)
 
 
+def test_install_kernel_offline_with_revision(monkeypatch, local_kernel_path):
```
danieldk, line 248: Love these tests!

kernels/src/kernels/utils.py
```diff
@@ -62,6 +62,18 @@ def _check_trust_remote_code(repo_id: str, trust_remote_code: bool | list[str])
...
         )
 
+    if constants.HF_HUB_OFFLINE:
+        # Publisher trust cannot be verified offline. The user opted into
+        # offline mode and the kernel must already be in the local cache,
+        # so trust was established when it was originally downloaded.
+        warnings.warn(
+            f"Skipping publisher trust check for '{repo_id}' because Hugging Face "
+            "Hub is in offline mode. Pass trust_remote_code=True to suppress this "
+            "warning.",
```
danieldk:
```suggestion
            f"Skipping publisher trust check for '{repo_id}' because Hugging Face "
            "Hub is in offline mode.",
```

Just to avoid incentivizing people to use `trust_remote_code`.

kernels/src/kernels/utils.py
```diff
@@ -513,7 +548,29 @@ def load_kernel(
...
+                    )
+                )
+            )
+        except LocalEntryNotFoundError as e:
+            raise FileNotFoundError(
+                f"Locked kernel `{repo_id}` was not downloaded. Make sure it's downloaded locally."
+            ) from e
+        variants = get_variants_local(local_repo_path / "build")
+    else:
+        variants = get_variants(api, repo_id=repo_id, revision=locked_sha)
```
danieldk: These should actually be the same code path. I think the current implementation si wrong in that it calls `get_variants`, `load_kernel` should already be local/offline, but I think the network dependency got accidentally introduced when adding the `get_variants` code.

So, it should be your code in both cases (`constants.HF_HUB_OFFLINE` and `constants.HF_HUB_OFFLINE`).

kernels/src/kernels/utils.py
danieldk: I think if we add `revision` here as well, then `install_kernel` could just call `load_kernel` when `constants.HF_HUB_OFFLINE` is set. So the cases would be:

* `revision=None`, `lockfile=None` -> get lock file from the caller package's metadata and get revision.
* `revision=something`, `lockfile=None` -> use the given revision.
* `revision=None`, `lockfile=some_path` -> get the revision from the lockfile.
* `revision=something`, `lockfile=some_path` -> raise an exception

kernels/src/kernels/utils.py
```diff
@@ -244,11 +256,34 @@ def install_kernel(
...
+                    )
+                )
+            )
+        except LocalEntryNotFoundError as e:
+            raise FileNotFoundError(
+                f"Cannot find a local snapshot for {repo_id} (revision: {revision}). "
+                "When Hugging Face Hub is in offline mode the kernel must already "
+                "be present in the local cache."
+            ) from e
+        variants = get_variants_local(local_repo_path / "build")
```
danieldk: See below, I think with a small change to `load_kernel`, we could call out to `load_kernel` here.

kernels/src/kernels/utils.py
```diff
@@ -244,10 +254,21 @@ def install_kernel(
         `Path`: The path to the variant directory.
     """
     api = _get_hf_api(user_agent=user_agent)
+    offline = local_files_only or constants.HF_HUB_OFFLINE
+
+    if offline:
```
danieldk: Since `offline` is not used anywhere else, I'd:


```suggestion
    if local_files_only or constants.HF_HUB_OFFLINE:
```

kernels/src/kernels/utils.py
```diff
@@ -497,13 +570,20 @@ def load_kernel(
...
             The backend will be detected automatically if not provided.
+        revision (`str`, *optional*):
+            The specific revision (branch, tag, or commit) to download. Cannot be used together with `version`.
 
     Returns:
         `ModuleType`: The imported kernel module.
     """
-    if lockfile is None:
+    if lockfile and revision:
+        raise ValueError("`lockfile` and `revision` both cannot be specified at the same time.")
```
danieldk:
```suggestion
    if lockfile is not None and revision is not None:
        raise ValueError("`lockfile` and `revision` both cannot be specified at the same time.")
```

Otherwise this would pass with two 'falsy' values (e.g. an empty string).

kernels/src/kernels/utils.py
```diff
@@ -244,10 +254,21 @@ def install_kernel(
         `Path`: The path to the variant directory.
     """
     api = _get_hf_api(user_agent=user_agent)
+    offline = local_files_only or constants.HF_HUB_OFFLINE
+
+    if offline:
+        # Same local-cache resolution path used by `load_kernel`, which is
+        # always offline. Sharing the helper avoids the network dependency
+        # that `get_variants` would otherwise introduce.
+        return _resolve_local_variant_path(
```
danieldk, line 261: Can't we `load_kernel` here and keep the body of `_resolve_local_variant_path` inside `load_kernel`? It would more clearly express the relation between the functions.

sayakpaul: Thought about it a bit but `install_kernel` returns `Path`, whereas `load_kernel` returns `ModuleType`. To recover the path from a module return by `load_kernel()`, we will have to do some kind of parsing. 

Then the `_import_from_path()` utility also needs some changes. 

The flow would become `get_kernel` -> `install_kernel` > `load_kernel` -> `_import_from_path(repo_info=None)` — that caches `LoadedKernel(repo_info=None)`. 

Then `get_kernel` does its own `_import_from_path(variant_path, repo_info=RepoInfo(...))`, hits the cache, and returns the module while silently discarding the actual `repo_info`. 

As a result, `get_loaded_kernels()` would report `repo_info=None` for kernels loaded via `get_kernel` in offline mode which would be a regression, if not handled in `_import_from_path`.

I think the shared utility solution is a cleaner approach here.

danieldk: Ah, got it, thanks!

danieldk (approved):

danieldk:
Really nice! Let's get it merged to have it in the release that we'll do soon.

PR #582, opened 2026-05-25
feat: implement coverage reporting

sayakpaul:
Some in-line comments.

Should we also run coverage on `main` pushes and keep a status on README? I think we don't need it. Open to suggestions.

### DIFF
--- a/.github/workflows/coverage_comment.yaml
+++ b/.github/workflows/coverage_comment.yaml
@@ -48,8 +48,16 @@ jobs:
             exit 1
           fi
 
+          # Only match comments authored by the GITHUB_TOKEN bot, so a human
+          # comment that happens to start with the marker doesn't get
+          # silently overwritten.
           EXISTING=$(gh api "repos/${REPO}/issues/${PR_NUMBER}/comments" --paginate \
-            --jq 'first(.[] | select(.body | startswith("<!-- coverage-report -->")) | .id) // empty')
+            --jq 'first(
+                    .[]
+                    | select(.user.login == "github-actions[bot]"
+                             and (.body | startswith("<!-- coverage-report -->")))
+                    | .id
+                  ) // empty')
 
           # Body comes from the untrusted upstream run; pass via --rawfile +
           # --input - so it's never shell-interpolated.
--- a/.github/workflows/coverage_comment.yaml
+++ b/.github/workflows/coverage_comment.yaml
@@ -1,0 +1,62 @@
+name: Post coverage comment
+
+on:
+  workflow_run:
+    workflows: ["Test kernels"]
+    types: [completed]
+
+concurrency:
+  group: coverage-comment-${{ github.event.workflow_run.head_sha }}
+  cancel-in-progress: true
+
+jobs:
+  post:
+    name: Post coverage comment
+    runs-on: ubuntu-latest
+    if: >-
+      github.event.workflow_run.conclusion == 'success' &&
+      github.event.workflow_run.event == 'pull_request'
+    permissions:
+      contents: read
+      pull-requests: write
+
+    steps:
+      - name: Download coverage artifact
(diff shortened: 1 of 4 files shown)


.github/workflows/test_kernels.yaml
```diff
@@ -98,3 +106,63 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          mkdir -p coverage-artifact
+
+          PCT=$(jq -r '.totals.percent_covered' coverage.json)
+          PCT_INT=$(printf '%.0f' "$PCT")
+          PCT_DISPLAY=$(printf '%.1f' "$PCT")
+          if [ "$PCT_INT" -ge 80 ]; then
+            EMOJI=":white_check_mark:"
+          elif [ "$PCT_INT" -ge 70 ]; then
```
sayakpaul, line 129: Open to other percentages.

.github/workflows/test_kernels.yaml
```diff
@@ -31,6 +31,7 @@ jobs:
 
     env:
       UV_PYTHON_PREFERENCE: only-managed
+      COVERAGE_CELL: ${{ matrix.python-version == '3.10' && matrix.torch-version == '2.12.0' }}
```
sayakpaul: Only run for a single path. But can increase it.

.github/workflows/coverage_comment.yaml
```diff
@@ -0,0 +1,62 @@
...
+          # Validate: must be a positive integer. Defends against a malicious
+          # PR replacing pr_number.txt with arbitrary content that we'd then
+          # interpolate into a URL.
+          if ! [[ "${PR_NUMBER}" =~ ^[1-9][0-9]*$ ]]; then
+            echo "::error::pr_number.txt does not contain a valid PR number: ${PR_NUMBER}"
+            exit 1
+          fi
+
+          EXISTING=$(gh api "repos/${REPO}/issues/${PR_NUMBER}/comments" --paginate \
+            --jq 'first(.[] | select(.body | startswith("<!-- coverage-report -->")) | .id) // empty')
```
danieldk: What if a user makes a comment with that marker? I don't think there is an issue, since it'd just overwrite their comment?

.github/workflows/test_kernels.yaml
danieldk: Doing a single path sounds good to me!

danieldk: The naming seems a bit strange? Maybe `CHECK_COVERAGE`?

.github/workflows/test_kernels.yaml
```diff
@@ -98,3 +106,63 @@ jobs:
...
+            EMOJI=":red_circle:"
+          fi
+
+          uv run coverage report --format=markdown > coverage-artifact/cov-table.md
+
+          {
+            echo "<!-- coverage-report -->"
+            echo "## Coverage report — \`kernels/\`"
+            echo
+            echo "> Measured on the canonical matrix cell: **Python ${PY_VERSION} / Torch ${TORCH_VERSION}**."
```
danieldk:
```suggestion
            echo "> Measured on: **Python ${PY_VERSION} / Torch ${TORCH_VERSION}**."
```

.github/workflows/test_kernels.yaml
```diff
@@ -98,3 +106,63 @@ jobs:
...
+            echo "> Other matrix cells are not included in this number."
```
danieldk:
```suggestion
            echo "> Other CI configurations are not included in this number."
```

danieldk (dismissed):
Looks good, some nits about the naming, feel free to ignore.

.github/workflows/coverage_comment.yaml
sayakpaul: Good catch. Addressed.

sayakpaul:
There is an unrelated failure in the CI. I will fix it later. @danieldk

danieldk (approved):

PR #588, opened 2026-05-27
[ci] optimize ci runtime ⚡️

sayakpaul:
Most workflows trigger on every PR that touches anything outside `docs/**` and `*.md`. 

For example, on [#583](https://github.com/huggingface/kernels/pull/583), for small changes w.r.t `ruff` in `.pre-commit-config.yaml` and `Makefile` triggered ~50 CI jobs. This is not desirable. An unnecessarily long CI is probably not the best use of the resources configured for our CI and also generally, not a good developer/contribution experience.

To put things in numbers and perspective (with the help of Claude):

| Job (ran on #583) | Runtime | Could the diff affect it? |
|---|---|---|
| Build kernel (windows-2022, py3.12, torch 2.9.1) | 32m 47s | no |
| Test kernels (UBI 8) — Docker GPU | 7m 40s | no |
| Test kernels (UBI 9) — Docker GPU | 7m 40s | no |
| Build kernels (x86_64-linux) — Nix | 4m 44s | no |
| Build kernels (aarch64-linux) — Nix | 3m 23s | no |
| Build kernel (ROCm) | 8m 58s | no |
| Build kernel (XPU) | 8m 0s | no |
| Build kernel (macOS / Nix) | 4m 4s | no |
| Nix checks | 4m 22s | no |
| `kernel_abi_python_release` wheel matrix (linux × 6 arches, musllinux × 4, win × 3, mac × 2, sdist) | ~3 min × 16 | no |
| `kernels_data_python_release` wheel matrix (same shape) | ~3 min × 16 | no |
| Run kernels tests on `aws-g6-24xlarge` (GPU) × 4 matrix combos | ~2 min × 4 | no |
| Rust fmt/clippy/test × 3 crates | ~3 min total | no |
| Check build variants | 33s | no |
| Check API compatibility / griffe | ~10s | no |
| Run lints (ruff) | 5s | **yes** |

This PR tries to optimize our CI _without_ breaking robustness and reliability.

---

## Major changes

* For each workflow, declare the **set of paths whose changes can affect this workflow's outcome**. Anything else is skipped on PRs and should be safe enough that way.
* No changes to the workflows that are triggered when pushing to `main`. 
* Add concurrency groups. Several workflows (`test_kernels.yaml`, `test_python.yaml`, etc.) still lack concurrency groups. When a contributor force-pushes a PR, the previous run keeps consuming runners.
* Other changes are commented in-line.

After this PR, we should expect to see good amount of reductions in the CI runtimes where relevant.

### DIFF
--- a/.github/workflows/build_kernel.yaml
+++ b/.github/workflows/build_kernel.yaml
@@ -5,10 +5,19 @@ on:
   pull_request:
     branches: [main]
     types: [opened, synchronize, reopened] # trigger on PRs
-    paths-ignore:
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "nix-builder/**"
+      - "kernel-builder/**"
+      - "examples/kernels/**"
+      - "flake.nix"
+      - "flake.lock"
+      - "default.nix"
+      - ".github/workflows/build_kernel.yaml"
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   build:
--- a/.github/workflows/build_kernel.yaml
+++ b/.github/workflows/build_kernel.yaml
@@ -9,9 +9,8 @@ on:
       - "nix-builder/**"
       - "kernel-builder/**"
       - "examples/kernels/**"
-      - "flake.nix"
-      - "flake.lock"
-      - "default.nix"
+      - "**/*.nix"
+      - "**/flake.lock"
       - ".github/workflows/build_kernel.yaml"
   workflow_dispatch:
 
--- a/.github/workflows/build_kernel_macos.yaml
+++ b/.github/workflows/build_kernel_macos.yaml
@@ -5,10 +5,19 @@ on:
   pull_request:
     branches: [main]
     types: [opened, synchronize, reopened] # trigger on PRs
-    paths-ignore:
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "nix-builder/**"
+      - "kernel-builder/**"
+      - "examples/kernels/**"
+      - "flake.nix"
+      - "flake.lock"
+      - "default.nix"
+      - ".github/workflows/build_kernel_macos.yaml"
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   build:
--- a/.github/workflows/build_kernel_macos.yaml
+++ b/.github/workflows/build_kernel_macos.yaml
@@ -9,9 +9,8 @@ on:
       - "nix-builder/**"
       - "kernel-builder/**"
       - "examples/kernels/**"
-      - "flake.nix"
-      - "flake.lock"
-      - "default.nix"
+      - "**/*.nix"
+      - "**/flake.lock"
       - ".github/workflows/build_kernel_macos.yaml"
   workflow_dispatch:
 
--- a/.github/workflows/build_kernel_rocm.yaml
+++ b/.github/workflows/build_kernel_rocm.yaml
@@ -5,10 +5,19 @@ on:
   pull_request:
     branches: [main]
     types: [opened, synchronize, reopened] # trigger on PRs
-    paths-ignore:
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "nix-builder/**"
+      - "kernel-builder/**"
+      - "examples/kernels/**"
+      - "flake.nix"
+      - "flake.lock"
+      - "default.nix"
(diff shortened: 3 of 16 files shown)


.github/workflows/build_kernel.yaml
```diff
@@ -5,11 +5,20 @@ on:
...
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "nix-builder/**"
+      - "kernel-builder/**"
+      - "examples/kernels/**"
+      - "flake.nix"
+      - "flake.lock"
+      - "default.nix"
+      - ".github/workflows/build_kernel.yaml"
```
sayakpaul, line 14: Didn't see any other dependency.

.github/workflows/test_kernels.yaml
```diff
@@ -6,18 +6,20 @@ on:
...
     branches: [main]
     types: [opened, synchronize, reopened] # trigger on PRs
-    paths-ignore:
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "kernels/**"
+      - "kernels-data/bindings/python/**"
+      - "kernel-abi-check/bindings/python/**"
+      - ".github/workflows/test_kernels.yaml"
```
sayakpaul, line 13: I don't think there's any dependency on `kernel-builder` here. So, we probably shouldn't trigger it when changes are made to `kernel-builder`, `nix-builder`, etc.

.github/workflows/check_variants.yaml
```diff
@@ -5,11 +5,18 @@ on:
   pull_request:
```
sayakpaul, line 5: I don't think there's any dependency on `kernel-builder` changes?

.github/workflows/build_kernel.yaml
danieldk: * Other Rust parts (kernel-abi-check, kernels-data).
* The `kernels` Python package (to validate that `get_kernel` works correctly as a post-build hook).

In other words, pretty much everything.

.github/workflows/build_kernel_macos.yaml
```diff
@@ -5,11 +5,19 @@ on:
...
-    paths-ignore:
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "nix-builder/**"
+      - "kernel-builder/**"
+      - "examples/kernels/**"
+      - "**/*.nix"
+      - "**/flake.lock"
+      - ".github/workflows/build_kernel_macos.yaml"
```
danieldk, line 14: See above.

.github/workflows/build_kernel_rocm.yaml
```diff
@@ -5,11 +5,19 @@ on:
...
-    paths-ignore:
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "nix-builder/**"
+      - "kernel-builder/**"
+      - "examples/kernels/**"
+      - "**/*.nix"
+      - "**/flake.lock"
+      - ".github/workflows/build_kernel_rocm.yaml"
```
danieldk, line 14: See above.

.github/workflows/build_kernel_windows.yaml
```diff
@@ -5,11 +5,19 @@ on:
...
-    paths-ignore:
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "nix-builder/**"
+      - "kernel-builder/**"
+      - "examples/kernels/**"
+      - "**/*.nix"
+      - "**/flake.lock"
+      - ".github/workflows/build_kernel_windows.yaml"
```
danieldk, line 14: Not dependent on nix-builder, apart from that, see above.

.github/workflows/check_variants.yaml
```diff
@@ -5,11 +5,18 @@ on:
...
     types: [opened, synchronize, reopened] # trigger on PRs
-    paths-ignore:
-      - "docs/**"
-      - "*.md"
+    paths:
+      - "nix-builder/**"
+      - "**/*.nix"
+      - "**/flake.lock"
+      - "docs/source/builder/build-variants.md"
+      - ".github/workflows/check_variants.yaml"
```
danieldk, line 13: This check is really cheap, so I think it's better not to run the risk of missing a dependency.

danieldk (commented):
I am not a big fan of doing this. It is easy to miss dependencies except for in very obvious cases (which are typically fast anyway). Also, parts can gain new dependencies inside the tree that are missed and checks do not fire correctly.

As an example, the most expensive cases (building kernels) miss a lot of dependencies in this PR (e.g. the Rust parts, the `kernels` Python package). Once you add them, pretty much every change will trigger them.

I think more is gained by making CI runs faster. I improved the CUDA kernel builds a lot by adding more concurrency. I was planning to do the same for ROCm/CPU/Metal (also opens up the opportunity to wire up some more test kernels for these, e.g. the Torch stable ABI tests).

I think by far the largest issue currently is Windows CI and it would be best to see what is wasting time there and how we can speed it up.

sayakpaul:
It's better for me to close it than to address the feedback if the initial signal is negative. Effort is best put elsewhere.

PR #591, opened 2026-05-27
kernel-abi-check: add support for checking that a kernel does not use non-stable Torch ABI symbols

danieldk:
And wire up in the nix-builder.


kernel-abi-check/kernel-abi-check/src/torch_stable_abi/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+use std::str::FromStr;
+
+use cpp_demangle::Symbol as CppSymbol;
+use eyre::Result;
+use object::{BinaryFormat, ObjectSymbol, Symbol};
+use once_cell::sync::Lazy;
+
+use crate::version::Version;
+
+// https://raw.githubusercontent.com/pytorch/pytorch/refs/heads/main/torch/csrc/stable/c/shim_function_versions.txt
```
sayakpaul, line 11: (nit): Could use the permalink.

kernel-abi-check/kernel-abi-check/src/torch_stable_abi/shim_function_versions.txt
```diff
@@ -0,0 +1,114 @@
+# Auto-generated file listing shim functions and their minimum required versions
```
sayakpaul, line 1: How do we ensure this is also the latest stable?

kernel-abi-check/kernel-abi-check/src/main.rs
```diff
@@ -84,9 +88,41 @@ fn main() -> Result<()> {
...
     }
 
+    if let Some(torch_stable_abi) = args.torch_stable_abi {
+        eprintln!("🔥 Checking for compatibility with Torch stable ABI version {torch_stable_abi}");
+        let torch_stable_abi_violations =
+            check_torch_stable_abi(&torch_stable_abi, file.format(), file.symbols())?;
+        print_torch_stable_abi_violations(&torch_stable_abi_violations, &torch_stable_abi);
+
+        if !torch_stable_abi_violations.is_empty() {
+            return Err(eyre::eyre!("Torch stable ABI compatibility issues found"));
```
sayakpaul, line 98:
```suggestion
            return Err(eyre::eyre!("❌ Torch stable ABI compatibility issues found"));
```

kernel-abi-check/kernel-abi-check/src/main.rs
```diff
@@ -84,9 +88,41 @@ fn main() -> Result<()> {
...
+        } else {
+            eprintln!("✅ No Torch stable ABI compatibility issues found");
+        }
+    }
+
     Ok(())
 }
 
+fn print_torch_stable_abi_violations(
```
sayakpaul, line 107: Ouf, this is superb!

sayakpaul (approved):
Not a merge blocker but would it make sense to add `check_torch_stable_abi()` to the Python bindings?

kernel-abi-check/kernel-abi-check/src/torch_stable_abi/mod.rs
danieldk: I didn't make it a permalink, because we typically want to use the URL to get the latest version and update our copy.

kernel-abi-check/kernel-abi-check/src/torch_stable_abi/shim_function_versions.txt
danieldk: Same question for the manylinux and Python ABI3 files. We should probably have a CI job to check for this and maybe make a PR automatically one one of the files is outdated.

kernel-abi-check/kernel-abi-check/src/main.rs
danieldk: Already printed with an stop sign below.

kernel-abi-check/kernel-abi-check/src/torch_stable_abi/shim_function_versions.txt
sayakpaul: I can look into it after the PR is merged: https://github.com/huggingface/kernels/issues/593

sayakpaul (approved):

PR #596, opened 2026-05-28
Move ABI check to `kernel-builder check-abi`

danieldk:
This change combines several changes for the ABI checker:

* Merge `kernels check` and the `kernel-abi-check` utilities into `kernel-builder check-abi`. Since this is a development tool, it did not belong in `kernels`.
* To make the `check-abi` subcommand similar to other subcommands, take a kernel directory and check the ABI of all variants of the kernel.
* Remove the Python binding, this is now unused and doing so removes a lot of maintenance burden.
* Fix the output of the ABI check a bit.


sayakpaul (approved):
Very nice! Are we missing tests?

danieldk:
> Very nice! Are we missing tests?

Nope, gets tested through the kernel builds. We might add some `.so` files in the future, but we need a way to make very compact ones.

PR #598, opened 2026-05-28
kernels: order no C++ tag before C++11 tag

danieldk:


IlyasMoutawwakil (approved):
LGTM ! thanks @danieldk

PR #599, opened 2026-05-28
Use C++11 ABI tag for a bit longer

danieldk:
To preserve compatibility with older kernels (<0.13) versions.


sayakpaul (approved):
Do the changes in `build_kernel_xpu.yaml` also belong to other builder workflows?

danieldk:
> Do the changes in `build_kernel_xpu.yaml` also belong to other builder workflows?

No, XPU (and macOS) haven't switched to concurrent builds yet. Will make a PR for that.

PR #600, opened 2026-05-29
[ci] add concurrency

sayakpaul:
Add concurrency to the workflow runs and allow runs to be canceled for previous commits.

Previously, we deleted it from the `test_e2e.yaml` to prevent race conditions. But we should be good now because we're using unique repo ids?


danieldk:
Is the Windows failure is related? Do you know why the Python builds got cancelled?

sayakpaul:
Could disable it for the windows CI and investigate in the next PR?

danieldk (dismissed):

sayakpaul:
@danieldk there was a merge conflict which had to be resolved. So, need another approval.

danieldk (approved):

PR #607, opened 2026-06-01
[`FuncRepository`] Add ability of detecting flags

vasqu:
Similar to `LayerRepository` we should allow flags to be passed. In this case it's a bit weird because we attach it to a callable but it does enable torch compile with other functions such as the RoPE kernel.

### DIFF
--- a/kernels/src/kernels/layer/func.py
+++ b/kernels/src/kernels/layer/func.py
@@ -296,6 +296,10 @@ def _create_func_module(func: Callable) -> Type["nn.Module"]:
     from torch import nn
 
     class Func(nn.Module):
+        # Same flags as possible with normal `nn.Module` objects that are exchanged
+        can_torch_compile = getattr(func, "can_torch_compile", False)
+        has_backward = getattr(func, "has_backward", True)
+
         def forward(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


kernels/src/kernels/layer/func.py
```diff
@@ -296,6 +296,10 @@ def _create_func_module(func: Callable) -> Type["nn.Module"]:
     from torch import nn
 
     class Func(nn.Module):
+        # Same flags as possible with normal `nn.Module` objects that are exchanged
+        can_torch_compile = getattr(func, "can_torch_compile", False)
+        has_backward = getattr(func, "has_backward", True)
```
vasqu, line 301: I took the defaults from the getattr checks so this should be BC

danieldk (approved):
Setting these on functions is a bit weird, but I also don't know of a better way, so let's do it!

vasqu:
Haha agree it's a bit weird but imo the cleanest without changing too much

kernels/src/kernels/layer/func.py
sayakpaul: Should these be documented as well? Otherwise, they feel too specific to me.

PR #609, opened 2026-06-02
[core] feat: refactor validate dependency

sayakpaul:
This PR refactors how we validate dependencies:

* Fetches the `metadata.json` before doing the expensive `snapshot_download()`. This allows us to error out early so that the user doesn't have to wait for the error _after_ the build variant download is completed.
* Propagates related changes to `get_local_kernel()` as well.

### DIFF
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -235,6 +235,7 @@ def install_kernel(
     backend: str | None = None,
     variant_locks: dict[str, VariantLock] | None = None,
     user_agent: str | dict | None = None,
+    validate_dependencies: bool = False,
 ) -> Path:
     """
     Download a kernel for the current environment to the cache.
@@ -255,6 +256,9 @@ def install_kernel(
             Optional dictionary of variant locks for validation.
         user_agent (`Union[str, dict]`, *optional*):
             The `user_agent` info to pass to `snapshot_download()` for internal telemetry.
+        validate_dependencies (`bool`, defaults to False):
+            When set to True, performs dependency validation. Useful for kernels that have
+            extra Python dependencies.
 
     Returns:
         `Path`: The path to the variant directory.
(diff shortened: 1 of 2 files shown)


kernels/src/kernels/utils.py
```diff
@@ -277,6 +281,17 @@ def install_kernel(
...
+    # Validate Python dependencies before downloading the variant.
+    metadata_path = api.hf_hub_download(
+        repo_id,
+        repo_type="kernel",
+        filename=f"build/{variant.variant_str}/metadata.json",
+        cache_dir=CACHE_DIR,
+        revision=revision,
+        local_files_only=False,
+    )
+    _validate_variant_dependencies(Path(metadata_path).parent)
```
danieldk: This can't be in `install_kernel`, because it's also used to predownload kernels to e.g. bake into Docker containers (e.g. when using kernel locking), so we cannot assume that it's run in the final environment where it has access to the final set of dependencies.

kernels/src/kernels/utils.py
```diff
@@ -354,7 +369,9 @@ def _resolve_local_variant_path(
             )
         )
     )
-    return _find_kernel_in_repo_path(repo_path, variant=variant, variant_locks=variant_locks)
+    variant_path = _find_kernel_in_repo_path(repo_path, variant=variant, variant_locks=variant_locks)
```
danieldk: Also can't be in these functions, since we want to support getting a kernel directory without actually loading it.

danieldk (commented):
I am not sure what the solution is without downloading, but the dependency verification needs to be at load/import time. The issue is that downloads, getting kernel paths, etc. can be done for other reasons (e.g. predownloading a kernel, populating a Docker container, kernel validation, etc.) where we cannot assume that the dependencies are in the environment (yet).

kernels/src/kernels/utils.py
danieldk: Thinking more about it, maybe in `get_kernel`/`get_local_kernel`/`get_locked_kernel`, since these are the two functions that actually end up loading kernels and in `get_kernel` we can put it before any other downloads?

Or maybe even better and closer to your PR, we can add a `validate_dependencies=False` option to `install_kernel` and then set it to `True` in the calls where we need validation?

sayakpaul:
I can remove them from the functions you mentioned but I would like to discuss the after steps before I do so. Specifically, for stuff like

> I am not sure what the solution is without downloading, but the dependency verification needs to be at load/import time. The issue is that downloads, getting kernel paths, etc. can be done for other reasons (e.g. predownloading a kernel, populating a Docker container, kernel validation, etc.)

We cannot assume the users are ready to use them already and hence, it would not make sense to hard raise on validation. So, I propose the following:

* Log a warning to the users that they don't have the requirements installed or
* We do it exclusively from `get_kernel()` since it does more than _just_ downloading a kernel.

I prefer the second as warnings tend to get ignored.

kernels/src/kernels/utils.py
sayakpaul: Just [commented](https://github.com/huggingface/kernels/pull/609#issuecomment-4600898886) with something similar.

danieldk (approved):
Nice!!!

PR #612, opened 2026-06-02
support capabilities in no-arch

### LINKED ISSUE #465: Include capabilities for noarch kernels in build.toml and export to metadata

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/465


docs/source/builder/writing-kernels.md
```diff
@@ -206,6 +206,7 @@ The following sections enumerate all supported options for `build.toml`.
 
 ### `general.cuda`
 
+- `capabilities` (optional): a list of CUDA compute capabilities.
```
danieldk: I think we have to put it somewhere else that is specific to `torch-noarch` kernels, since we already have a way of setting capabilities in arch kernels. It would be confusing to put it here. What would it mean for arch kernels? People might expect to set it on all kernel components, etc.

kernel-builder/src/pyproject/common.rs
```diff
@@ -46,7 +46,7 @@ pub fn write_metadata(
             upstream: general.upstream.clone(),
             python_depends,
             backend: BackendInfo {
-                archs: None,
+                archs: general.backend_archs(*backend).map(<[String]>::to_vec),
```
danieldk: It is already a `Vec<String>`, cloning should be enough.

kernel-builder/src/pyproject/common.rs
```diff
@@ -84,3 +84,75 @@ pub fn write_cmake_file(file_set: &mut FileSet, filename: &str, content: &[u8])
...
+mod tests {
+    use kernels_data::config::{Build, CurrentConfig};
+    use kernels_data::metadata::Metadata;
+
+    use super::write_metadata;
+    use crate::pyproject::fileset::FileSet;
+    use crate::pyproject::ops_identifier::KernelIdentifier;
+
+    fn metadata_for(toml: &str, backend: &str) -> Metadata {
+        let build: Build = toml::from_str::<CurrentConfig>(toml).unwrap().into();
```
danieldk: Why go through the toml parsing? We can rely on the `serde-toml` crate to have tests for that.

kernels-data/src/config/mod.rs
```diff
@@ -134,9 +135,21 @@ impl General {
...
+                .as_ref()
+                .and_then(|cuda| cuda.capabilities.as_deref()),
+            Backend::Rocm => self.rocm.as_ref().and_then(|rocm| rocm.archs.as_deref()),
+            _ => None,
+        }
+    }
 }
 
 pub struct CudaGeneral {
+    pub capabilities: Option<Vec<String>>,
```
danieldk: This should be specific to noarch. arch kernels already have a way of encoding capabilities.

kernel-builder/src/pyproject/common.rs
```diff
@@ -24,6 +24,28 @@ pub fn write_metadata(
...
+    noarch: &TorchNoarch,
+    kernel_id: &KernelIdentifier,
+    file_set: &mut FileSet,
+) -> Result<()> {
+    write_metadata_impl(general, Some(noarch), kernel_id, file_set)
+}
+
+fn write_metadata_impl(
+    general: &General,
+    noarch: Option<&TorchNoarch>,
```
danieldk, line 46: Not very keen on this, because it adds noarch-specific things to agnostic/common functions.

I think we should keep it like this for this PR, but I have to think a bit about it. I like a certain amount of parallelism between arch/noarch, but I am not sure yet how to do that nicely (maybe by writing the bits to `setup.py` and extending the metadata from there).

Generally, it becomes hard to understand how the generation works if for different frameworks (torch, torch-noarch, tvm-ffi) we do things in different phases (create-pyproject, configure, build, install). So even though it might be slightly less elegant to update the metadata from Python, it would be clean because it's the same between all cases and we don't need different metadata writing paths.

I can do a follow-up PR once I have a clearer idea of what it should look like.

danieldk (approved):

kernel-builder/src/pyproject/common.rs
sayakpaul: Thanks for this insight. I agree with that. I actually want to hold off on merging this PR now that I have a better picture from you. It leads to an unpleasant solution that we might have to discard anyway. I will also think it over a bit and see if I can repurpose this PR.

sayakpaul:
Closing in favor of https://github.com/huggingface/kernels/pull/668.

PR #618, opened 2026-06-04
kernel-builder: add `hash` subcommand and hook up in Nix

danieldk:
This change adds a `hash` subcommand to `kernel-builder` that computes the hashes of all kernel files and stores it in `metadata.json`. The hash of the metadata itself is not computed (since it could only be hashed without the hashes stored) and will be protected by a detached signature.

Also hook up this command in the Nix builder, so that all kernels will have file hashes going forward.

Python bindings will follow in later PRs (done already, but I'm breaking it up in smaller, reviewable parts).

### DIFF
--- a/docs/source/kernel-requirements.md
+++ b/docs/source/kernel-requirements.md
@@ -73,9 +73,9 @@ metadata. Currently the following top-level keys are supported:
   list of [supported license identifiers](https://huggingface.co/docs/hub/repositories-licenses).
 - `backend` (`dict`, required): information about the compute backend that
   this build variant supports.
+- `digest` (`Digest`, required): hash digest of the kernel files.
 - `python-depends` (`list[str]`, optional): list of Python dependencies
   from a curated set of Python dependencies.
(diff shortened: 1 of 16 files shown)


docs/source/kernel-requirements.md
```diff
@@ -75,6 +75,7 @@ metadata. Currently the following top-level keys are supported:
   this build variant supports.
 - `python-depends` (`list[str]`, optional): list of Python dependencies
   from a curated set of Python dependencies.
+- `digest` (`Digest`, required): hash digest of the kernel files.
```
sayakpaul: (nit): required dependencies could come before optionals.

sayakpaul (dismissed):
Beautiful. Such a clean PR!

docs/source/kernel-requirements.md
danieldk: Good idea, fixed.

sayakpaul (approved):
🌪️

PR #620, opened 2026-06-04
[`LayerRepo`] Allow `ext` to be a source of layers

vasqu:
As per title, currently we only allow `layers` to expose layer repositories but we want to change conventions to allow external modifications for others like transformers

### DIFF
--- a/kernels/src/kernels/layer/layer.py
+++ b/kernels/src/kernels/layer/layer.py
@@ -386,10 +386,12 @@ def kernelize_layer(module: "nn.Module", *, mode: Mode, device_type: Device, use
 def _get_kernel_layer(repo: LayerRepositoryProtocol, kernel: ModuleType) -> Type["nn.Module"]:
     """Get a layer from a kernel."""
 
-    if getattr(kernel, "layers", None) is None:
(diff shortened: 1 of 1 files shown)


kernels/src/kernels/layer/layer.py
```diff
@@ -386,10 +386,12 @@ def kernelize_layer(module: "nn.Module", *, mode: Mode, device_type: Device, use
 def _get_kernel_layer(repo: LayerRepositoryProtocol, kernel: ModuleType) -> Type["nn.Module"]:
     """Get a layer from a kernel."""
 
-    if getattr(kernel, "layers", None) is None:
-        raise ValueError(f"Kernel repo {repo} does not define any layers.")
+    kernel_layers = getattr(kernel, "layers", None)
```
danieldk, line 389: No, we shouldn't do this. All kernel layers go into `layers`. They are already custom to Hub kernels by definition upstream kernels do not have kernel layers except by accident (e.g. parameterless activation `Modules`).

vasqu:
Closing because of my misunderstanding sorry!

PR #622, opened 2026-06-05
Simplify handling of kernelized functions

danieldk:
This is a set of related changes based on internal discussions:

- Deprecate the `FuncRepository` classes. Kernel-side functions are not a great interface, because one typically has to set attributes to permit `torch.compile` and backward passes. However for functions, these are detached. It also hides that the functions are converted to `nn.Module` underneath.
- Deprecate the `use_kernel_from_hub` decorator and extend `use_kernel_forward_from_hub` to provide the same functionality. This makes it clearer that when decorating a function, it gets replaced by a module instance.
- Adopt `use_kernelized_func` from transformers. This annotation makes it unnecessary to assign functions as member variables, which cause issues in some applications. Instead, they are registered in a module separate from (sub-)modules. `kernelize` is updated to pick up these attached functions as well.

### DIFF
--- a/docs/source/layers.md
+++ b/docs/source/layers.md
@@ -64,7 +64,7 @@ implementation of `__call__` that delegates to `forward`.
 
 For [`~kernels.kernelize`] to see the function, you must use
 [`~kernels.use_kernelized_func`] on an `torch.nn.Module` that is part
-of the to-be kernlized module to make the function visible to
+of the to-be kernlized model to make the function visible to
 [`~kernels.kernelize`]. The function is typically attached to the module
 that uses it. For example:
 
@@ -77,6 +77,9 @@ class FeedForward(nn.Module):
   def forward(self, x: torch.Tensor) -> torch.Tensor:
       return silu_and_mul(self.linear(x))
 ```
+
+Functions used by [`~kernels.use_kernelized_func`] must always have a
+[`~kernels.use_kernel_forward_from_hub`] decorator.
 
 ## Kernelizing a model
 
--- a/docs/source/layers.md
+++ b/docs/source/layers.md
@@ -62,7 +62,7 @@ This will replace the function by an instantiated `torch.nn.Module`
 still behave as a function, since `torch.nn.Module` provides an
 implementation of `__call__` that delegates to `forward`.
 
-For kernelization to see the function, you must use
+For [`~kernels.kernelize`] to see the function, you must use
 [`~kernels.use_kernelized_func`] on an `torch.nn.Module` that is part
 of the to-be kernlized module to make the function visible to
 [`~kernels.kernelize`]. The function is typically attached to the module
--- a/docs/source/layers.md
+++ b/docs/source/layers.md
@@ -47,30 +47,35 @@ compatible with layers from the hub.
 
 Sometimes it can be useful to make a function extensible, for example
 because the function cannot be replaced by a layer. In such cases, you
-can annotate the function with the [`~kernels.use_kernel_func_from_hub`] decorator:
+can use the [`~kernels.use_kernel_forward_from_hub`] decorator on a
+function:
 
 ```python
-@use_kernel_func_from_hub("silu_and_mul")
+@use_kernel_forward_from_hub("silu_and_mul")
 def silu_and_mul(x: torch.Tensor) -> torch.Tensor:
     d = x.shape[-1] // 2
     return F.silu(x[..., :d]) * x[..., d:]
 ```
 
 This will replace the function by an instantiated `torch.nn.Module`
-(singleton) that calls the function itself in its forward method.
+(singleton) that calls the function itself in its forward method. It will
+still behave as a function, since `torch.nn.Module` provides an
+implementation of `__call__` that delegates to `forward`.
 
-**Note:** for kernelization to see the function, it must be a member of
-another `torch.nn.Module` that is part of the model. For example:
+For kernelization to see the function, you must use
+[`~kernels.use_kernelized_func`] on an `torch.nn.Module` that is part
+of the to-be kernlized module to make the function visible to
+[`~kernels.kernelize`]. The function is typically attached to the module
+that uses it. For example:
 
 ```python
+@use_kernelized_func(silu_and_mul)
 class FeedForward(nn.Module):
   def __init__(self, in_features: int, out_features: int):
       self.linear = nn.Linear(in_features, out_features)
-      # Note: silu_and_mul is a Torch module.
-      self.silu_and_mul = silu_and_mul
 
   def forward(self, x: torch.Tensor) -> torch.Tensor:
-      return self.silu_and_mul(self.linear(x))
+      return silu_and_mul(self.linear(x))
 ```
 
 ## Kernelizing a model
--- a/kernels/src/kernels/layer/func.py
+++ b/kernels/src/kernels/layer/func.py
@@ -27,8 +27,9 @@ class FuncRepository:
     """
     Repository and name of a function for kernel mapping.
 
-    **Warning**: `FuncRepository` is deprecated and will be removed in kernels
-    0.17. Use [`~kernels.LayerRepository`] instead.
+    > [!WARNING]
+    > `FuncRepository` is deprecated and will be removed in kernels 0.17.
+    > Use [`~kernels.LayerRepository`] instead.
 
     Args:
         repo_id (`str`):
@@ -135,8 +136,9 @@ class LocalFuncRepository:
     """
     Repository and function name from a local directory for kernel mapping.
 
-    **Warning**: `LocalFuncRepository` is deprecated and will be removed in kernels
-    0.17. Use [`~kernels.LocalLayerRepository`] instead.
+    > [!WARNING]
+    > `LocalFuncRepository` is deprecated and will be removed in kernels 0.17.
+    > Use [`~kernels.LocalLayerRepository`] instead.
 
     Args:
         repo_path (`Path`):
@@ -203,8 +205,9 @@ def use_kernel_func_from_hub(func_name: str):
     kernelized, it **must** be a member of another `torch.nn.Module` that is
     part of the model (see the example).
 
-    **Warning**: `use_kernel_func_from_hub` is deprecated and will be removed in kernels
-    0.17. Use [`~kernels.use_kernel_forward_from_hub`] instead.
+    > [!WARNING]
+    > `use_kernel_func_from_hub` is deprecated and will be removed in kernels 0.17.
+    > Use [`~kernels.use_kernel_forward_from_hub`] instead.
 
     Args:
         func_name (`str`):
@@ -256,8 +259,9 @@ class LockedFuncRepository:
     In contrast to `FuncRepository`, this class uses repositories that
     are locked inside a project.
 
-    **Warning**: `LockedFuncRepository` is deprecated and will be removed in kernels
-    0.17. Use [`~kernels.LockedLayerRepository`] instead.
+    > [!WARNING]
+    > `LockedFuncRepository` is deprecated and will be removed in kernels 0.17.
+    > Use [`~kernels.LockedLayerRepository`] instead.
 
     Args:
         repo_id (`str`): The Hub repository containing the function.
--- a/kernels/src/kernels/layer/func.py
+++ b/kernels/src/kernels/layer/func.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+import warnings
 from inspect import Parameter, Signature
 from pathlib import Path
 from types import ModuleType
@@ -27,6 +28,9 @@ class FuncRepositoryProtocol(RepositoryProtocol, Protocol):
 class FuncRepository:
     """
     Repository and name of a function for kernel mapping.
+
+    **Warning**: `FuncRepository` is deprecated and will be removed in kernels
+    0.17. Use [`LayerRepository`] instead.
 
     Args:
         repo_id (`str`):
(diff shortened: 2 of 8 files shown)


docs/source/layers.md
```diff
@@ -47,30 +47,35 @@ compatible with layers from the hub.
...
-**Note:** for kernelization to see the function, it must be a member of
-another `torch.nn.Module` that is part of the model. For example:
+For kernelization to see the function, you must use
+[`~kernels.use_kernelized_func`] on an `torch.nn.Module` that is part
+of the to-be kernlized module to make the function visible to
+[`~kernels.kernelize`]. The function is typically attached to the module
+that uses it. For example:
 
 ```python
+@use_kernelized_func(silu_and_mul)
```
vasqu, line 72: Nice docs 🫡

kernels/src/kernels/layer/kernelize.py
```diff
@@ -258,10 +258,18 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
 
     for _, module in model.named_modules():
         module_class = type(module)
-        if not hasattr(module_class, "kernel_layer_name"):
-            continue
 
-        kernelize_layer(module, mode=mode, device_type=device_type, use_fallback=use_fallback)
+        # Kernel functions attached to a module through `use_kernelized_func`.
+        for kernel_func in getattr(module, "_kernel_funcs", {}).values():
```
vasqu, line 263: Just for BC, I would like to keep the same original naming as in transformers `_hidden_kernels`

kernels/src/kernels/layer/layer.py
```diff
@@ -505,3 +595,27 @@ def _get_layer_memoize(repo: RepositoryProtocol, module_class: Type["nn.Module"]
     _CACHED_LAYER[repo] = layer
 
     return layer
+
+
+def _create_func_module(func: Callable) -> Type["nn.Module"]:
```
vasqu, line 607: Don't know how import it is for you guys but should we keep a BC reference in func.py for now?

kernels/src/kernels/layer/kernelize.py
```diff
@@ -258,10 +258,18 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            kernelize_layer(
+                kernel_func,
+                mode=mode,
+                device_type=device_type,
+                use_fallback=use_fallback,
+            )
```
vasqu, line 269: I think we should detach them afterwards or at least add this as an optional flag to enable this, wdyt?

vasqu (commented):
My initial comments, I still need to test this on transformers side but it definitely would make our lives easier ❤️ only thing I'm personally missing is the option to also remove the hidden modules, we kernelize once tbh and should not allow rekernelizing (for now)

kernels/src/kernels/layer/kernelize.py
danieldk: I think it's better to give it a new name, or if would conflict with the existing implementation in transformers. `_kernel_funcs` also seems to be a more descriptive name?

kernels/src/kernels/layer/layer.py
danieldk: We only consider things exported from the top-level `__init__.py` stable API, so we can move things around freely in other places.

kernels/src/kernels/layer/kernelize.py
danieldk: Detaching sounds like a bad idea, because then you cannot kernelize the same model multiple times, which is a scenario that we want to support. (E.g. for kernelizing with different mappings or kernelizing with a different mode.)

kernels/src/kernels/layer/kernelize.py
vasqu: Hmm, yea should be fine as well. I see the point :+1:

kernels/src/kernels/layer/layer.py
vasqu: That's neat

kernels/src/kernels/layer/kernelize.py
vasqu: I have to check whether deepspeed still plays well with this. Not sure if it still takes a look at these hidden dicts, it shouldn't but we never know :sweat:

danieldk: 🤞

docs/source/layers.md
```diff
@@ -47,30 +47,35 @@ compatible with layers from the hub.
...
 
 This will replace the function by an instantiated `torch.nn.Module`
-(singleton) that calls the function itself in its forward method.
+(singleton) that calls the function itself in its forward method. It will
+still behave as a function, since `torch.nn.Module` provides an
+implementation of `__call__` that delegates to `forward`.
 
-**Note:** for kernelization to see the function, it must be a member of
-another `torch.nn.Module` that is part of the model. For example:
+For kernelization to see the function, you must use
```
sayakpaul:
```suggestion
For [`~kernels.kernelize`] to see the function, you must use
```

docs/source/layers.md
sayakpaul, line 72: Does `silu_and_mul` need to _always_ have the `use_kernel_forward_from_hub` decorator here?

(wasn't clear from the docs, hence asking)

kernels/src/kernels/layer/func.py
```diff
@@ -92,7 +100,9 @@ def _resolve_revision(self) -> str:
 
     def load(self) -> Type["nn.Module"]:
         kernel = get_kernel(
-            self._repo_id, revision=self._resolve_revision(), trust_remote_code=self._trust_remote_code
+            self._repo_id,
+            revision=self._resolve_revision(),
+            trust_remote_code=self._trust_remote_code,
```
sayakpaul, line 106: (nit): Seems like a formatting-related change?

kernels/src/kernels/layer/func.py
```diff
@@ -28,6 +27,9 @@ class FuncRepository:
     """
     Repository and name of a function for kernel mapping.
 
+    **Warning**: `FuncRepository` is deprecated and will be removed in kernels
+    0.17. Use [`~kernels.LayerRepository`] instead.
```
sayakpaul: (nit): since it appears on our docs, we could even use the "> [!WARNING]" block to contain this. It'd appear like so:

Image

kernels/src/kernels/layer/kernelize.py
sayakpaul: The ability to detach them _through a flag_ doesn't sound so bad, though.

kernels/src/kernels/layer/layer.py
```diff
@@ -302,11 +310,93 @@ def forward(self, x: torch.Tensor):
 
         # The layer can now be kernelized:
         # model = kernelize(model, mode=Mode.TRAINING | Mode.TORCH_COMPILE, device="cuda")
+
+        # Use on a function (converts the function to `nn.Module`):
```
sayakpaul: Perhaps, also mention what "MyCustomLayer" means in the following line?

kernels/src/kernels/layer/layer.py
```diff
@@ -302,11 +310,93 @@ def forward(self, x: torch.Tensor):
...
+    Example:
+        ```python
+        import torch
+        import torch.nn as nn
+
+        from kernels import use_kernel_forward_from_hub
+        from kernels import use_kernelized_func
+        from kernels import Mode, kernelize
+
+        # Use on a function (converts the function to `nn.Module`):
```
sayakpaul: Same as above.

kernels/src/kernels/layer/layer.py
```diff
@@ -302,11 +310,93 @@ def forward(self, x: torch.Tensor):
...
+
     def decorator(cls):
-        replace_kernel_forward_from_hub(cls, layer_name)
+        orig_init = cls.__init__
+
+        def new_init(self, *args, **kwargs):
+            orig_init(self, *args, **kwargs)
+
+            # Register new function as non-submodule within the modules dict
+            hidden_kernels = self.__dict__.setdefault("_kernel_funcs", {})
```
sayakpaul, line 399: Would it be a bad idea to hold "_kernel_funcs" and "kernel_layer_name" as constants as I see them being referred from two different sites?

kernels/src/kernels/layer/layer.py
```diff
@@ -505,3 +595,27 @@ def _get_layer_memoize(repo: RepositoryProtocol, module_class: Type["nn.Module"]
...
     return layer
+
+
+def _create_func_module(func: Callable) -> Type["nn.Module"]:
+    from torch import nn
+
+    class Func(nn.Module):
+        # Same flags as possible with normal `nn.Module` objects that are exchanged
+        can_torch_compile = getattr(func, "can_torch_compile", False)
+        has_backward = getattr(func, "has_backward", True)
```
sayakpaul, line 613: Didn't see these two flags documented.

kernels/src/kernels/layer/func.py
```diff
@@ -210,13 +240,13 @@ def forward(self, x):
         # model = kernelize(model, mode=Mode.TRAINING | Mode.TORCH_COMPILE, device="cuda")
         ```
     """
+    warnings.warn(
+        "use_kernel_func_from_hub is deprecated and will be removed in kernels 0.17. Use [`use_kernel_func_from_hub`] instead.",
```
sayakpaul:
```suggestion
        "use_kernel_func_from_hub is deprecated and will be removed in kernels 0.17. Use [`use_kernel_forward_from_hub`] instead.",
```

kernels/src/kernels/layer/layer.py
```diff
@@ -302,11 +310,93 @@ def forward(self, x: torch.Tensor):
...
+
+            # Register new function as non-submodule within the modules dict
+            hidden_kernels = self.__dict__.setdefault("_kernel_funcs", {})
+            for fn in decorator_args:
+                name = getattr(fn, "__name__", None) or getattr(fn, "kernel_layer_name", None)
+                if name is None:
+                    raise ValueError(f"Could not infer kernel function name for {fn!r}")
+
+                # Do not register as submodule! Hide it behind a dict to be removed later after registering it
+                hidden_kernels[name] = fn
```
sayakpaul: Should this behaviour be tested?

kernels/src/kernels/layer/layer.py
```diff
@@ -302,11 +310,93 @@ def forward(self, x: torch.Tensor):
...
+    def decorator(ty):
+        if inspect.isfunction(ty):
+            Func = _create_func_module(ty)
+            replace_kernel_forward_from_hub(Func, layer_name)
+            return Func()
+        elif inspect.isclass(ty):
+            replace_kernel_forward_from_hub(ty, layer_name)
+            return ty
+        else:
+            raise TypeError("@use_kernel_forward_from_hub can only be applied to classes or functions")
```
sayakpaul, line 337: Do we test for the error path? If not, maybe worth testing it? Okay, if not in this PR.

kernels/src/kernels/layer/layer.py
```diff
@@ -302,11 +310,93 @@ def forward(self, x: torch.Tensor):
...
+
+        def new_init(self, *args, **kwargs):
+            orig_init(self, *args, **kwargs)
+
+            # Register new function as non-submodule within the modules dict
+            hidden_kernels = self.__dict__.setdefault("_kernel_funcs", {})
+            for fn in decorator_args:
+                name = getattr(fn, "__name__", None) or getattr(fn, "kernel_layer_name", None)
+                if name is None:
+                    raise ValueError(f"Could not infer kernel function name for {fn!r}")
```
sayakpaul: This path could be tested.

sayakpaul (commented):
Thanks for shipping this! My comments are mostly test-related. That doesn't have to block this PR, we can ship them in a next PR. Additionally, I think we should probably start testing deprecation warnings are properly raised.

kernels/src/kernels/layer/kernelize.py
danieldk: Why would you want to do this? I am trying to understand the use case. And why for kernel functions and not kernel layers? If you don't want a function to be kernelized, then you can kernelize it with a mapping that does not have a layer for that function. Having two ways to do the same thing sounds really confusing? Plus it would get permanently detached.

sayakpaul: > If you don't want a function to be kernelized, then you can kernelize it with a mapping that does not have a layer for that function.

I was missing this aspect. Thanks for explaining.

docs/source/layers.md
danieldk: Clarified and added check + test.

kernels/src/kernels/layer/layer.py
danieldk: It's in `kernel-requirements.md`.

kernels/src/kernels/layer/layer.py
danieldk: That's a stale comment that came from transformers. Irrelevant to `kernels`, so removed it.

kernels/src/kernels/layer/layer.py
danieldk: Cannot fail anymore due to an earlier check that I added (which is tested), so I have changed this to an assertion.

kernels/src/kernels/layer/layer.py
danieldk: Probably not worth it? It's like type checking in Rust :).

kernels/src/kernels/layer/layer.py
danieldk: Not sure, we'd get something like

```
_KERNEL_LAYER_NAME_ATTR = "kernel_layer_name"
_KERNEL_FUNCS_ATTR = "_kernel_funcs"
```

which is not that great. Also, we cannot 'constantize' all cases, since in some cases we'll do `foo.kernel_layer_name` and replacing them by `getattrs` seems to make the code harder to read?

kernels/src/kernels/layer/func.py
danieldk: Fixed.

sayakpaul (approved):
Let's go!

vasqu (approved):
Sorry for the delays, finally got around it. See https://github.com/huggingface/transformers/pull/46883 for more details

Also checked deepspeed and it had no issues 🥳

PR #623, opened 2026-06-05
Improve security of handling pyc files

danieldk:
We cannot consider pyc files in digest, because they are generated by the interpreter and are not static. However, a blanket ignore of pyc files in hashing can allow an attacker to smuggle in a malicious pyc file. Tighten up security in the following way:

- Never download `.pyc` files or `__pycache__` directories. This ensures that we are only downloading files that are covered by the digest.
- When hashing files, exclude `__pycache__`, since these are locally generated.
- When hashing files, do not exclude `pyc` files, since the Python interpreter does not generate them outside `__pycache__`. The first mitigation should cover this case, but let's include them in hashing to be extra paranoid.

### DIFF
--- a/kernels-data/src/digest.rs
+++ b/kernels-data/src/digest.rs
@@ -30,8 +30,8 @@ impl Digest {
                 // hashes. It is protected by the detached signature.
                 || entry.path().file_name().and_then(|f| f.to_str()) == Some("metadata.json")
                 || entry.path().file_name().and_then(|f| f.to_str()) == Some("metadata.json.sigstore")
-                // Python likes to create .pyc files in the cache.
-                || entry.path().extension().and_then(|e| e.to_str()) == Some("pyc")
+                // Python likes to create .pyc files in __pycache__/ directories.
+                || entry.path().components().any(|c| c.as_os_str() == "__pycache__")
             {
                 continue;
             }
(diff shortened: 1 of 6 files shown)


kernels-data/src/digest.rs
```diff
@@ -124,18 +124,28 @@ mod tests {
...
+        let pycache_dir = dir.path().join("pkg").join("__pycache__");
+        fs::create_dir_all(&pycache_dir)?;
         fs::write(dir.path().join("_extension.so"), b"content")?;
         fs::write(dir.path().join("metadata.json"), b"{}")?;
         fs::write(dir.path().join("metadata.json.sigstore"), b"sig")?;
-        fs::write(dir.path().join("cache.pyc"), b"bytecode")?;
+        fs::write(pycache_dir.join("mod.cpython-311.pyc"), b"bytecode")?;
+        // A .pyc outside __pycache__/ must be included — it was not generated
+        // locally by Python and should be covered by the digest.
+        fs::write(dir.path().join("payload.pyc"), b"smuggled")?;
```
sayakpaul, line 135: "smuggled" 🤪

sayakpaul (dismissed):

sayakpaul (approved):

PR #625, opened 2026-06-08
Unify bytecode handling across downloads/builds/hashing

danieldk:


sayakpaul (commented):
Thanks! Are we missing tests?

danieldk:
> Thanks! Are we missing tests?

I think the only case where the behavior has changed (hashing) has adjusted tests. Kernel locking is already covered (the repo has .pyc files).

sayakpaul (approved):

PR #626, opened 2026-06-08
fix card filling to include version

sayakpaul:
`version` should not be missing in the Hub kernel cards: 

* https://huggingface.co/kernels/kernels-community/flash-attn-ops
* https://huggingface.co/kernels/drbh/yamoe


kernel-builder/src/card.rs
```diff
@@ -102,21 +102,24 @@ fn extract_layers(kernel_dir: &Path, module_name: &str) -> Option<Vec<String>> {
...
 
-fn render_card(build: &Build, kernel_dir: &Path) -> Result<String> {
-    let card_template_path = kernel_dir.join("CARD.md");
-    if !card_template_path.exists() {
-        bail!(
-            "CARD.md template not found at `{}`",
-            card_template_path.display()
-        );
-    }
+const CARD_TEMPLATE: &str = include_str!("init/templates/CARD.md");
```
danieldk, line 105: This looks wrong. We should render the card template that is in the repo, not the init template.

sayakpaul: Then I am not sure why kernels (example https://huggingface.co/kernels/drbh/yamoe) built with the latest version of the builder is rendering an example usage that doesn't have `version`.

sayakpaul: One possibility could be that the card was copied from elsewhere and that version didn't have `version` in the `get_kernel()` example usage. 

With the current state of things (`main`), I think it will keep on rendering it without `version`. Should there then be any detection of the presence of `version` and if it's not present, it's injected? But that seems fragile to me.

danieldk: > Then I am not sure why kernels (example https://huggingface.co/kernels/drbh/yamoe) built with the latest version of the builder is rendering an example usage that doesn't have version.

Because the card template in the source repo does not use the version that is passed template:

https://github.com/drbh/yamoe/blob/main/CARD.md

For kernels-community we should do this:

https://github.com/huggingface/kernels-community/pull/937

We cannot just switch to the init template, because some kernels may have added extra information to the `CARD.md` in their repo (like extra sections with information, tutorials, acknowledgments, etc). Kernel developers should be able to customize what is put in the card and that won't work anymore if we render the init template.

sayakpaul:
Closing in favor of https://github.com/huggingface/kernels-community/pull/937/. Much better solution. Thanks.

PR #627, opened 2026-06-10
Add acknowledgment for the xpu-kernels skill / Xe-Forge

jiqing-feng:
## Summary

The `cpu-kernels` skill adapted its overall methodology — the YAML knowledge
base, the benchmark/validation harnesses, and the branching trial-manager
optimization loop — from the existing `xpu-kernels` skill (built by the Intel
XPU kernels team), which is itself based on [Xe-Forge](https://github.com/IntelLabs/Xe-Forge).
This PR adds proper attribution so the original authors get visibility.

## Changes

- `kernel-builder/skills/cpu-kernels/SKILL.md`
  - Add an **Acknowledgments** section crediting the `xpu-kernels` skill and
    Xe-Forge as the origin of the workflow methodology.
  - Add the `xpu-kernels` skill and Xe-Forge to **External Resources**.

## Notes

The CPU-specific content (AVX2/AVX512 intrinsics, runtime dispatch, brgemm/AMX
integration, NUMA and threading discipline) remains original to this skill; the
acknowledgment is scoped to the shared workflow methodology.

### DIFF
--- a/kernel-builder/skills/cpu-kernels/SKILL.md
+++ b/kernel-builder/skills/cpu-kernels/SKILL.md
@@ -458,4 +458,4 @@ cpu-kernels/
 
 ## Acknowledgments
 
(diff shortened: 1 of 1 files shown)


danielfleischer (commented):
LGTM, thanks!

sayakpaul (approved):

PR #630, opened 2026-06-11
skills: update cuda skills to make the generated kernels compliant with the builder

sayakpaul:
As discussed internally. 

There were several issues with the CUDA skills instructions. For example, it cannot compile with the builder as it uses pybind.

I have checked on my Spark that with the updated instructions, the kernels can successfully build and be consumed with `get_local_kernel()`.

It also uses the latest `kernel-builder` subcommands.


kernel-builder/skills/cuda-kernels/references/a100-optimization-guide.md
```diff
@@ -300,12 +309,12 @@ ncu --set full -o a100_metrics.ncu-rep python your_script.py
...
 
 ```bash
-cd examples/ltx_video
+cd <your-kernel-project>
 
-# Build for A100
-# Ensure build.toml includes cuda-capabilities = ["8.0"]
-uv pip install -e .
+# Leave cuda-capabilities unspecified in build.toml unless the kernel
+# truly requires specific architectures (A100 is sm_80).
```
danieldk:
```suggestion
# truly requires specific architectures (A100 is `cuda-capabilities = ["8.0"]`).
```

I guess this would provide a more direct hint of what should be in `build.toml` in this case (linking it to other documentation).

kernel-builder/skills/cuda-kernels/references/diffusers-integration.md
```diff
@@ -306,9 +306,9 @@ See SKILL.md "Common Issues and Solutions" for:
 For a self-contained, runnable example that demonstrates all patterns above:
 
 ```bash
-cd examples/ltx_video
-uv pip install -e .  # Build kernels
-python ../../.claude/skills/h100-diffusers-kernels/references/ltx_kernel_injection_example.py
+cd <your-kernel-project>     # kernel-builder project with your kernels
+nix run .#build-and-copy -L  # Build kernels with kernel-builder
+python path/to/skills/cuda-kernels/scripts/ltx_kernel_injection_example.py
```
danieldk, line 311: Is this run command still correct?

kernel-builder/skills/cuda-kernels/references/huggingface-kernels-integration.md
```diff
@@ -34,22 +34,20 @@ Download and load a kernel from the Hub:
...
-# Basic usage
-kernel = get_kernel("kernels-community/activation")
-
-# With specific version
+# Load a specific major version (the standard way). A bare
+# get_kernel(repo_id) raises ValueError — version= or revision= is required.
 kernel = get_kernel("kernels-community/activation", version=1)
 
-# With specific revision (branch/tag/commit)
+# Or pin an explicit revision (branch/tag/commit)
```
danieldk:
```suggestion
# Or pin an explicit revision (branch/tag/commit). This is for exceptional cases, using `version` is strongly recommended.
```

kernel-builder/skills/cuda-kernels/references/huggingface-kernels-integration.md
```diff
@@ -133,7 +140,7 @@ print(dir(flash_attn))
 import torch
 from kernels import get_kernel
 
-layer_norm = get_kernel("kernels-community/triton-layer-norm")
+layer_norm = get_kernel("kernels-community/triton-layer-norm", revision="main")
```
danieldk: We should probably not have have `revision="main"` in the agent skill, since it's a bad pattern we should avoid.

kernel-builder/skills/cuda-kernels/references/huggingface-kernels-integration.md
```diff
@@ -149,7 +156,7 @@ import torch.nn as nn
 from kernels import get_kernel
 
 # Load RMSNorm kernel
-rmsnorm_kernel = get_kernel("kernels-community/triton-layer-norm")
+rmsnorm_kernel = get_kernel("kernels-community/triton-layer-norm", revision="main")
```
danieldk: Same here.

kernel-builder/skills/cuda-kernels/references/huggingface-kernels-integration.md
```diff
@@ -273,21 +301,49 @@ def forward(x: torch.Tensor, out: Optional[torch.Tensor] = None) -> torch.Tensor
     return out
 ```
 
+### Layers (Optional)
+
+A kernel can also export `torch.nn.Module` layers that `kernels.kernelize()` swaps in for matching model layers. Per the [kernel requirements](https://huggingface.co/docs/kernels/kernel-requirements), layers must be **pure**: subclass `torch.nn.Module`, define no constructor, no class variables (except `has_backward` / `can_torch_compile`), and no methods other than `forward`. Put them in `torch-ext/my_kernel/layers.py` and export the module from the main `__init__.py`:
```
danieldk, line 306: I think we should add here, that it is ok to use member variables that are defined by the layers the kernel layer is extending. In that case, the member variable types can be stated for type checking. E.g.:

```python
class RMSNorm(torch.nn.Module):
    has_backward: bool = True
    can_torch_compile: bool = True

    weight: torch.Tensor
    variance_epsilon: float
    
    def forward(self, hidden_states):
        # `rmsnorm` is defined by the kernel.
        return rmsnorm(
            hidden_states,
            self.weight, 
            self.variance_epsilon,
        )
```

kernel-builder/skills/cuda-kernels/scripts/huggingface_kernels_example.py
```diff
@@ -157,12 +158,13 @@ def demo_model_integration():
         # Check for layer norm kernel
         repo_id = "kernels-community/triton-layer-norm"
 
-        if not has_kernel(repo_id):
+        # This repo has no version branches yet, so pin an explicit revision.
+        if not has_kernel(repo_id, revision="main"):
```
danieldk: See above.

kernel-builder/skills/cuda-kernels/scripts/huggingface_kernels_example.py
```diff
@@ -157,12 +158,13 @@ def demo_model_integration():
...
             print(f"   {repo_id} not available, skipping")
             return
 
         print(f"\n1. Loading {repo_id}...")
-        layer_norm = get_kernel(repo_id)
+        layer_norm = get_kernel(repo_id, revision="main")
```
danieldk: See above.

kernel-builder/skills/cuda-kernels/references/diffusers-integration.md
sayakpaul: Would you rather `kernel-builder build-and-copy -L`?

The other stuff, e.g., `cd <>` and `python ...` are okay because the users are supposed to change them. Or did you mean something else?

kernel-builder/skills/cuda-kernels/references/huggingface-kernels-integration.md
sayakpaul: Very good point! Addressed.

kernel-builder/skills/cuda-kernels/references/diffusers-integration.md
danieldk: I meant the `python path/to/skills` etc. I would expect a placeholder in place of `skills/cuda-kernels/scripts/ltx_kernel_injection_example.py`.

sayakpaul: Not sure if I follow. `ltx_kernel_injection_example.py` is the script we want the agent to refer to here.

danieldk (approved):

PR #635, opened 2026-06-11
Add support for verifying kernels

danieldk:
This change adds support for verifying kernels. Verification consists of two parts:

1. Verify that `metadata.json` is correctly signed using the detached signature in `metadata.json.sigstore`; and if so,
2. verify that the kernel file hashes are the same as the digest in `metadata.json`.

The digest in `metadata.json` protects the kernel files, the detached signature in `metadata.json.sigstore` protects the metadata. For signing, we use sigstore, so that kernels can be signed using ephemeral keys. Ephemeral keys reduce the attack surface of leaked keys.

In this change, only the API is added. It is not exposed at the top-level, since it is not fully stable yet. It is also not wired up in `get_kernel`, etc. yet because we still want to test code signing for some time. See `test_verify.py` for actual use of the API. Besides that we'll add a `kernels` subcommand for verifying kernels in the PR after this one.

### DIFF
--- a/kernels-data/bindings/python/kernels_data.pyi
+++ b/kernels-data/bindings/python/kernels_data.pyi
@@ -36,8 +36,8 @@ class Backend(Enum):
         """Parse a backend name.
 
         Args:
-            s: One of ``"cann"``, ``"cpu"``, ``"cuda"``, ``"metal"``,
-               ``"neuron"``, ``"rocm"``, ``"xpu"``.
+            s: One of `"cann"`, `"cpu"`, `"cuda"`, `"metal"`,
+               `"neuron"`, `"rocm"`, `"xpu"`.
 
         Raises:
             ValueError: If the backend name is unknown.
@@ -65,14 +65,14 @@ class BackendInfo:
 
 @final
 class Version:
-    """A dotted numeric version (e.g. ``12.8.0``).
+    """A dotted numeric version (e.g. `12.8.0`).
 
     Trailing zeros are stripped during normalization.
     """
 
     @staticmethod
     def from_str(s: str) -> "Version":
-        """Parse a version string of the form ``X``, ``X.Y``, ``X.Y.Z``, ...
+        """Parse a version string of the form `X`, `X.Y`, `X.Y.Z`, ...
 
         Raises:
             ValueError: If the string is empty or contains non-numeric parts.
@@ -90,10 +90,10 @@ class Version:
 
 @final
 class KernelName:
-    """A validated kernel name matching ``^[a-z][-a-z0-9]*[a-z0-9]$``."""
+    """A validated kernel name matching `^[a-z][-a-z0-9]*[a-z0-9]$`."""
 
     def __new__(cls, name: str) -> "KernelName":
-        """Create a new ``KernelName``.
+        """Create a new `KernelName`.
 
         Raises:
             ValueError: If the name does not match the required pattern.
@@ -128,7 +128,7 @@ class Digest:
     def hash_variant(
         algorithm: DigestAlgorithm, variant_path: os.PathLike[str] | str
     ) -> "Digest":
-        """Hash the files in ``variant_path`` using ``algorithm``.
+        """Hash the files in `variant_path` using `algorithm`.
 
         Args:
             algorithm: Digest algorithm to use.
@@ -151,13 +151,13 @@ class Digest:
         ...
 
     def validate(self, other: "Digest") -> None:
-        """Validate ``other`` (actual) against this digest (expected).
+        """Validate `other` (actual) against this digest (expected).
 
         Returns when the digests match. Otherwise, a `DigestValidationError` is
         raised.
 
         Raises:
-            DigestValidationError: If ``other`` deviates from this digest.
+            DigestValidationError: If `other` deviates from this digest.
         """
         ...
 
@@ -167,7 +167,7 @@ class DigestViolation:
     """A violation of a digest when validated against a reference digest.
 
     This tagged union covers the types of violations. Each violation can be
-    converted to a string using ``str(violation)``.
+    converted to a string using `str(violation)`.
     """
 
     @final
@@ -202,8 +202,8 @@ class DigestViolation:
     class AlgorithmMismatch(DigestViolation):
         """The digest algorithms differ.
 
-        The digest with algorithm ``got`` cannot be validated against the
-        reference digest with algorithm ``expected``.
+        The digest with algorithm `got` cannot be validated against the
+        reference digest with algorithm `expected`.
         """
 
         expected: DigestAlgorithm
@@ -216,7 +216,7 @@ class DigestViolation:
     def __str__(self) -> str: ...
 
 class DigestValidationError(Exception):
-    """Raised by :meth:`Digest.validate` when a digest cannot be validated against the reference."""
+    """Raised by `Digest.validate` when a digest cannot be validated against the reference."""
 
     @property
     def violations(self) -> list[DigestViolation]:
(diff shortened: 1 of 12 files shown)


kernels/src/kernels/verify.py
```diff
@@ -0,0 +1,204 @@
...
+        verifier.verify_artifact(
+            metadata_path.read_bytes(),
+            signature_bundle,
+            verify_policy,
+        )
+    except VerificationError as e:
+        return VerificationResult.SignatureVerificationFailure(reason=str(e))
+
+    try:
+        metadata = Metadata.read_from_file(metadata_path)
```
drbh: maybe we should add a `Metadata.from_bytes` so we can ensure that the verified contents is the same contents that we read as metadata. then we can add something like

```python
metadata_bytes = metadata_path.read_bytes()

...
verifier.verify_artifact(
    metadata_bytes,
    ...
)

...
metadata = Metadata.from_bytes(metadata_bytes)
```

kernels-data/src/digest.rs
```diff
@@ -69,6 +126,43 @@ impl Digest {
...
     }
 
+    /// Validate a digest against the reference.
+    ///
+    /// Returns `Ok(())` if the two digests match. Otherwise returns a
+    /// [`DigestError`] describing every way in which `other` deviates from
+    /// `self`.
+    pub fn validate(&self, other: &Self) -> Result<(), DigestError> {
+        let mut violations = Vec::new();
+
```
drbh, line 146: maybe we should check the algorithm?

played around with this locally and this patch may be useful

```patch
diff --git a/kernels-data/src/digest.rs b/kernels-data/src/digest.rs
index 5b0e229..22e5545 100644
--- a/kernels-data/src/digest.rs
+++ b/kernels-data/src/digest.rs
@@ -32,6 +32,12 @@ pub enum DigestViolation {
         expected: String,
         got: String,
     },
+
+    /// The digest algorithms differ.
+    AlgorithmMismatch {
+        expected: DigestAlgorithm,
+        got: DigestAlgorithm,
+    },
 }
 
 impl fmt::Display for DigestViolation {
@@ -47,6 +53,10 @@ impl fmt::Display for DigestViolation {
                 f,
                 "File hash mismatch for {path}: expected: {expected}, got: {got}"
             ),
+            DigestViolation::AlgorithmMismatch { expected, got } => write!(
+                f,
+                "Algorithm mismatch: expected: {expected:?}, got: {got:?}"
+            ),
         }
     }
 }
@@ -134,6 +144,13 @@ impl Digest {
     pub fn validate(&self, other: &Self) -> Result<(), DigestError> {
         let mut violations = Vec::new();
 
+        if self.algorithm != other.algorithm {
+            violations.push(DigestViolation::AlgorithmMismatch {
+                expected: self.algorithm,
+                got: other.algorithm,
+            });
+        }
+
         for (path, expected) in &self.files {
             match other.files.get(path) {
                 None => {
@@ -169,7 +186,7 @@ impl Digest {
     }
 }
 
-#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
 pub enum DigestAlgorithm {
     #[serde(rename = "sha256")]
     SHA256,
@@ -380,6 +397,29 @@ mod tests {
              Unknown file: extra.bin"
         );
     }
+
+    #[test]
+    fn validate_detects_algorithm_mismatch_for_empty_digest() {
+        let expected = Digest {
+            algorithm: DigestAlgorithm::SHA256,
+            files: BTreeMap::new(),
+        };
+        let actual = Digest {
+            algorithm: DigestAlgorithm::SHA512,
+            files: BTreeMap::new(),
+        };
+
+        assert_ne!(expected.algorithm(), actual.algorithm());
+
+        let err = expected.validate(&actual).unwrap_err();
+
+        assert_eq!(err.violations().len(), 1);
+        assert_eq!(
+            err.to_string(),
+            "kernel does not match the expected digest:\n\
+             Algorithm mismatch: expected: SHA256, got: SHA512"
+        );
+    }
 }
 
 impl From<DigestAlgorithm> for Box<dyn DynDigest> {
```

drbh (commented):
looks really good! excited for this feature, just left a couple comments with small suggestions to harden the impl

kernels-data/bindings/python/tests/test_digest.py
```diff
@@ -0,0 +1,154 @@
...
+def test_hash_variant_hashes_file_with_sha512(tmp_path):
+    (tmp_path / "_extension.so").write_bytes(b"hello world")
+
+    digest = Digest.hash_variant("sha512", tmp_path)
+
+    assert len(digest.files) == 1
+    assert (
+        digest.files["_extension.so"]
+        == "MJ7MSJwS1utMxA9QyQLytNDtd+5RGnx6m808qG1M2G+YndNbxf9JlnDaNCVbRbDP2DDoH2Bdz33FVC6TrpzXbw=="
+    )
```
sayakpaul, line 26: (nit): The two tests could be clubbed into a single ` test_hash_variant_hashes_file_with_algo()` test with `parametrize`. But I will understand if we want to keep them separate. No strong opinions.

kernels-data/bindings/python/tests/test_digest.py
```diff
@@ -0,0 +1,154 @@
...
+
+    message = str(exc_info.value)
+    assert "Missing file: gone.py" in message
+    assert "Unknown file: extra.bin" in message
+    assert "File hash mismatch for shared.so" in message
+
+    # `str(violation)` reuses the same human-readable rendering.
+    rendered = {str(v) for v in exc_info.value.violations}
+    assert "Missing file: gone.py" in rendered
+    assert "Unknown file: extra.bin" in rendered
```
sayakpaul, line 154: Love the tests!

kernels-data/bindings/python/kernels_data.pyi
```diff
@@ -99,6 +109,93 @@ class KernelName:
...
+        got: str
+        __match_args__ = ("path", "expected", "got")
+        def __new__(
+            cls, path: str, expected: str, got: str
+        ) -> "DigestViolation.HashMismatch": ...
+
+    def __str__(self) -> str: ...
+
+class DigestValidationError(Exception):
+    """Raised by :meth:`Digest.validate` when a digest cannot be validated against the reference."""
```
sayakpaul: Do we want to keep `:meth:` denotations? I don't think they exist in the rest of the library codebase.

kernels/tests/test_verify.py
```diff
@@ -0,0 +1,112 @@
+from kernels_data import DigestViolation
+from sigstore.verify import policy
+
+from kernels import install_kernel
+from kernels.utils import select_revision_or_version
+from kernels.verify import VerificationResult, verify_variant
+
+TEST_POLICIES: list[policy.VerificationPolicy] = [
```
sayakpaul, line 8: For users / orgs outside the predefined set of policies, should users bypass this verification? What is the approach we're thinking of for trusted publishers outside HF (SGLang, NVIDIA, etc.) -- what would be the recommended way for them to sign?

sayakpaul (commented):
Thanks, just left some questions.

kernels/src/kernels/verify.py
danieldk: Fixed.

kernels-data/src/digest.rs
danieldk: Yeah, good idea. Fixed this. Also added `DigestAlgorithm` to the Python side to make it more strongly typed.

kernels/tests/test_verify.py
danieldk: > For users / orgs outside the predefined set of policies, should users bypass this verification?

Yep, similar to how they have to use `trust_remote_code` now.

> What is the approach we're thinking of for trusted publishers outside HF (SGLang, NVIDIA, etc.) -- what would be the recommended way for them to sign?

Hooking signing up in CI is very simple:

https://github.com/huggingface/kernels-community/blob/398180d39558d614a0461e7eee79e78890ef44ac/.github/workflows/build.yaml#L220

Then we'd add these organizations' identities/GH repos to the policies to accept them. Though ideally, we'd be able to get the list from the Hub, as we do now with the trusted publishers.

sayakpaul (dismissed):
Excited to see this feature.

I don't think we need any documentation yet.

kernels-data/bindings/python/kernels_data.pyi
danieldk: Ah yes. The stubs were generated automatically. Should be fixed now 👍 .

drbh (approved):
looks great to me!

PR #637, opened 2026-06-12
feat: implement load telemetry.

### LINKED ISSUE #634: Add telemetry ping to track exact kernel load count
## Summary

When a kernel is loaded (whether from Hub or local cache), we should send a single telemetry ping to `/api/telemetry/kernels/loaded` so we can track **exact kernel load counts** over time.

## Motivation

Currently we have two metrics that don't quite capture what we need:

- **`user_agent` via `huggingface_hub` requests** — over-reports, because a single `get_kernel()` call may trigger multiple HTTP requests (resolve + LFS chunks), each carrying the same UA.
- **Hub download statistics** — under-reports, because cached kernels aren't re-downloaded (especially with stable ABI reducing download frequency).

Neither gives us the true number of times a kernel is loaded and used.

## Proposed solution

Add a call to `huggingface_hub`'s `send_telemetry()` utility in the **kernel import/load code path** (not just `get_kernel()`), so it fires on every load including local/cached ones:

```python
from huggingface_hub.utils import send_telemetry

send_telemetry(
    "kernels/loaded",
    library_name="kernels",
    library_version=__version__,
    user_agent={"kernel": repo_id},
)
```

This is a lightweight HEAD request on a background thread (non-blocking), respects `HF_HUB_DISABLE_TELEMETRY`, and provides a clean 1:1 signal per kernel load.

## Reading the data back

Once instrumented, the telemetry events land in Elasticsearch (Hub structured logs). A cron job can query ES for `/api/telemetry/kernels/loaded` events to produce periodic usage reports. ES access setup is tracked separately with infra.

## Context

Discussed in: https://huggingface.slack.com/archives/C090JN2P8NB/p1749549498161619

cc @sayakpaul @danieldk

sayakpaul:
Fixes #634

### DIFF
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -108,10 +108,13 @@ class RepoInfo:
 
     - `repo_id` (`str`): the Hub repository containing the kernel.
     - `revision` (`str`): the specific revision of the kernel.
+    - `local` (`bool`): whether the kernel was loaded from pre-downloaded local
+      files (`True`) rather than downloaded from the Hub (`False`).
     """
 
     repo_id: str
     revision: str
+    local: bool = False
 
 
 @dataclass(frozen=True)
@@ -121,10 +124,10 @@ class LoadedKernel:
 
     - `metadata` (`Metadata`): kernel metadata.
     - `module` (`ModuleType`): the imported kernel module.
-    - `repo_info` (`kernels.utils.RepoInfo | None`): populated only for
-      kernels loaded via `get_kernel`. Loaders that work from a local path
-      (`get_local_kernel`) or a lockfile (`get_locked_kernel`, `load_kernel`)
-      leave this as `None`.
+    - `repo_info` (`kernels.utils.RepoInfo | None`): populated for kernels loaded
+      from a known Hub repository, i.e. via `get_kernel`, `get_locked_kernel`, or
+      `load_kernel`. Only `get_local_kernel`, which loads from an arbitrary local
+      path, leaves this as `None`.
 
     The metadata includes the following properties that describe a kernel:
 
@@ -207,8 +210,6 @@ def _validate_variant_dependencies(variant_path: Path) -> None:
 def _import_from_path(
     variant_path: Path,
     repo_info: RepoInfo | None = None,
-    *,
-    telemetry_repo_info: RepoInfo | None = None,
 ) -> ModuleType:
     if (loaded_kernel := _loaded_kernels.get(variant_path)) is not None:
         return loaded_kernel.module
@@ -238,7 +239,7 @@ def _import_from_path(
     )
 
     # Telemetry request is non-blocking.
-    _send_load_telemetry(metadata, repo_info or telemetry_repo_info)
+    _send_load_telemetry(metadata, repo_info)
 
     return module
 
@@ -698,7 +699,7 @@ def load_kernel(
         ) from e
     return _import_from_path(
         variant_path,
-        telemetry_repo_info=RepoInfo(repo_id=repo_id, revision=locked_sha),
+        repo_info=RepoInfo(repo_id=repo_id, revision=locked_sha, local=True),
     )
 
 
@@ -729,7 +730,7 @@ def get_locked_kernel(repo_id: str, local_files_only: bool = False) -> ModuleTyp
 
     return _import_from_path(
         variant_path,
-        telemetry_repo_info=RepoInfo(repo_id=repo_id, revision=locked_sha),
+        repo_info=RepoInfo(repo_id=repo_id, revision=locked_sha),
     )
 
 
@@ -886,6 +887,7 @@ def _send_load_telemetry(metadata: Metadata, repo_info: RepoInfo | None) -> None
         if repo_info is not None:
             user_agent["repo_id"] = repo_info.repo_id
             user_agent["revision"] = repo_info.revision
+            user_agent["local"] = str(repo_info.local).lower()
         user_agent["kernel_name"] = str(metadata.name)
         user_agent["kernel_version"] = str(metadata.version)
         user_agent.update(_system_user_agent())
--- a/kernels/src/kernels/utils.py
+++ b/kernels/src/kernels/utils.py
@@ -204,7 +204,12 @@ def _validate_variant_dependencies(variant_path: Path) -> None:
     validate_dependencies(metadata.name.python_name, metadata.python_depends, _backend())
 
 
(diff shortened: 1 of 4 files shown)


kernels/src/kernels/utils.py
```diff
@@ -204,7 +204,12 @@ def _validate_variant_dependencies(variant_path: Path) -> None:
     validate_dependencies(metadata.name.python_name, metadata.python_depends, _backend())
 
 
-def _import_from_path(variant_path: Path, repo_info: RepoInfo | None = None) -> ModuleType:
+def _import_from_path(
+    variant_path: Path,
+    repo_info: RepoInfo | None = None,
+    *,
+    telemetry_repo_info: RepoInfo | None = None,
```
danieldk: Why? We should just send `repo_info`.

kernels/src/kernels/utils.py
```diff
@@ -231,6 +236,12 @@ def _import_from_path(variant_path: Path, repo_info: RepoInfo | None = None) ->
         module=module,
         repo_info=repo_info,
     )
+
+    # Under the hood, we use `send_telemetry()` from `hugginggface_hub`.
+    # That already respect the env var concerning disabling telemetry
+    # and is also thread non-blocking.
```
danieldk: I would move the part about the disabling telemetry to the function itself (not relevant here) and just say:


```suggestion
    # Telemetry request is non-blocking.
```

kernels/src/kernels/utils.py
```diff
@@ -715,7 +729,10 @@ def get_locked_kernel(repo_id: str, local_files_only: bool = False) -> ModuleTyp
         validate_dependencies=True,
     )
 
-    return _import_from_path(variant_path)
+    return _import_from_path(
+        variant_path,
+        telemetry_repo_info=RepoInfo(repo_id=repo_id, revision=locked_sha),
```
danieldk: Ok, I see why this is passed, but `repo_info` is only used to store the loaded kernel metadata and this may actually be useful there as well?

kernels/src/kernels/utils.py
```diff
@@ -827,6 +844,65 @@ def _platform() -> str:
...
+
+        from . import __version__
+
+        # `kernels_event/load` is set only on load pings and on no other Hub
+        # request, so loads can be separated from other requests by the UA
+        # alone. The kernel-specific fields that follow identify which kernel
+        # was loaded.
+        user_agent: dict[str, str] = {"kernels_event": "load"}
+        if repo_info is not None:
+            user_agent["kernel"] = repo_info.repo_id
```
danieldk: any reason not to call this `repo_id`?

kernels/src/kernels/utils.py
```diff
@@ -827,6 +844,65 @@ def _platform() -> str:
...
+            user_agent["revision"] = repo_info.revision
+        user_agent["kernel_name"] = metadata.name.python_name
```
danieldk: I'd prefer the normal name (as a str) over the Python name.

danieldk (commented):
Some small nits.

kernels/src/kernels/utils.py
sayakpaul: Are you suggesting passing `repo_info` as `RepoInfo(repo_id=repo_id, revision=locked_sha)`? Wouldn't it change the documented behavior where lockfile loaders leave `repo_info` as None?

kernels/src/kernels/utils.py
sayakpaul: Nothing specific. Just felt a bit idiomatic to refer a `repo_id` (which is of type `kernel`) a kernel in the logs. But changed to `repo_id` for consistency.

kernels/src/kernels/utils.py
sayakpaul: Done (and updated the tests too).

kernels/src/kernels/utils.py
danieldk: Is that documented behavior? I think it's still nice to be able to query the loaded kernels and see their repo_info. We could add a bool or some other field that marks whether it was loaded 'remotely' or locally.

sayakpaul: Yeah I think so:

> - `repo_info` (`kernels.utils.RepoInfo | None`): populated only for
      kernels loaded via `get_kernel`. Loaders that work from a local path
      (`get_local_kernel`) or a lockfile (`get_locked_kernel`, `load_kernel`)
      leave this as `None`.

> I think it's still nice to be able to query the loaded kernels and see their repo_info. We could add a bool or some other field that marks whether it was loaded 'remotely' or locally.

Good point but I was thinking if `repo_info` it would mean that it's supposed to be used locally. But there's no way to track its lineage to see if it was loaded remotely. Adding the info on `repo_id` and `locked_sha` seems like a better solution to me than adding a `bool`. 

WDYT?

danieldk: I am not sure if I understand. `RepoInfo` already contains the `repo_id` and the locked SHA (as `revision`), so we only need something to flag whether it was downloaded or locally loaded? Suppose we have

```python
@dataclass
class RepoInfo:
    repo_id: str
    revision: str
    local: bool = False
```

Then we can use `RepoInfo(repo_id=repo_id, revision=locked_sha, local=True)` for `load_kernel` (but not `get_locked_kernel`, since it downloads).

We cannot do anything for `get_local_kernel`, because it could be anything. Probably not used that frequently anyway.

danieldk: Sorry, extended the comment a bit post-hoc.

sayakpaul: Thanks. To me, the notion of `local` is a bit ambiguous here. Are we referring to it as `local` because we don't make a call to the Hub in this case? If so, yeah, I like the idea.

sayakpaul: Fixed.

danieldk (approved):
Looks great.

If it's a clean cherry-pick, I'll also backport this to 0.15.x, so that we can roll this out immediately.

kernels/src/kernels/utils.py
```diff
@@ -827,6 +843,66 @@ def _platform() -> str:
...
+        user_agent["kernel_name"] = str(metadata.name)
+        user_agent["kernel_version"] = str(metadata.version)
+        user_agent.update(_system_user_agent())
+
+        send_telemetry(
+            "kernels/loaded",
+            library_name="kernels",
+            library_version=__version__,
+            user_agent=user_agent,
+        )
```
julien-c, line 900: it was not possible to piggy-back on the download (HEAD and GET) we already make to the Hub, like we do for most of our open source libraries? cc @LysandreJik @Pierrci

PR #638, opened 2026-06-12
[ci] remove unsupported 3.13t for maturin

sayakpaul:
Fixes https://github.com/huggingface/kernels/actions/runs/27402037890/job/80982190809?pr=627 (removes 3.13t build in CI as maturin refuses to produce an abi3 wheel against free-threaded 3.13 (3.13t))

### DIFF
--- a/.github/workflows/kernels_data_python_release.yaml
+++ b/.github/workflows/kernels_data_python_release.yaml
@@ -9,8 +9,8 @@
 #    `PyO3/maturin-action` + PyPI trusted publishing (see plan.md §3a, §3b)
 #  - pull_request trigger gained `paths-ignore` for docs-only PRs
 #  - concurrency block added (cancel-in-progress only for PRs, never on tag/main)
-#  - free-threaded 3.13t build steps removed: abi3 wheels for free-threaded
-#    CPython are only supported on 3.15+ (maturin rejects 3.13t with abi3)
+#  - free-threaded 3.13t build steps removed: not supported by maturin
+#    anymore.
 name: CI
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/kernels_data_python_release.yaml
```diff
@@ -9,6 +9,8 @@
 #    `PyO3/maturin-action` + PyPI trusted publishing (see plan.md §3a, §3b)
 #  - pull_request trigger gained `paths-ignore` for docs-only PRs
 #  - concurrency block added (cancel-in-progress only for PRs, never on tag/main)
+#  - free-threaded 3.13t build steps removed: abi3 wheels for free-threaded
+#    CPython are only supported on 3.14+ (maturin rejects 3.13t with abi3)
```
danieldk:
```suggestion
#    CPython are only supported on 3.15+ (maturin rejects 3.13t with abi3)
```

also not abi3, but abi3t.

.github/workflows/kernels_data_python_release.yaml
```diff
@@ -61,13 +63,6 @@ jobs:
           args: --release --out dist --manifest-path kernels-data/bindings/python/Cargo.toml
           sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}
           manylinux: auto
-      - name: Build free-threaded wheels (3.13t)
```
danieldk, line 64: This was supported before. Seems that, from the error, maturin removed support?

.github/workflows/kernels_data_python_release.yaml
danieldk: I would change this to something like:

```suggestion
#  - free-threaded 3.13t build steps removed: not supported by maturin
#    anymore.
```

danieldk (approved):

PR #644, opened 2026-06-15
kernels: tighten up code signing security

danieldk:
- Allow specifying the workflow name and branch to the GitHub workflow policy. Doing so reduces the attack surface (e.g. signing from branches).
- Only exclude top-level `metadata.json`/`metadata.json.sigstore` from the digest.


kernels/src/kernels/verify.py
```diff
@@ -19,20 +19,44 @@ class GitHubWorkflowPolicy(VerificationPolicy):
...
             The GitHub repository in the format `owner/repo` that matches
             the workflow that signed the kernel.
+        workflowNames (`list[str]`, *optional*):
+            The workflows within the repository from which signed artifacts
+            are accepted. It is strongly recommended to set this to block
+            signatures from arbitrary workflows.
+        workflowRef (`str`, *optional*):
+            The ref of the workflow that signed the artifact, such as
+            `refs/heads/main`. It is strongly recommended to set the ref to
+            block signatures from arbitrary branches or tags.
```
sayakpaul, line 29: Any reason why this is camel case?

sayakpaul (dismissed):

kernels/src/kernels/verify.py
danieldk: Maybe looked at a bit too much Kotlin the last few days 😓 . Fixed!

sayakpaul (approved):

PR #649, opened 2026-06-16
kernels: verify using the signer URIs

danieldk:
This is more exact and also makes it easier to maintain Hub-side(?)
whitelists in the future.

Also remove kernels-test/kernels-staging for now. Rationale:

- Staging kernels are signed by the kernels-community CI.
- We still need to set up CI for kernels-test.

### DIFF
--- a/kernels/src/kernels/verify.py
+++ b/kernels/src/kernels/verify.py
@@ -16,45 +16,29 @@ class GitHubWorkflowPolicy(VerificationPolicy):
     given GitHub repository.
 
     Args:
-        repo_id (`str`):
-            The GitHub repository in the format `owner/repo` that matches
-            the workflow that signed the kernel.
-        workflowNames (`list[str]`, *optional*):
-            The workflows within the repository from which signed artifacts
-            are accepted. It is strongly recommended to set this to block
-            signatures from arbitrary workflows.
-        workflowRef (`str`, *optional*):
-            The ref of the workflow that signed the artifact, such as
-            `refs/heads/main`. It is strongly recommended to set the ref to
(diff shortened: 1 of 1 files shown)


kernels/src/kernels/verify.py
```diff
@@ -16,45 +16,29 @@ class GitHubWorkflowPolicy(VerificationPolicy):
...
-            are accepted. It is strongly recommended to set this to block
-            signatures from arbitrary workflows.
-        workflowRef (`str`, *optional*):
-            The ref of the workflow that signed the artifact, such as
-            `refs/heads/main`. It is strongly recommended to set the ref to
-            block signatures from arbitrary branches or tags.
+        signer_uris (`list[str]`):
+            List of workflows URIs that are allowed as kernel signers. Each
+            URI has the shape:
+            `https://github.com/<org>/<repo>/.github/workflows/<workflow>.yaml@<ref>`
```
sayakpaul, line 22: So, this means the valid signing entity can only come from a GitHub CI, right? Fine by me, just trying to understand.

sayakpaul (approved):
Should the tests also take effect from this?

kernels/src/kernels/verify.py
danieldk: Yep and we lock it down using `OIDCIssuer`.

danieldk:
> Should the tests also take effect from this?

The current tests use a user-bound OIDC signatures because I had to make all the broken cases as well, that we cannot generate from CI. But as soon as we have `kernels-test` CI running, we could also add workflow-signed kernels as test cases.

PR #650, opened 2026-06-16
kernels: add verify-signature subcommand

danieldk:
`kernels verify-signature <repo_id> <version>` verifies the signature of the given kernel and kernel version.

### DIFF
--- a/.github/workflows/test_extra_commands.yaml
+++ b/.github/workflows/test_extra_commands.yaml
@@ -33,3 +33,5 @@ jobs:
           USER: runner
       - name: Test nix run .#kernels
         run: ( cd examples/kernels/relu ; nix run .#kernels -- lock ../../../nix-builder/tests/run-kernels )
+      - name: Test verify-signature
+        run: nix run .#kernels -- verify-signature kernels-community/relu 1
--- a/docs/source/cli-verify-signature.md
+++ b/docs/source/cli-verify-signature.md
@@ -36,36 +36,10 @@ kernels verify-signature kernels-community/relu 1 --all-variants
 ## Example Output
 
 ```bash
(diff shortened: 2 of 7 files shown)


docs/source/cli-verify-signature.md
```diff
@@ -0,0 +1,42 @@
...
+
+```bash
+kernels verify-signature kernels-community/relu 1
+```
+
+Verify all build variants of the same kernel:
+
+```bash
+kernels verify-signature kernels-community/relu 1 --all-variants
+```
```
sayakpaul, line 34: Embed expected outputs too? And provide an example in case the signature verification fails?

sayakpaul (commented):
Tests missing?

danieldk:
> Tests missing?

See the CI change. We typically don't have tests outside that (here it'd also be pretty variable, since we are still adding this to kernels).

docs/source/cli-verify-signature.md
danieldk: Added, created an extra section for that, like in some of the existing CLI docs.

.github/workflows/test_extra_commands.yaml
```diff
@@ -33,3 +33,5 @@ jobs:
           USER: runner
       - name: Test nix run .#kernels
         run: ( cd examples/kernels/relu ; nix run .#kernels -- lock ../../../nix-builder/tests/run-kernels )
+      - name: Test verify-signature
+        run: nix run .#kernels -- verify-signature kernels-community/relu 1
```
sayakpaul, line 37: Ah I missed it completely. I was thinking about a dedicated tes suite for `verify-signature`.

sayakpaul (approved):

PR #652, opened 2026-06-16
kernels: also require `OIDCSourceRepositoryURI`

danieldk:
`OIDCBuildSignerURI` checks the callee, so an attacker could try to reuse our actions to get their code signed. So also require `OIDCSourceRepositoryURI` to pin down the caller (the source repo).

### DIFF
--- a/kernels/src/kernels/verify.py
+++ b/kernels/src/kernels/verify.py
@@ -25,17 +25,20 @@ class GitHubWorkflowPolicy(VerificationPolicy):
     def __init__(
         self,
         *,
+        repo_id: str,
         signer_uris: list[str],
     ):
         if not signer_uris:
             raise ValueError("At least one signer URI must be provided")
 
+        self.repo_id = repo_id
         self.signer_uris = signer_uris
 
     def verify(self, cert: Certificate) -> None:
         policies: list[VerificationPolicy] = [
             policy.OIDCIssuer("https://token.actions.githubusercontent.com"),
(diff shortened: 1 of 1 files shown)


kernels/src/kernels/verify.py
```diff
@@ -50,6 +53,7 @@ def verify(self, cert: Certificate) -> None:
 """
 DEFAULT_POLICIES: list[VerificationPolicy] = [
     GitHubWorkflowPolicy(
+        repo_id="huggingface/kernels-community",
         signer_uris=[
             "https://github.com/huggingface/kernels-community/.github/workflows/build.yaml@refs/heads/main",
```
drbh, line 58: is there a case where the repo id and the path in the signer uri will differ? can the repo id be derived? 

just wondering because it looks a bit redundant, however I'm probably not thinking of the scenario where this happens

drbh (approved):
lgtm just left a question

kernels/src/kernels/verify.py
danieldk: Yep. If the action in a repo is used by a different repo. E.g. if `evilhax0r/evilfork` uses the workflow from`https://github.com/goodorg/goodrepo/.github/workflows/build.yaml@refs/heads/main`, then the signer URI would be `https://github.com/goodorg/goodrepo/.github/workflows/build.yaml@refs/heads/main`.

Adding `OIDCSourceRepositoryURI` fixes this case, because it would be `evilhax0r/evilfork`.

PR #653, opened 2026-06-16
kernels: do not skip malformed bundles when using `--filter-unsigned`

danieldk:


sayakpaul (approved):
Now this is a hard failure then.

PR #658, opened 2026-06-17
Support `pyext` in `torch-noarch`

danieldk:


kernel-builder/src/pyproject/torch/noarch.rs
```diff
@@ -101,6 +107,7 @@ fn write_pyproject_toml(
                 python_dependencies => python_dependencies,
                 backend_dependencies => backend_dependencies,
                 name => name,
+                python_name => general.name.python_name(),
```
sayakpaul, line 110: Makes sense!

sayakpaul (approved):

PR #663, opened 2026-06-18
kernel-builder hash: do not try to join two full paths

danieldk:
`discover_variants` already returns absolute paths, so joining
with the build dir was redundant and incorrect for relative paths.


sayakpaul (approved):
Makes sense! Any test that might have helped us catch it?

danieldk:
I think we need a more generic framework for command-line tests maybe?

PR #665, opened 2026-06-18
feat: enable cutlass 4.5 on nix builder

mfuntowicz:
This PR attempts to enable more recent version of cutlass, v4.5.2 in this case, to support more recent architecture and dispatching like SM120

### DIFF
--- a/nix-builder/pkgs/cutlass/default.nix
+++ b/nix-builder/pkgs/cutlass/default.nix
@@ -28,8 +28,14 @@ in
     version = "3.9.2";
     hash = "sha256-teziPNA9csYvhkG5t2ht8W8x5+1YGGbHm8VKx4JoxgI=";
   };
+
   cutlass_4_0 = builder {
     version = "4.0.0";
     hash = "sha256-HJY+Go1viPkSVZPEs/NyMtYJzas4mMLiIZF3kNX+WgA=";
+  };
+
+  cutlass_4_5 = builder {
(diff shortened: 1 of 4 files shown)


nix-builder/pkgs/cutlass/default.nix
```diff
@@ -28,8 +28,14 @@ in
     version = "3.9.2";
     hash = "sha256-teziPNA9csYvhkG5t2ht8W8x5+1YGGbHm8VKx4JoxgI=";
   };
+
   cutlass_4_0 = builder {
     version = "4.0.0";
     hash = "sha256-HJY+Go1viPkSVZPEs/NyMtYJzas4mMLiIZF3kNX+WgA=";
   };
+
+  cutlass_4_5 = builder {
```
danieldk, line 37: Thanks! `cutlass_4_5` needs to be added to some other places as well,

```
❯ git grep cutlass_4_0                                                                                                                                                                                                               
kernels-data/src/config/deps.rs:    #[serde(rename = "cutlass_4_0")]                                                                                                                                                                 
nix-builder/lib/deps.nix:    "cutlass_4_0" = [                                                                                                                                                                                       
nix-builder/lib/deps.nix:      pkgs.cutlass_4_0                                                                                                                                                                                      
nix-builder/pkgs/cutlass/default.nix:  cutlass_4_0 = builder {
```

danieldk (approved):
Thanks a lot!

PR #666, opened 2026-06-19
feat: append builder info to detect builds made with uncommitted changes in the builder

### LINKED ISSUE #648: Signal when a kernel build comes from an uncommitted (dirty) kernel-builder version
## Context

We want to make it clear to users when a build comes from a `kernel-builder` version that contains uncommitted changes, to improve reproducibility guarantees.

## Proposal

Add the git commit SHA to the build metadata, along with a `dirty` bool. If `dirty` is set, we know that the kernel was built with uncommitted sources.

### Suggested behavior when `dirty` is true:

1. **At load time** — Show warnings to the user when they're trying to load the kernel.
2. **At upload time** — Throw warnings when the builds are going to get uploaded to the Hub.
3. **On the Hub repo** — Show warnings on the corresponding Hub repo page.

## References

From discussion by @SayakPaul and @danieldk.

---
_Created via Moon Bot for Sayak Paul (Slack U03AU4E7DJB) — [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1781581193698689?thread_ts=1781581193.698689&cid=C090JN2P8NB)_

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/648

<details>
<summary>Obsolete for now</summary>

**Considerations**:

* The Nix sandbox has no `.git`, so this information has to be passed in explicitly. So, builder-info related stuff become command-line arguments embedded in the derivation's build script at eval time. Is there a better way to handle it?
* `nix build` / `nix flake show` writes an untracked `flake.lock`, which makes the kernel tree dirty aka `kernel.dirty=true` unless the lock is committed. I have made a note about it in the docs. Is there a better way to handle it?

</details>

**Example `metadata.json` with builder-info**:

```json

"provenance": {
  "kernel-builder": {
    "version": "0.17.0-dev0",
    "sha": "ebfedf0ffbf7b12da51cc336f90a26650090fdca",
    "dirty": false
  },
  "kernel": {
    "sha": "ebfedf0ffbf7b12da51cc336f90a26650090fdca",
    "dirty": false
  }
}
```

---

We will throw a warning during upload to the Hub and also show a warning on the Hub card that the corresponding builds were made with a "dirty" version. Kernel loading also logs similar warnings. 

As discussed over Slack, this PR will only tackle the metadata-related stuff. Warnings and related things will be added in subsequent PRs.


kernel-builder/src/pyproject/ops_identifier.rs
```diff
@@ -28,9 +29,25 @@ pub fn git_identifier(target_dir: impl AsRef<Path>) -> Result<String> {
...
+    let dirty = !repo.statuses(Some(&mut status_options)).ok()?.is_empty();
+
+    Some(GitInfo { sha, dirty })
+}
+
 pub struct KernelIdentifier {
     name: String,
     unique_id: String,
+    git_info: Option<GitInfo>,
+    kernel_builder: Option<KernelBuilderInfo>,
```
danieldk: We shouldn't overload this struct, it's purely for uniquely identifying a kernel (e.g. for ops name generation). Let's keep it small and atomic.

kernel-builder/src/pyproject/common.rs
```diff
@@ -20,11 +20,30 @@ pub fn write_compat_py(file_set: &mut FileSet) -> Result<()> {
...
     general: &General,
     kernel_id: &KernelIdentifier,
     file_set: &mut FileSet,
 ) -> Result<()> {
+    // Prefer externally-provided `kernel-builder` provenance (e.g. from Nix),
+    // falling back to the provenance baked in at compile time.
+    let kernel_builder = kernel_id
+        .kernel_builder()
+        .cloned()
+        .unwrap_or_else(kernel_builder_info);
```
danieldk: This seems the wrong place to do this. The builder's hash/version should be burned into the builder, so we should adjust the derivation to give the build access to it.

kernel-builder/src/pyproject/mod.rs
```diff
@@ -39,16 +40,38 @@ pub fn create_pyproject_file_set(build: Build, kernel_id: &KernelIdentifier) ->
...
 }
 
+#[allow(clippy::too_many_arguments)]
 pub fn create_pyproject(
     kernel_dir: Option<PathBuf>,
     target_dir: Option<PathBuf>,
     force: bool,
     unique_id: Option<String>,
+    kernel_sha: Option<String>,
+    kernel_dirty: bool,
```
danieldk, line 53: This can lead to incorrect metadata in case someone generates `pyproject.toml` and then adds more commits. So doing this from CMake (and then giving the kernel build sandbox access to git files) seems like the way to do it. For the Nix build (since we probably do not want to copy all the git objects to the sandbox), we could pass the SHA to CMake. Not sure how important this is though.

This is currently also an issue for the sha hash we use in the ops name. We are generating it when making the project files, but it should probably be done inside CMake (but would better be a separate PR to avoid PRs becoming too big).

kernel-builder/src/upload.rs
```diff
@@ -150,6 +150,11 @@ fn run_upload_typed<T: RepoType>(args: UploadArgs) -> Result<()> {
         build_dir.display()
     );
 
+    let dirty_build = detect_dirty_build(&variants);
```
danieldk: Can we split the dirty warnings into a separate PR? I think it's best to iterate on the metadata first as a smaller unit.

kernel-builder/build.rs
```diff
@@ -1,3 +1,52 @@
+use std::process::Command;
+
 fn main() {
     minijinja_embed::embed_templates!("src/pyproject/templates");
+    emit_git_info();
+}
+
+fn emit_git_info() {
```
danieldk: Might be worth using `built` over our own custom code:

https://crates.io/crates/built

Also uses `git2` rather than shelling out, which is generally more reliable than shelling out (e.g. `git` may not be available in a build sandbox).

Also seems to be used by at least one very high-profile crate (rav1e).

kernels-data/src/metadata.rs
```diff
@@ -15,6 +15,40 @@ pub struct BackendInfo {
     pub archs: Option<Vec<String>>,
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct GitInfo {
```
danieldk: I think something like `GitHash` would be nicer.

kernels-data/src/metadata.rs
```diff
@@ -15,6 +15,40 @@ pub struct BackendInfo {
...
+}
+
+/// Provenance of the `kernel-builder` that produced a build.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct KernelBuilderInfo {
+    pub version: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sha: Option<String>,
+    pub dirty: bool,
```
danieldk: Embed `GitInfo`/`GitHash` and use the serde option that flattens it.

kernels-data/src/metadata.rs
```diff
@@ -15,6 +15,40 @@ pub struct BackendInfo {
...
+#[serde(rename_all = "kebab-case")]
+pub struct GitInfo {
+    pub sha: String,
+    pub dirty: bool,
+}
+
+/// Provenance of the `kernel-builder` that produced a build.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct KernelBuilderInfo {
```
danieldk: I think `KernelBuilderVersion` is nicer.

kernels-data/src/metadata.rs
```diff
@@ -15,6 +15,40 @@ pub struct BackendInfo {
...
+    pub version: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sha: Option<String>,
+    pub dirty: bool,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct BuildInfo {
```
danieldk:
```suggestion
pub struct Provenance {
```

Maybe?

kernels-data/src/metadata.rs
```diff
@@ -15,6 +15,40 @@ pub struct BackendInfo {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub kernel_builder: Option<KernelBuilderInfo>,
```
danieldk: Do we want this to be `None` ever?

kernels-data/src/metadata.rs
```diff
@@ -15,6 +15,40 @@ pub struct BackendInfo {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub kernel: Option<GitInfo>,
```
danieldk: Same here, do we want this to be `None` ever?

kernels-data/src/metadata.rs
```diff
@@ -31,6 +65,8 @@ pub struct Metadata {
     pub backend: BackendInfo,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub digest: Option<Digest>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub build_info: Option<BuildInfo>,
```
danieldk:
```suggestion
    pub provenance: Option<Provenance>,
```

Seems nicer.

danieldk: Also, this may be a good moment to start introducing versions and bump up the version for this change, so that we can make this mandatory.

To keep things reviewable, we should probably do the introduction of the version as a separate PR.

nix-builder/lib/extension/torch/arch.nix
```diff
@@ -71,6 +71,10 @@
   # Revision to bake into the ops name.
   rev,
 
+  # Extra `create-pyproject` flags recording git provenance (commit SHA and
+  # dirty state) of the kernel source and `kernel-builder`.
+  provenanceArgs ? "",
```
danieldk: I think it's nicer if the args are just the atomic data, not flags.

nix-builder/lib/build.nix
```diff
@@ -113,6 +113,7 @@ rec {
       rev,
       doGetKernelCheck,
       stripRPath ? false,
+      provenanceArgs ? "",
```
danieldk: Why set defaults? Makes it easy to miss.

nix-builder/lib/gen-flake-outputs.nix
```diff
@@ -11,6 +11,12 @@
   rev ? null,
   self ? null,
 
+  # Git provenance of the `kernel-builder` itself (the flake revision it was
+  # evaluated from), recorded in the build metadata. `null`/`false` when
+  # `kernel-builder` is used from a non-git source (e.g. a local `path:`).
```
danieldk: A local path can also be a git source. I think in flakes it must even be.

nix-builder/lib/gen-flake-outputs.nix
```diff
@@ -11,6 +11,12 @@
...
+  builderRev ? null,
+  builderDirty ? false,
```
danieldk: This makes it kernel-overridable, I don't think we want to expose this in the public API. We should probably let the function take two attrsets. Also, I think we can pass through the builder flake and get it from there, rather than separate attributes. Finding these in `flake.nix` seems like the wrong place.

nix-builder/lib/gen-flake-outputs.nix
```diff
@@ -36,6 +42,31 @@ let
...
+      kernelDirty = self != null && !(self ? rev);
+      builderSha = if builderRev == null then null else lib.removeSuffix "-dirty" builderRev;
+      parts =
+        lib.optional (kernelSha != null) "--kernel-sha ${kernelSha}"
+        ++ lib.optional (kernelSha != null && kernelDirty) "--kernel-dirty"
+        ++ lib.optional (builderSha != null) "--kernel-builder-sha ${builderSha}"
+        ++ lib.optional (builderSha != null && builderDirty) "--kernel-builder-dirty";
+    in
+    lib.concatStringsSep " " parts;
+
```
danieldk: As mentioned above in `arch.nix`, this is a very local concern, doing this `gen-flake-outputs.nix` seems like the wrong level of abstraction.

sayakpaul:
Will break the PR into multiple parts, starting with appending the builder info, as discussed. 

Will wait for the metadata versioning PR, first before continuing. This so that we don't have to make the provenance data optional,

kernel-builder/src/pyproject/ops_identifier.rs
sayakpaul: Addressed.

kernel-builder/src/upload.rs
sayakpaul: Done.

kernel-builder/src/pyproject/mod.rs
sayakpaul: Will table it for later as mentioned.

We go with our approach and document the bad practice of running `create-pyproject` and committing after that.

kernel-builder/src/pyproject/common.rs
sayakpaul: Done.

danieldk (approved):
Great! Let's merge this and then rebuild all kernels-community kernels once the Torch 2.13 RC 15 PR is also merged (working on it).

PR #676, opened 2026-06-25
nix-builder: redesign kernel testing

danieldk:
This PR combines several changes to improve testing of kernels.

What we did so far: in tests for e.g. flash-attn3, we imported `flash_attn3` in tests. However, this does not properly test the real-world use case of kernels, dynamically loaded kernels. It also lead to issues in some cases because it relies on the compatibility module being used. The compatibility module sometimes triggers a reload of a part of a kernel (first through the randomized name and then again through the regular name), resulting in operator registration clashes, etc.

What we will do instead: use `get_kernels` in tests. In development/tests shells set `LOCAL_KERNELS` to ensure that the local kernel gets tested.

Concrete changes:

- Set `LOCAL_KERNELS` in `nix develop` and `nix develop .#test`.
- Add a `setup.py build_kernel` subcommand to arch builds as a replacement for `pip install`. This subcommand was already supported by noarch kernels.
- Update the documentation to use `setup.py build_kernel` in place of pip install.
- Add a note about using `get_kernel` to the _Kernel tests_ section.


docs/source/builder/agents-guide.md
```diff
@@ -31,14 +31,14 @@ kernel-builder skills add --claude
 
 Writing kernels is a hard problem, so be specific to agents. A robust prompt will declare all core attributes, including:
 
-- the library, for example `transformers` or `diffusers`
```
sayakpaul, line 34: How were doc related formatting changes started showing up? Any version updates?

docs/source/builder/agents-guide.md
```diff
@@ -69,7 +69,7 @@ examples/your_model/
 │   └── torch_binding.cpp       # PyTorch C++ bindings
 ├── benchmark_rmsnorm.py        # Micro-benchmark script
 ├── build.toml                  # kernel-builder config
-├── setup.py                    # pip install -e .
+├── setup.py                    # python setup.py build_kernel
```
sayakpaul, line 72: Does this mean that the existing skill files need to be updated?

docs/source/builder/build.md
```diff
@@ -69,25 +69,27 @@ for monitoring the build. The compiled kernel will then be in the local
...
 ```
 
-If you want to test the kernel as a Python package, you can do so.
-`kernel-builder devshell` will automatically create a virtual environment in
-the `.venv` and activate it. You can install the kernel as a regular
-Python package in this virtual environment:
+This will build the kernel and puts the output in the `build` directory
+and can be used with the `kernels` library.
+
+Noarch (JIT-compiled) kernels do not use CMake. For these reason, we also
```
sayakpaul:
```suggestion
Noarch (JIT-compiled) kernels do not use CMake. For this reason, we also
```

docs/source/builder/writing-kernels.md
```diff
@@ -447,10 +409,35 @@ def relu_fwd_fake(input: torch.Tensor) -> torch.Tensor:
...
+Kernel tests are stored in the `tests` directory. Tests must not use direct
+imports, but instead use `get_kernel` to test the kernel as it will be used.
+For example:
+
+```python
+import kernels
+import torch
+import torch.nn.functional as F
+
+relu = kernels.get_kernel("kernels-community/relu", version=1)
```
sayakpaul, line 423: This assumes that the kernel has been pushed to the Hub and the tests require loading remotely and NOT locally? Should we perhaps make that point a bit clearer?

kernel-builder/src/pyproject/templates/torch/setup.py
```diff
@@ -140,6 +148,41 @@ def build_extension(self, ext: CMakeExtension) -> None:
...
+        super().initialize_options()
+
+    def finalize_options(self) -> None:
+        super().finalize_options()
+
+    def run(self) -> None:
+        project_root = Path(__file__).parent
+
+        debug = int(os.environ.get("DEBUG", 0))
+        cfg = "Debug" if debug else "Release"
```
sayakpaul, line 167: Do we need to document this env var?

sayakpaul (commented):
Nice!

I think it'd be nice to have a short section in `local-dev.md` or `build.md` covering:

* what the command does under the hood (the local_install target, arch vs. noarch)
* the supported env vars (DEBUG, MAX_JOBS, NVCC_THREADS, ccache/sccache).

docs/source/builder/agents-guide.md
danieldk: Possibly my editor reformatting these superfluous EOL whitespaces. If it's compatible with hf doc builder, we could consider using `prettier` to standardize formatting.

docs/source/builder/writing-kernels.md
danieldk: It doesn't, see the note about `LOCAL_KERNELS` below.

kernel-builder/src/pyproject/templates/torch/setup.py
danieldk: This is part of the standardized CMake/setuptools integration stuff. I am not sure if we really want to document it? We should probably come up with a more principled way of enabling debugging if that's what we want.

sayakpaul (approved):
Thanks!

Just one question:
https://github.com/huggingface/kernels/pull/676#discussion_r3475201132

docs/source/builder/agents-guide.md
danieldk: I think so, I'll make a separate PR for that to move this forward.

PR #680, opened 2026-06-26
kernel-builder: support `cxx-flags` for Torch and tvm-ffi sources

danieldk:
We supported the `cxx-flags` option on kernel components, but not on
framework files. This change rectifies that.


drbh (approved):
nice looks good to me!

PR #681, opened 2026-06-30
Add support for Torch 2.13

danieldk:
Note: there is still an issue with ROCm hanging on importing `torch` on single GPU and no GPU machines, the Torch update is commented out for this reason. I propose that we focus on merging this anyway, so that we can rebuild all non-ROCm kernels already.

### DIFF
--- a/nix-builder/pkgs/python-modules/pyzes/default.nix
+++ b/nix-builder/pkgs/python-modules/pyzes/default.nix
@@ -1,0 +1,39 @@
+{
+  lib,
+
+  buildPythonPackage,
+  fetchPypi,
+  setuptools,
+
+  python,
+
+  level-zero,
+}:
+
+buildPythonPackage rec {
+  pname = "pyzes";
+  version = "0.1.2";
+  pyproject = true;
+
+  src = fetchPypi {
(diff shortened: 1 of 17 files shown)


nix-builder/pkgs/python-modules/pyzes/default.nix
```diff
@@ -0,0 +1,39 @@
+{
```
sayakpaul, line 1: Where is this module needed?

nix-builder/pkgs/xpu-packages/intel-deep-learning-2026.0.0.json
```diff
@@ -0,0 +1,792 @@
+{
```
sayakpaul, line 1: 👀

sayakpaul (dismissed):

nix-builder/pkgs/python-modules/pyzes/default.nix
danieldk: It's a new dependency of Torch on XPU.

drbh (approved):

PR #682, opened 2026-06-30
feat: scope stable abi to backend

drbh:
this is a work in progress PR that scopes torches stable abi to individual back ends rather than having it be a global attribute.


kernel-builder/src/util.rs
```diff
@@ -56,9 +56,37 @@ pub(crate) fn parse_and_validate(kernel_dir: impl AsRef<Path>) -> Result<Current
...
+    validate_stable_abi(&build)
+        .wrap_err_with(|| format!("Invalid configuration in {}", build_toml.to_string_lossy()))?;
+
     Ok(build)
 }
 
+/// Validate that a per-backend `[torch.stable-abi]` mapping only references
+/// backends that are declared in `[general].backends`.
+fn validate_stable_abi(build: &CurrentConfig) -> Result<()> {
+    let v4::Framework::Torch(torch) = &build.framework else {
```
danieldk: I know this is still in draft, but `vN` should not spread across the code base. This validation should probably be in the `v4` file itself in the deserialization of `Build` (since that holds both `General` and `Framework`).

Though I'm not 100% sure if this is worth validating. It's nice to be able to remove a backend from the list for quick testing without removing every other occurrence of that backend in `build.toml`.

nix-builder/lib/kernel-config.nix
```diff
@@ -15,6 +15,19 @@ let
...
+    backend:
+    let
+      cfg = lib.attrByPath [ "torch" "stable-abi" ] null toml;
+    in
+    if cfg == null then
+      null
+    else if builtins.isString cfg then
+      cfg
+    else
+      cfg.${backend} or null;
```
danieldk:
```suggestion
  torchStableAbiVersionForBackend =
    backend:
    let
      cfg = lib.attrByPath [ "torch" "stable-abi" ] {} toml;
    in
    if builtins.isString cfg then
      cfg
    else
      cfg.${backend} or null;
```

kernels-data/src/config/mod.rs
```diff
@@ -195,10 +200,30 @@ pub struct Torch {
...
 }
 
+/// Torch stable ABI version: a single version for all backends, or per-backend
+/// versions. Backends absent from a mapping are built without the stable ABI.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum TorchAbi {
+    All(Version),
+    Mapping(BTreeMap<Backend, Version>),
+}
```
danieldk, line 216: Pasted from Slack to have it in the review:

I thought a bit more about the extended Stable ABI mapping. Let's do this:

• Make v5 of the Build format. In v4 it stays String, in v5 we make it the mapping.
• Conversion from v4 to v5 is simple: when stable-abi is set, enumerate over the backends and set the stable ABI for the available backends..
• If we make v5, we should add `edition` to `general` and make it mandatory to set `edition = 5`. In this way we can start to unambiguously parse different versions in the future..
• If we don't detect `edition`, ask the user to run `kernel-builder update-build` (or maybe auto-update from v4 to v5, not on disk, but just in the builder, so that it works without running `update-build`).

By doing this, we avoid a lot of conditionals that check whether stable-abi is a string or mapping/attrset.

What do you think?

drbh: these are great recommendations

1. added v5 in latest changes with edition field and expects a mapping for stable abi
2. added logic to upgrade v4 to v5 in memory so existing stable-abi="string" should work without issues
3. updated `update-builder` to target updating to v5

these changes also include config changes to all of the examples so there are many more files changed in latest but most are one line additions to configs

drbh:
this kernel community PR uses this branch to validate the new version/scoped stable abi builds 
- https://github.com/huggingface/kernels-community/pull/1007

kernel-builder/src/pyproject/templates/torch/preamble.cmake
```diff
@@ -156,7 +155,7 @@ if(GPU_LANG STREQUAL "CUDA")
   add_compile_definitions(USE_CUDA=1)
   add_compile_definitions(CUDA_KERNEL)
 elseif(GPU_LANG STREQUAL "HIP")
-  if(NOT HIP_FOUND)
+  if(NOT HIP_FOUND AND NOT PYTORCH_FOUND_HIP)
```
danieldk, line 158: Why is this additional condition needed? Seems to make it less strict?

kernels-data/src/config/compat.rs
```diff
@@ -20,10 +33,35 @@ impl<'de> Deserialize<'de> for BuildCompat {
...
+            .map_err(|err| de::Error::custom(format!("invalid `general.edition`: {err}")))?
+            .general
+            .and_then(|general| general.edition);
+
+        match edition {
+            // Configs with editions
+            Some(5) => v5::Build::deserialize(value)
+                .map(BuildCompat::V5)
+                .map_err(de::Error::custom),
+
```
danieldk, line 46: Nice!

kernels-data/src/config/v5.rs
```diff
@@ -0,0 +1,586 @@
...
+
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub struct General {
+    pub name: KernelName,
+
+    pub version: usize,
+
+    /// Build format edition. Must be `5` for this schema.
+    pub edition: usize,
```
danieldk: Would be nice to hard-set this to 5. The serde author has a crate for this:

https://crates.io/crates/monostate

danieldk (commented):
Looks great!

One question about a CMake change, plus a suggestion to hardcode the version in the `v5` `General`.

kernel-builder/src/pyproject/templates/torch/preamble.cmake
drbh: I ran into a build issue with rocm which seems related to torch 2.13. It seems like they no longer expose `HIP_FOUND` and now expose `PYTORCH_FOUND_HIP` only

build issue:
https://github.com/huggingface/kernels-community/actions/runs/28620863071/job/84876068475

pr with changes to `HIP_FOUND` 
https://github.com/pytorch/pytorch/commit/d921fd000eef2ff35b3f0cb89cde8d988e35b47e

`HIP_FOUND` is not longer found in the codebase when searching on main

I opt'ed to adding `PYTORCH_FOUND_HIP` instead of replacing in order to avoid any regressions with older torch versions, this way we will only throw if both `HIP_FOUND`  and `PYTORCH_FOUND_HIP` are false

kernels-data/src/config/v5.rs
drbh: ah nice improvement, I've added in the latest commits.

danieldk (approved):

PR #683, opened 2026-07-01
skills: update to use new testing methodology

danieldk:
- Use `setup.py build_kernel`, not `pip install`.
- Give hints on how to load the kernel.


sayakpaul (approved):
Very nice!

PR #684, opened 2026-07-01
kernel-builder: ensure that `HIP_ARCHITECTURES` gets properly cleared

danieldk:
We want to clear `HIP_ARCHITECTURES`, on the shared library target, since we want to set archs per kernel component. However, conditioning this on `GPU_ARCHES` caused the property clearing to never happen. Since we do not use `GPU_ARCHES` anymore, rip it out.

Also add tests that verify kernel archs and kernel archs intersections to avoid such regressions in the future.

### DIFF
--- a/examples/kernels/relu-archs-subset/build.toml
+++ b/examples/kernels/relu-archs-subset/build.toml
@@ -1,0 +1,45 @@
+[general]
+name = "relu-archs-subset"
+version = 1
+license = "Apache-2.0"
+backends = [
+    "cuda",
+    "rocm",
+]
+
+[general.hub]
+repo-id = "kernels-test/relu-archs-subset"
+
+[torch]
+src = [
+    "torch-ext/torch_binding.cpp",
+    "torch-ext/torch_binding.h",
+]
+
+# The capability/architecture sets below intentionally contain entries that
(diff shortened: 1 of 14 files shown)


examples/kernels/relu-archs-subset/build.toml
```diff
@@ -0,0 +1,45 @@
...
+src = [
+    "torch-ext/torch_binding.cpp",
+    "torch-ext/torch_binding.h",
+]
+
+# The capability/architecture sets below intentionally contain entries that
+# are not supported by the toolchain versions used in CI (e.g. CUDA 10.0/12.0
+# capabilities require CUDA >= 12.8, gfx940 is not in the supported arch set).
+# CI checks that the built kernels contain exactly the intersection with the
+# supported sets.
```
sayakpaul, line 23: Do we reflect this in `build_kernel_rocm.yaml`?

I think we only test build?
https://github.com/huggingface/kernels/blob/b79face32a072b46e4e95c37869b3d1cc2e7735f/.github/workflows/build_kernel_rocm.yaml#L40-L41

danieldk:
```
nix build -L ./examples/kernels#ci-build-rocm
```

Builds all the test kernels and these tests fire on builds.

sayakpaul (approved):

PR #685, opened 2026-07-02
feat: implement kernels info

### LINKED ISSUE #674: Add a `kernels info <name>` command to describe a kernel
It would be nice to extend the `kernels` CLI with a command to describe a kernel from a local path or a Hub ID, e.g. `kernels info <name>`.

This came out of a broader discussion on improving the `kernels` CLI experience (partly motivated by making kernels more amenable to agentic workflows, similar to [kernel-builder PR #636](https://github.com/huggingface/kernels/pull/636/)).

Context on related ideas:
- **Checking variant compatibility for a given machine** — already supported via `kernels versions <repo>` (which lists each variant and why it is/isn't compatible). A `--json` flag could be added for machine-readable output, though it's unclear how much that helps LLMs.
- **Describing a kernel** — not yet available; `kernels info <name>` would fill this gap by surfacing metadata/description for a kernel given a local path or Hub ID.

---
Requested by Sayak Paul - [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1781927649766039?thread_ts=1781927649.766039&cid=C090JN2P8NB) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-06-23T14-24-09-563Z_dc390e38-c55b-4c1f-a531-c4f33d56f7ec.jsonl)

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/674

Example: 

```bash
kernels info kernels-community/flash-attn3 
```

```bash
Repository: kernels-community/flash-attn3
Revision: v1
Name: flash-attn3
Version: 1
License: BSD-3-Clause
Upstream: -
Source: -
Python dependencies: -
Backends: cuda
Variants:

torch-stable-abi29-cu128-x86_64-linux: CPU (x86_64) does not match system CPU (aarch64)
torch-stable-abi29-cu126-x86_64-linux: CPU (x86_64) does not match system CPU (aarch64)
torch-stable-abi29-cu130-x86_64-linux: CPU (x86_64) does not match system CPU (aarch64)
torch211-cxx11-cu128-x86_64-linux: CPU (x86_64) does not match system CPU (aarch64)
torch210-cxx11-cu128-x86_64-linux: CPU (x86_64) does not match system CPU (aarch64)
torch212-cxx11-cu130-x86_64-linux: CPU (x86_64) does not match system CPU (aarch64)
torch211-cxx11-cu130-x86_64-linux: CPU (x86_64) does not match system CPU (aarch64)
torch210-cxx11-cu130-x86_64-linux: CPU (x86_64) does not match system CPU (aarch64)
```

(tested on a Macbook)

I decided to add info on the supported/unsupported variants for comprehensive info reporting. Can drop it if needed.


kernels/src/kernels/cli/info.py
```diff
@@ -0,0 +1,168 @@
...
+        "upstream": None,
+        "source": None,
+        "python_depends": [],
+        "backends": sorted(backends),
+        "variants": [
+            {
+                "variant": decision.variant.variant_str,
+                "compatible": isinstance(decision, VariantAccepted),
+                "reason": None if isinstance(decision, VariantAccepted) else decision.reason,
+            }
```
danieldk: This has strong overlap with the existing `kernels versions` subcommand, I would therefore leave it out (always make orthogonal functions). Also, since some kernels accumunate(d) a lot of variants over time, it would push away all the other information.

kernels/src/kernels/cli/info.py
```diff
@@ -0,0 +1,168 @@
...
+            backends.add(variant.arch.backend.name)
+        else:
+            backends.add(variant.arch.backend_name)
+
+    info: dict[str, Any] = {
+        "name": None,
+        "version": None,
+        "license": None,
+        "upstream": None,
+        "source": None,
```
danieldk: Why do we need these none/empty initializations? We can check if the field is there?

kernels/uv.lock
```diff
@@ -2,15 +2,12 @@ version = 1
 revision = 3
```
danieldk, line 2: `uv.lock` update seems unrelated to the PR?

danieldk (commented):
Nice! Added some comments.

kernels/src/kernels/cli/info.py
sayakpaul: Can you elaborate more on this suggestion? Didn't quite get what's wanted.

kernels/src/kernels/cli/info.py
sayakpaul: Addressed.

kernels/uv.lock
sayakpaul: Addressed.

kernels/src/kernels/cli/info.py
sayakpaul: Ah got it.

danieldk (approved):
Really nice, let's do it!

PR #694, opened 2026-07-07
feat: implement kernel uploads through PRs.

sayakpaul:
Example: https://huggingface.co/kernels/sayakpaul/activation-create-pr-test/discussions/1

In the subsequent PR, I will push modifications to `kernels-community` CI so that PRs are raised for kernels like https://huggingface.co/kernels/MiniMaxAI/msa


docs/source/builder/build.md
```diff
@@ -212,6 +212,28 @@ repo-id = "kernels-community/flash-attn4"
...
+
+```bash
+$ kernel-builder build-and-upload --create-pr
+```
+
+Instead of committing directly, this opens a pull request on the Hub for
+each branch that has changes: one against `main` for the kernel card and
+one against the `v<version>` branch for the build artifacts and
+benchmarks. The pull request URLs are printed after the upload, so that
+the repository maintainers can be pointed to them for review.
```
sayakpaul, line 230: Happy to also scope it to just version branches if we want.

docs/source/builder/writing-kernels.md
```diff
@@ -211,7 +211,10 @@ The following sections enumerate all supported options for `build.toml`.
 ### `general.hub`
 
 - `repo-id`: the Hub repository to upload the kernel to when the `upload` or
-  `build-and-upload` subcommands of `kernel-builder` are used.
+  `build-and-upload` subcommands of `kernel-builder` are used. If you do not
+  have write access to the repository, both subcommands accept the
+  `--create-pr` option to propose the upload through a pull request instead
+  (see [Uploading through a pull request](build.md#uploading-through-a-pull-request)).
```
danieldk: I don't think this should be in the description of the `build.toml`. I think it _is_ useful in this file though. Maybe we can add a section add the end called something like _Uploading a kernel to a third-party organization_?

kernel-builder/src/main.rs
```diff
@@ -427,6 +434,39 @@ fn main() -> Result<()> {
...
+    fn test_build_and_upload_create_pr_flag() {
+        let cli =
+            Cli::try_parse_from(["kernel-builder", "build-and-upload", "--create-pr"]).unwrap();
+        match cli.command {
+            Commands::BuildAndUpload { create_pr, .. } => assert!(create_pr),
+            _ => panic!("Expected build-and-upload subcommand"),
+        }
+    }
+}
+
```
danieldk: Not 100% sure what we are testing here. Isn't this already tested in the tests of the `clap` crate itself?

kernel-builder/src/upload.rs
```diff
@@ -171,12 +179,15 @@ fn run_upload_typed<T: RepoType>(args: UploadArgs) -> Result<()> {
...
-        .to_owned();
+        .as_ref()
+        .and_then(|repo_url| {
+            repo_url
+                .url
+                .trim_end_matches('/')
+                .strip_prefix("https://huggingface.co/")
+                .map(|s| s.strip_prefix("kernels/").unwrap_or(s).to_owned())
+        })
+        .unwrap_or(repo_id);
```
danieldk: This reads weird, especially the `unwrap_or(repo_id)` at the end. Unless I'm misreading, isn't this:


```suggestion
        .as_ref()
        .map(|repo_url| {
            repo_url
                .url
                .trim_end_matches('/')
                .strip_prefix("https://huggingface.co/")
                .map(|s| s.strip_prefix("kernels/").unwrap_or(s).to_owned())
        });
```

kernel-builder/src/upload.rs
```diff
@@ -188,10 +199,21 @@ fn run_upload_typed<T: RepoType>(args: UploadArgs) -> Result<()> {
...
+            if args.create_pr {
+                created.wrap_err_with(|| {
+                    format!(
+                        "Pull requests can only target an existing branch. Ask a \
+                         maintainer of `{repo_id}` to create the branch `{branch}` first."
+                    )
+                })?;
+            } else {
+                created?;
+            }
```
danieldk:
```suggestion
                .wrap_err_with(||
                if args.create_pr {
                  format!(
                        "Pull requests can only target an existing branch. Ask a \
                         maintainer of `{repo_id}` to create the branch `{branch}` first."
                    )
                }
                else {
                    format!("Cannot create branch `{branch}`"))
                 }
                 );
```

Minus typos and with correct formatting, but you get the gist :).

kernel-builder/src/upload.rs
```diff
@@ -281,21 +309,43 @@ fn run_upload_typed<T: RepoType>(args: UploadArgs) -> Result<()> {
                 "Uploaded using `kernel-builder`.".to_owned()
             };
 
-            repo.create_commit()
+            let open_pr = args.create_pr && pr_revision.is_none();
+            let revision = pr_revision.clone().unwrap_or_else(|| branch.clone());
```
danieldk, line 312: Meh, it's annoying that `hf-hub` doesn't take `&str` or `impl ToOwned`, forcing all this work on the caller. Could be a much cleaner `pr_revision.unwrap_or(&branch)` otherwise.

Maybe worth raising an issue?

danieldk (commented):
Some small suggestions

docs/source/builder/writing-kernels.md
sayakpaul: Where is it being added to the description of `build.toml` though? I guess what you meant is that the description shouldn't lie in `build.md` and should only be in `writing-kernels.md`? If so, which guide between the two (`build.md` and `writing-kernels.md`) talks more about uploads? I think it's `build.md` since it has an entire section on https://huggingface.co/docs/kernels/main/en/builder/build#uploading-your-kernel-to-the-hub? On the other hand, `writing-kernels.md` almost doesn't talk about the upload stuff.

kernel-builder/src/upload.rs
sayakpaul: https://github.com/huggingface/hf-hub/issues/184

docs/source/builder/writing-kernels.md
danieldk: > Where is it being added to the description of build.toml though?

This part of the document is about `build.toml` and the available fields in `build.toml`. (_The following sections enumerate all supported options for `build.toml`._)

> If so, which guide between the two (build.md and writing-kernels.md) talks more about uploads? I think it's build.md since it has an entire section on https://huggingface.co/docs/kernels/main/en/builder/build#uploading-your-kernel-to-the-hub?

Yeah, then probably `build.md` is the best place?

kernel-builder/src/upload.rs
```diff
@@ -160,7 +165,10 @@ fn run_upload_typed<T: RepoType>(args: UploadArgs) -> Result<()> {
         .exist_ok(true)
         .send()
     {
-        Ok(url) => url,
+        Ok(url) => Some(url),
+        // With --create-pr, write access is not required: the repository only
+        // needs to exist, so continue and let the Hub open a pull request.
+        Err(HFError::Forbidden { .. }) if args.create_pr => None,
```
drbh: maybe we can check if `args.create_pr` exists, and skip the repo creation call instead of swallowing the error when it fails

kernel-builder/src/upload.rs
drbh: maybe we can avoid changing this logic by skipping the repo url logic all together when the create pr flag is used. I think we are only operating on the repo_url response in the case the repo creation returned a sanitized repo id. 

since we shouldn't create a new repo when the create pr flag is used maybe when the flag exists we simply use the provided repo id?

drbh (commented):
looking good, just left some comments/nits

kernel-builder/src/upload.rs
sayakpaul: I believe this should have been addressed by now.

drbh: nice! looks great

drbh (approved):
lgtm

PR #702, opened 2026-07-08
Update nixpkgs

danieldk:


nix-builder/lib/build-sets.nix
```diff
@@ -1,10 +1,6 @@
 {
   nixpkgs,
   rust-overlay,
-
-  # Git provenance (`{ sha, dirty }` or `null`) of the `kernel-builder` flake,
-  # so it can be burned into the binary.
-  builderProvenance ? null,
 }:
```
sayakpaul, line 9: Unused?

nix-builder/overlay.nix
```diff
@@ -44,6 +38,17 @@ final: prev:
...
   );
 
+  # nixpkgs overrides
+  ucc = prev.ucc.overrideAttrs (
+    _: prevAttrs: {
+      # Concurrent build flags do not seem to get passed through? So instead, let's
+      # add parallelism to building CUDA files.
+      preConfigure = prevAttrs.preConfigure + ''
+        export NVCC_CFLAGS="--threads $NIX_BUILD_CORES"
+      '';
```
sayakpaul, line 55: I wonder if we could benefit from this in the CI of `kernels-community` and the GitHub Actions workflow we have for building? Cc: @drbh

sayakpaul (dismissed):

nix-builder/overlay.nix
danieldk: This is already done automatically when building kernels using the nvcc threads hook.

nix-builder/lib/build-sets.nix
danieldk: Indeed, probably, probably a leftover from when we split things into separate files.

sayakpaul (approved):

PR #703, opened 2026-07-08
feat: warn for dirty builds.

sayakpaul:
Follow up of https://github.com/huggingface/kernels/pull/666/

### DIFF
--- a/kernel-builder/src/upload.rs
+++ b/kernel-builder/src/upload.rs
@@ -150,6 +150,9 @@ fn run_upload_typed<T: RepoType>(args: UploadArgs) -> Result<()> {
         build_dir.display()
     );
 
+    let dirty_variants = dirty_variant_names(&variants);
+    warn_dirty_variants(&dirty_variants);
+
     let (repo_id, branch) = get_repo_and_branch(&kernel_dir, args.repo_id, args.branch, &variants)?;
 
     let repo_url = match api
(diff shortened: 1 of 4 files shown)


kernel-builder/src/upload.rs
```diff
@@ -363,16 +367,79 @@ fn collect_benchmark_commit_ops(
...
+        }
+        // If the card cannot be read as UTF-8, fall back to uploading it
+        // verbatim rather than dropping the README entirely.
+        Err(_) => AddSource::File(card_path),
+    };
     operations.push(CommitOperation::Add {
         path_in_repo: "README.md".to_owned(),
-        source: AddSource::File(card_path),
+        source,
     });
```
drbh: instead of commit this info and adding to the readme, maybe it would be best to surface this information in the hub? 

maybe we can add a component that lists the builds and if they are dirty or not? this way we can avoid storing this info as a string in the readme

sayakpaul: Copying over what I said on Slack:

> I preferred README because it's a cheap thing to do without adding any more logic to the Hub side backend. Plus also unlike supported accelerator devices (which is best determined after the build is done), we can almost know from the getgo if the build is dirty or not. So, preferred to do that in the README. LMK.

drbh (approved):
nice! lgtm

PR #708, opened 2026-07-09
fix: show all variants in e2e test

drbh:
this avoids filtering out torch 213 variants in the e2e tests

### DIFF
--- a/.github/workflows/test_e2e.yaml
+++ b/.github/workflows/test_e2e.yaml
@@ -84,8 +84,7 @@ jobs:
         id: variant
         run: |
           cd /tmp/${{ env.E2E_REPO_NAME }}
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):
So, the underlying machine became incompatible somehow with the variants?

drbh:
yea I think its an issue based on how the e2e tests choose the variant and torch version that conflicts if a new torch version is released.

for example I think we first choose the VARIANT, then later uv pip install torch, so in the case the variant is an older version than what gets pip installed an error gets thrown

https://github.com/huggingface/kernels/blob/2c516fc6678570c013319adeb64605fc8e8d5394/.github/workflows/test_e2e.yaml#L129-L136

PR #714, opened 2026-07-13
Add TPU backend support (torch_tpu / Pallas kernels)

### LINKED ISSUE #711: [TPU] I want to add support for torch tpu and pallas kernels
### Description

There is anew backend for TPUs, torch_tpu. While the release is not public yet, I want to be able to use kernels to use pallas kernels with this library, in particular to combine them with transformers. I want to add the necessary tools to add kernels.

### Motivation

The torch_tpu team is interesting in delivering a solution to provide better performance on transformers on TPU, and kernels is part of the equation. This will allow to use the kernels rather than custom code and distribute it using our hub.

### Proposed approach

I will open a PR that proposes an implementation.

### Alternatives considered

_No response_

### Additional context

_No response_

tengomucho:
Closes #711.

Adds TPU as a first-class backend across the stack, so a kernel repo declaring `backends = ["tpu"]` builds and loads a `torch-tpu` variant (Pallas/JAX kernels running through torch_tpu):

- **kernels**: TPU backend/device detection via `torch.backends.tpu`, `torch-tpu` noarch variant resolution, `kernelize` device support, `tpu_only` pytest marker.
- **kernels-data**: `Backend::Tpu` enum variant with serde/pyo3 bindings and TPU `python_depends` in backend metadata.
- **kernel-builder**: `tpu` accepted in `init --backends`, TPU detection in the noarch `_ops.py` template.
- **nix-builder**: TPU build set (jax, jaxlib, libtpu, torch_tpu), bundle/CI outputs, and the `torch-tpu` build variant. This part was made and tested with the help of Fable.
- **examples**: `relu-tpu` example kernel repository.

TPU support is experimental (Tier 3): the libtpu/torch_tpu wheels are not public yet and are fetched from Google's Artifact Registry using a gcloud token (`GCLOUD_ACCESS_TOKEN`), so `relu-tpu` is excluded from the examples CI for now (until we have access to the TPUs for CI).

Note: `KNOWN_BACKENDS` in `kernels.utils` intentionally does not gain `tpu`, any value change hits an API compatibility lint error in the CI.

## Checklist

- [x] This PR is linked to an issue that was discussed and approved
- [x] I have tested these changes locally
- [x] New/changed functionality has test coverage
- LLM disclosure:
  - [ ] I did not use an LLM to create this PR.
  - [x] I used and LLM for assistance while creating this PR.
  - [ ] This PR was mostly or completely generated by an LLM.


drbh (changes requested):
looking good! just left some comments around fetching deps and the kernel-builder init support

kernel-builder/src/init.rs
```diff
@@ -44,7 +44,7 @@ pub struct InitArgs {
     #[arg(long, value_name = "OWNER/REPO")]
     pub name: Option<RepoInfo>,
 
-    /// Backends to enable (`all`, `cpu`, `cuda`, `metal`, `neuron`, `rocm`, `xpu`).
+    /// Backends to enable (`all`, `cpu`, `cuda`, `metal`, `neuron`, `rocm`, `tpu`, `xpu`).
```
drbh: if we want to add support to initialize a tpu repo, I think we need to update `build_init_fileset` to handle `has_tpu` and render a tpu/noarch scaffold, including `[general.tpu]`, `[torch-noarch]`, and a tpu-specific Python module template. (similar to the other backends). otherwise we should probably leave it out of this list

nix-builder/pkgs/python-modules/libtpu/default.nix
```diff
@@ -0,0 +1,66 @@
...
+    netrcPhase = ''
+      if [ -z "''${GCLOUD_ACCESS_TOKEN:-}" ]; then
+        echo "GCLOUD_ACCESS_TOKEN is not set; cannot fetch libtpu." >&2
+        echo "Run: export GCLOUD_ACCESS_TOKEN=\$(gcloud auth print-access-token)" >&2
+        exit 1
+      fi
+      printf 'machine us-python.pkg.dev\nlogin oauth2accesstoken\npassword %s\n' \
+        "$GCLOUD_ACCESS_TOKEN" > netrc
+    '';
+  };
```
drbh, line 53: would it be possible to fetch this from pypi? I believe this is the library https://pypi.org/project/libtpu, maybe we can use `fetchPypi`?

nix-builder/scripts/helpers/get_torch_tpu_hash.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+# Print the wheel URL and SRI hash for a torch-tpu-virtual-registry
```
drbh: it would probably be best to avoid adding this file, I think we only need it since the `torch_tpu` library requires auth. 

in the case we move `libtpu` to pypi and only fetch `torch_tpu` from the google registry - I think we might be able to add the refresh command as a comment to the `torch_tpu` nix file? we might be able to get the hash with a one liner like below - and that would remove the need for this helper

```bash
nix store prefetch-file \
--json \
--hash-type sha256 \
--option netrc-file =({ printf 'machine us-python.pkg.dev\nlogin oauth2accesstoken\npassword '; gcloud auth print-access-token; }) \
'https://us-python.pkg.dev/ml-oss-artifacts-transient/torch-tpu-virtual-registry/torch-tpu/torch_tpu-0.1.1.dev20260707090224-cp313-cp313-manylinux_2_31_x86_64.whl'  \
| jq -r .hash
```

nix-builder/versions.nix
```diff
@@ -58,6 +58,12 @@
     systems = [ "x86_64-linux" ];
     bundleBuild = true;
   }
+  {
+    torchVersion = "2.11";
+    tpu = true;
+    systems = [ "x86_64-linux" ];
+    bundleBuild = true;
+  }
```
drbh, line 18: would it be possible to also add entries 2.12 and 2.13 (not sure if there is a specific reason we only target 2.11)

kernel-builder/src/init.rs
sayakpaul: Slightly related: neuron is already in that list and also has no `has_neuron` or template. Maybe we should remove it from here.

sayakpaul (commented):
Nothing much from my side. I will have to defer the nix related bits to @danieldk @drbh. But the non-Nix related bits look good to me.

nix-builder/versions.nix
tengomucho: Currently, torch_tpu is only compatible with 2.11, otherwise there is an ABI error (see https://github.com/google-pytorch/torch_tpu/issues/1533 if you have access)

nix-builder/pkgs/python-modules/libtpu/default.nix
tengomucho: The version available in pypi is an old version, Google hosts newer`libtpu` version on an internal index reachable with token, and torch_tpu depends on those (at least for now).

nix-builder/scripts/helpers/get_torch_tpu_hash.sh
tengomucho: I just pushed a commit that addressed you comment.

kernel-builder/src/init.rs
tengomucho: You are right @sayakpaul : but I think the neuron part should be addressed on another issue and PR.

nix-builder/pkgs/python-modules/libtpu/default.nix
```diff
@@ -0,0 +1,72 @@
...
+    hash = "sha256-X5LVmwJuRcNkMG+m7kKgcaOSKnWjeuw4HtCU+lfWamY="; # cp313
+    netrcImpureEnvVars = [ "GCLOUD_ACCESS_TOKEN" ];
+    netrcPhase = ''
+      if [ -z "''${GCLOUD_ACCESS_TOKEN:-}" ]; then
+        echo "GCLOUD_ACCESS_TOKEN is not set; cannot fetch libtpu." >&2
+        echo "Run: export GCLOUD_ACCESS_TOKEN=\$(gcloud auth print-access-token)" >&2
+        exit 1
+      fi
+      printf 'machine us-python.pkg.dev\nlogin oauth2accesstoken\npassword %s\n' \
+        "$GCLOUD_ACCESS_TOKEN" > netrc
```
danieldk: This requires an impure build though, we should avoid that people start doing impure builds, so I would remove this. Instead it's probably best to use `requireFile`:

https://nixos.org/manual/nixpkgs/stable/#requirefile

And then add a utility script that uses `prefetch-file` using `GCLOUD_ACCESS_TOKEN`.

In CI we also have to make sure that the output path doesn't get added to our cache, or we are publicly distributing it. I think, but am not 100%, sure that it does not get uploaded to the cache if the prefetch is done before installing Cachix (since IIRC we have use Cachix in store monitoring mode).

danieldk: By the way, is there any chance that we can convince Google to make this available without a token?

tengomucho: Agreed on both. I'll switch to requireFile + a prefetch helper script for the actual fetch (similar to the one I have just removed after @drbh comment), so builds stay pure.
I also checked about the cache point, and tpu is currently marked bundleBuild = true, which puts it in `.#forCache`, built right after `cachix-action` installs its store-watcher in `update_cache.yaml`. Reordering the prefetch before Cachix would keep the raw wheel out, but torch_tpu's `autoPatchelfHook` step means the built package is itself a modified copy of gated content, so I don't think step-ordering alone is safe. I propose to leave `bundleBuild = false` for tpu for now so it's excluded from the shared cache pipeline entirely, until we get public packages for this.

docs/source/builder-cli.md
```diff
@@ -83,7 +83,7 @@ Initialize a new kernel project from template
 
   Default value: `Apache-2.0`
 * `--name <OWNER/REPO>` — Name of the kernel repo (e.g. `drbh/my-kernel`)
-* `--backends <BACKENDS>` — Backends to enable (`all`, `cpu`, `cuda`, `metal`, `neuron`, `rocm`, `xpu`)
+* `--backends <BACKENDS>` — Backends to enable (`all`, `cpu`, `cuda`, `metal`, `rocm`, `xpu`)
```
danieldk, line 86: No `tpu`?

examples/kernels/relu-tpu/build.toml
```diff
@@ -0,0 +1,16 @@
...
+version = 1
+edition = 5
+license = "Apache-2.0"
+backends = ["tpu"]
+
+[general.hub]
+repo-id = "kernels-test/relu-tpu"
+
+[general.tpu]
+python-depends = ["torch_tpu", "jax"]
```
danieldk: Shouldn't we assume that at least `torch_tpu` is always installed when the `tpu` backend is used? Maybe the same for Jax, not 100% sure. In general, we should stop adding more Python dependencies.

I think this should largely be encapsulated in the loading logic of `kernels` itself, similar to how tvm-ffi variants are considered when `tvm_ffi` can be imported.

We use dashes in our dependency declarations (we can let the mapping take care of translating to underscores.

kernel-builder/src/init.rs
```diff
@@ -44,7 +44,7 @@ pub struct InitArgs {
     #[arg(long, value_name = "OWNER/REPO")]
     pub name: Option<RepoInfo>,
 
-    /// Backends to enable (`all`, `cpu`, `cuda`, `metal`, `neuron`, `rocm`, `xpu`).
+    /// Backends to enable (`all`, `cpu`, `cuda`, `metal`, `rocm`, `xpu`).
```
danieldk, line 47: Shouldn't `tpu` be added here?

danieldk: Also, why is neuron removed? (Also see the removal of neuron at the start.)

kernels-data/src/config/v4.rs
```diff
@@ -49,6 +49,8 @@ pub struct General {
 
     pub python_depends: Option<Vec<String>>,
 
+    pub tpu: Option<TpuGeneral>,
```
danieldk: We don't add options to older `build.toml` versions once a new one is out, so adding to `v5` is enough.

kernels-data/src/python_dependencies.json
```diff
@@ -34,6 +34,20 @@
...
+        "python": [{ "pkg": "torch_tpu", "import": "torch_tpu" }]
+      },
+      "jax": {
+        "nix": ["jax"],
+        "python": [{ "pkg": "jax", "import": "jax" }]
+      },
+      "libtpu": {
+        "nix": ["libtpu"],
+        "python": [{ "pkg": "libtpu" }]
+      }
```
danieldk: I think of all these should be tested in kernels compat instead. Possibly with the exception of Jax.

We might also want to support pure Jax variants like `jax-tpu`, but that might be better in a follow-up PR.

nix-builder/pkgs/python-modules/libtpu/default.nix
```diff
@@ -0,0 +1,66 @@
...
+    message = ''
+      libtpu is served from a gated Google Artifact Registry and cannot
+      be fetched by a pure Nix build. Fetch and register it with:
+
+        GCLOUD_ACCESS_TOKEN=$(gcloud auth print-access-token) \
+          nix-builder/scripts/helpers/prefetch-tpu-wheel.sh libtpu ${version} ${abi}
+    '';
+  };
+
+  dependencies = [ ];
```
danieldk: Not used, so can be removed.

nix-builder/pkgs/python-modules/torch_tpu/default.nix
```diff
@@ -0,0 +1,91 @@
+{
+  lib,
+  autoPatchelfHook,
+  buildPythonPackage,
+  requireFile,
+  python,
+  stdenv,
+
+  # jax is not a hard dependency of the torch_tpu wheel, but TPU kernels
+  # import torch_tpu._internal.pallas (jax_op), which needs jax at import
```
danieldk: Aha, so if Jax always gets pulled in when importing `torch_tpu`, we also do not want Jax in `python_dependencies.json`.

nix-builder/versions.nix
danieldk: Oh, oops, we will be removing 2.11 in https://github.com/huggingface/kernels/pull/732 . Let me add enough of a 2.11 'skeleton' to make this work (basically setting other tritons, etc. to `null`).

danieldk (commented):
Great work! Added a bunch of comments.

docs/source/builder-cli.md
tengomucho: TPU backend support is in (detection, `Backend::Tpu`, Nix build variant, relu-tpu example). What's not here is kernel-builder init scaffolding. TPU kernels are pure Python/JAX (`torch-noarch` + `[general.tpu]`), not C++ torch-ext bindings like init generates, so advertising `--backends tpu` there would scaffold the wrong thing. Dropped it from init's help for now, we can add it later when/if we define a real has_tpu template path.

tengomucho: Note the same is valid for neuron

kernel-builder/src/init.rs
tengomucho: I explained it in [this comment](https://github.com/huggingface/kernels/pull/714#discussion_r3664821469).

examples/kernels/relu-tpu/build.toml
tengomucho: Agreed on jax and the naming. Fixed the dash issue (renamed the key to torch-tpu, still maps to the torch_tpu pkg/import).

On torch_tpu: the runtime check in kernels is indeed dead code (backend detection already requires torch_tpu to be present). But the same declaration also drives the nix-side dependency resolution in `nix-builder/lib/deps.nix`. That decides what's installed in the sandbox for `get-kernel-check` to import the kernel, nothing else pulls torch_tpu in there, so dropping it would break the nix build. Kept it for that reason.

danieldk: But TPU has it's own build variant (`tpu = true;`), so we can conditionally include any dependencies needed in the Torch derivation. It's the same principle as all the other backends, except rather than having a backend-specific Torch build, we have the CPU build with additional dependencies.

Now the outcome is quite weird. If you would `build .#torch.torch212-cxx11-tpu-x86_64-linux` in the top-level kernels repo, you would actually get a CPU-only build, not an XPU-enabled build.

kernels-data/src/python_dependencies.json
tengomucho: Right in principle, but one wrinkle: this file is also read directly by `nix-builder/lib/deps.nix` to pick the nix packages for the build sandbox (used by get-kernel-check), not just to generate the pip pyproject.toml extra. So we can't just drop torch-tpu/libtpu here without breaking the nix build. Doing this properly means splitting "nix build dep" from "pip-installable dep" in the schema. If we want to split that, alongside with jax variants, I would do it in another Issue+PR, wdyt?

examples/kernels/relu-tpu/build.toml
tengomucho: Agreed, torch_tpu's presence is already implied by backend detection `(hasattr(torch.backends, "tpu")` only goes true once torch_tpu is autoloaded), so declaring it as a dependency is redundant, and it's not on public PyPI anyway. Probably the same for libtpu. jax is the exception since it's public and not implied by detection. See [this comment](https://github.com/huggingface/kernels/pull/714#discussion_r3665869482) for why this isn't a trivial deletion though.

nix-builder/pkgs/python-modules/torch_tpu/default.nix
tengomucho: OK, I dropped both jax and libtpu from `python_dependencies.json`'s tpu block (only `torch-tpu` remains). Both are already hard, propagated Nix deps of `torch_tpu` itself. Verified end-to-end with a real build (`nix build .../relu-tpu#packages.x86_64-linux.redistributable."torch-tpu"`, gated wheels fetched via `prefetch-tpu-wheel.sh`): torch_tpu still imports fine (jax/libtpu pulled in transitively), and get-kernel-check loads the built relu_tpu kernel successfully. No regression.

kernels-data/src/config/mod.rs
```diff
@@ -444,3 +458,53 @@ pub enum ConfigError {
     #[error("Cannot migrate configuration: {reason:?}")]
     Migration { reason: String },
 }
+
+#[cfg(test)]
+mod tests {
```
danieldk, line 463: These were TPU dependencies tests, but since the TPU dependencies are gone now, I kept this and modified it to check CUDA dependencies.

kernels-data/src/python_dependencies.json
danieldk: Fixed.

examples/kernels/relu-tpu/build.toml
danieldk: Fixed.

drbh (approved):
really cool new feature! thanks for working on this. looks good to me!

PR #716, opened 2026-07-14
feat: add ability to output upload results as json

drbh:
this pr adds a `--output-json` flag to the upload command so the upload results are emitted in a machine readable format

example output
```json
{
  "status": "pull_request_created",
  "repo_id": "drbh/example",
  "branch": "v1",
  "url": null,
  "pull_requests": [
    { "branch": "main", "url": "https://hf.co/kernels/drbh/example/discussions/8" },
    { "branch": "v1",   "url": "https://hf.co/kernels/drbh/example/discussions/9" }
  ]
}
```


sayakpaul (dismissed):
I think this is also very nice from an agentic perspective since agents love machine-readable formats. I think it would be nice to also include a test in this PR?

sayakpaul (approved):

PR #724, opened 2026-07-20
experiment: support rust kernels

drbh:
this pr explores extending the kernel builder to support kernels written in rust. at the moment kernels are expected to be written a cpp or pure python. this pr updates the build process to:
- add two new backends `rust-cpu` and `rust-cuda`
- enable defining kernels written in rust in the build.toml
- update cmake build step to compile rust with cargo 
- add cuda-oxide to rust build env

the goal of this experiment is to explore the recently released https://github.com/NVlabs/cuda-oxide library as a backend for hf kernels so developers would have the option to author kernels purely in rust. 

additionally this change set is used in a small proof of concept rust impl of flash attn2 using cuda-oxide and tvmffi bindings here https://github.com/drbh/flash-attn2-oxide 

the published result can be seen https://huggingface.co/kernels/drbh/flash-attn2-oxide and can be run in comparison to the reference kernel

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "apache-tvm-ffi>=0.1.12",
#     "kernels>=0.16.0",
#     "numpy>=2.5.1",
#     "torch>=2.13.0",
# ]
# ///
import math

import torch

oxide = get_kernel("drbh/flash-attn2-oxide", version=1, trust_remote_code=True)
reference = get_kernel("kernels-community/flash-attn2", version=2)

torch.manual_seed(0)
b, s, h, d = 2, 128, 8, 64
q = torch.randn(b, s, h, d, device="cuda", dtype=torch.float16)
k = torch.randn(b, s, h, d, device="cuda", dtype=torch.float16)
v = torch.randn(b, s, h, d, device="cuda", dtype=torch.float16)
scale = 1.0 / math.sqrt(d)

for causal in (False, True):
    # fwd is positional: (q, k, v, alibi_slopes, softmax_scale, is_causal,
    # window_size_left, window_size_right, softcap)
    out, lse = oxide.fwd(q, k, v, None, scale, causal, -1, -1, 0.0)
    ref = reference.flash_attn_func(
        q, k, v, dropout_p=0.0, softmax_scale=scale, causal=causal
    )
    ref = ref[0] if isinstance(ref, (tuple, list)) else ref

    max_diff = (out.float() - ref.float()).abs().max().item()
    ok = torch.allclose(out.float(), ref.float(), atol=2e-3, rtol=2.5e-2)
    print(f"causal={causal!s:<5} max|diff|={max_diff:.2e} {'ok' if ok else 'MISMATCH'}")
    assert ok

```

then run 

```bash
uv run example/compare.py
```

and it outputs

```
causal=False max|diff|=4.88e-04 ok
causal=True  max|diff|=4.88e-04 ok
```

> [!WARNING]
> this work is an experiment, and may change or be closed at any point in time


docs/source/builder/writing-kernels.md
```diff
@@ -352,6 +352,51 @@ are available:
...
+backend = "cpu"
+dsl = "rust"
+depends = []
+src = ["Cargo.toml", "Cargo.lock", "src"]
+```
+
+The crate is built by cargo as a staticlib and whole-archive linked into the
+extension, so it must export the `__tvm_ffi_*` symbols itself. Rust kernels
+currently require the `[tvm-ffi]` framework, and the project root must include
+a `Cargo.lock`.
```
danieldk, line 373: Just to be sure: Rust doesn't mark the linkage of any symbols to evade `dlopen`'s `RTLD_LOCAL` right? I wouldn't expect so, but just to be sure.

docs/source/builder/writing-kernels.md
```diff
@@ -352,6 +352,51 @@ are available:
...
+use `dsl = "cuda-oxide"` and point at the device manifest:
+
+```toml
+[kernel.my_kernel]
+backend = "cuda"
+dsl = "cuda-oxide"
+depends = []
+src = ["Cargo.toml", "Cargo.lock", "src"]
+device-manifest = "kernels/Cargo.toml"
+ptx-dir = "kernels-ptx"
```
danieldk, line 393: Is it worth exposing this? Maybe we should just use a standard directory?

docs/source/builder/writing-kernels.md
```diff
@@ -352,6 +352,51 @@ are available:
...
+[cuda-oxide](https://github.com/NVlabs/cuda-oxide) `rustc` codegen backend,
+use `dsl = "cuda-oxide"` and point at the device manifest:
+
+```toml
+[kernel.my_kernel]
+backend = "cuda"
+dsl = "cuda-oxide"
+depends = []
+src = ["Cargo.toml", "Cargo.lock", "src"]
+device-manifest = "kernels/Cargo.toml"
```
danieldk, line 392: What does `device-manifest` mean here?

docs/source/builder/writing-kernels.md
```diff
@@ -352,6 +352,51 @@ are available:
...
+PTX). To have the device crate compiled to PTX during the build by the
+[cuda-oxide](https://github.com/NVlabs/cuda-oxide) `rustc` codegen backend,
+use `dsl = "cuda-oxide"` and point at the device manifest:
+
+```toml
+[kernel.my_kernel]
+backend = "cuda"
+dsl = "cuda-oxide"
+depends = []
+src = ["Cargo.toml", "Cargo.lock", "src"]
```
danieldk, line 391: `src`? Don't we always explicitly specify all files? I'm surprised that this works?

examples/kernels/relu-cuda-oxide/kernels/src/main.rs
```diff
@@ -0,0 +1,12 @@
...
+#[kernel]
+pub fn relu_fwd(x: &[f32], mut out: DisjointSlice<f32>) {
+    let idx = thread::index_1d();
+    let i = idx.get();
+    if let Some(out_elem) = out.get_mut(idx) {
+        *out_elem = if x[i] > 0.0 { x[i] } else { 0.0 };
+    }
+}
+
+fn main() {}
```
danieldk, line 12: I'm surprised that this is not built as a lib?

kernel-builder/src/pyproject/templates/tvm_ffi/tvm-ffi-extension.cmake
```diff
@@ -1,6 +1,12 @@
 # Avoid 'lib' prefix for the extension.
 set(CMAKE_SHARED_LIBRARY_PREFIX "")
 
+# rust kernels export the symbols; CMake just needs a stub source for the shared library.
+if(NOT SRC AND RUST_KERNEL_LIBS)
+  file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/_ops_stub.cpp "\n")
+  list(APPEND SRC ${CMAKE_CURRENT_BINARY_DIR}/_ops_stub.cpp)
```
danieldk, line 7: Why is this needed?

kernel-builder/src/pyproject/templates/tvm_ffi/tvm-ffi-extension.cmake
```diff
@@ -15,6 +21,20 @@ if(CXX_HAS_NO_GNU_UNIQUE)
   target_compile_options(${OPS_NAME} PRIVATE $<$<COMPILE_LANGUAGE:GPU_LANGUAGE>:-fno-gnu-unique>)
 endif()
 
+if(RUST_KERNEL_LIBS)
+  add_dependencies(${OPS_NAME} ${RUST_KERNEL_TARGETS})
+  find_package(Threads REQUIRED)
```
danieldk, line 26: This uses pthreads on Linux?

kernel-builder/src/pyproject/templates/tvm_ffi/tvm-ffi-extension.cmake
```diff
@@ -15,6 +21,20 @@ if(CXX_HAS_NO_GNU_UNIQUE)
...
+  target_link_libraries(${OPS_NAME} PRIVATE
+    "$<LINK_LIBRARY:WHOLE_ARCHIVE,${RUST_KERNEL_LIBS}>"
+    Threads::Threads
+    ${CMAKE_DL_LIBS})
+
+  if(GPU_LANG STREQUAL "CUDA")
+    find_package(CUDAToolkit REQUIRED)
+    target_link_libraries(${OPS_NAME} PRIVATE CUDA::cuda_driver)
```
danieldk, line 34: We should never link directly against the CUDA driver library.

https://huggingface.slack.com/archives/C083BS3K2F9/p1759135195942549?thread_ts=1759132908.571519&cid=C083BS3K2F9

(I think there was another discussion about it on Slack as well.)

kernel-builder/src/pyproject/templates/kernel.cmake
```diff
@@ -315,3 +315,108 @@ function(metal_kernel_component SRC_VAR)
...
+        if(NOT KERNEL_PTX_DIR)
+            set(KERNEL_PTX_DIR kernels-ptx)
+        endif()
+
+        get_filename_component(_DEVICE_MANIFEST ${KERNEL_DEVICE_MANIFEST} ABSOLUTE BASE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+        get_filename_component(_PTX_DIR ${KERNEL_PTX_DIR} ABSOLUTE BASE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+        set(_DEVICE_TARGET_DIR ${CMAKE_BINARY_DIR}/cargo/${KERNEL_NAME}-device)
+        set(_DEVICE_ENV)
+
+        if(_KERNEL_ARCHS)
```
danieldk, line 366: Isn't `_KERNEL_ARCHS` set unconditionally above?

kernel-builder/src/pyproject/templates/kernel.cmake
```diff
@@ -315,3 +315,108 @@ function(metal_kernel_component SRC_VAR)
...
+            list(SORT _KERNEL_ARCHS COMPARE NATURAL)
+            # PTX JITs forward, so target the lowest supported arch.
+            list(GET _KERNEL_ARCHS 0 _OXIDE_ARCH)
+            string(REPLACE "+PTX" "" _OXIDE_ARCH "${_OXIDE_ARCH}")
+            string(REPLACE "." "" _OXIDE_ARCH "${_OXIDE_ARCH}")
+            list(APPEND _DEVICE_ENV "CUDA_OXIDE_TARGET=sm_${_OXIDE_ARCH}")
```
danieldk, line 372: Could you explain this a bit more? Usually `sm_.*` generates SASS code, is that not the case here?

kernel-builder/src/pyproject/kernel.rs
```diff
@@ -34,6 +34,9 @@ fn render_kernel_component(
         .join("\n");
 
     match kernel {
+        Kernel::Cpu { .. } | Kernel::Cuda { .. } if kernel.dsl().is_cargo_built() => {
```
danieldk, line 37: In such a context, especially because the function is also called `render_kernel_component_rust`, `is_rust()` sounds nicer?

kernel-builder/src/pyproject/kernel.rs
```diff
@@ -54,6 +57,128 @@ fn render_kernel_component(
...
+            device_manifest.is_some(),
+            "Kernel `{kernel_name}`: `dsl = \"cuda-oxide\"` requires `device-manifest`"
+        );
+    } else {
+        ensure!(
+            device_manifest.is_none() && ptx_dir.is_none(),
+            "Kernel `{kernel_name}`: `device-manifest` and `ptx-dir` require \
+             `dsl = \"cuda-oxide\"`"
+        );
+    }
```
danieldk, line 129: `match` reads nicer

kernel-builder/src/pyproject/mod.rs
```diff
@@ -34,6 +34,20 @@ pub fn create_pyproject_file_set(
...
+        .values()
+        .any(|kernel| kernel.dsl().is_cargo_built())
+        && !matches!(build.framework, Framework::TvmFfi(_))
+    {
+        bail!(
+            "`dsl = \"rust\"` and `dsl = \"cuda-oxide\"` kernels are currently \
+             only supported together with the `[tvm-ffi]` framework"
+        );
+    }
+
```
danieldk, line 50: Seems very specific to put here and clutters what was a really clean function. Maybe we should pass the framework to `render_kernel_component_rust`? Not 100% sure.

kernel-builder/src/pyproject/mod.rs
```diff
@@ -55,7 +69,7 @@ pub fn create_pyproject(
 ) -> Result<()> {
     let kernel_dir = check_or_infer_kernel_dir(kernel_dir)?;
     let target_dir = check_or_infer_target_dir(&kernel_dir, target_dir)?;
-    let build = Build::open(&kernel_dir)?;
+    let build = parse_build(&kernel_dir)?;
```
danieldk, line 72: Can we keep `Build::open`? Reintroducing local parsing functions seems not good.

kernel-builder/src/util.rs
```diff
@@ -4,6 +4,15 @@ use std::path::{Path, PathBuf};
 
 use eyre::{bail, ensure, Context, Result};
 
+use kernels_data::config::{Build, Kernel};
+
+pub(crate) fn parse_build(kernel_dir: impl AsRef<Path>) -> Result<Build> {
+    let kernel_dir = kernel_dir.as_ref();
+    let mut build = Build::open(kernel_dir)?;
+    infer_rust_kernel_lib_names(&mut build, kernel_dir)?;
```
danieldk, line 12: This seems far too specific for configuration that is local to the kernel.

kernels-data/src/config/mod.rs
```diff
@@ -321,7 +321,10 @@ pub enum Kernel {
     Cpu {
         cxx_flags: Option<Vec<String>>,
         depends: Vec<Dependency>,
+        dsl: Option<Dsl>,
+        features: Option<Vec<String>>,
         include: Option<Vec<String>>,
+        lib_name: Option<String>,
```
danieldk, line 327: I don't think we should go here. We go for full declaritivity, convention over configuration. So we should enforce some rule like: the library name must be the same as the kernel name (with dashes underscored or something along those lines).

kernel-builder/src/util.rs
```diff
@@ -37,6 +46,46 @@ pub(crate) fn check_or_infer_target_dir(
...
+            *lib_name = Some(
+                name("lib")
+                    .or_else(|| name("package"))
+                    .ok_or_else(|| {
+                        eyre::eyre!("cannot infer `lib-name` from `{}`", manifest_path.display())
+                    })?
+                    .replace('-', "_"),
+            );
+        }
+    }
```
danieldk, line 85: I don't think we should start parsing Cargo files as part of parsing `build.toml`. Also, I think this invalidates that `Build::open` gives you the same between Python and Rust.

kernels-data/src/config/v5.rs
```diff
@@ -139,6 +139,7 @@ pub struct TorchNoarch {
 pub struct TvmFfi {
     pub include: Option<Vec<String>>,
     pub pyext: Option<Vec<String>>,
+    #[serde(default)]
```
danieldk, line 142: ???

nix-builder/lib/extension/tvm-ffi/arch.nix
```diff
@@ -104,6 +110,7 @@ let
     ];
 
   moduleName = builtins.replaceStrings [ "-" ] [ "_" ] kernelName;
+  kernelToml = (builtins.fromTOML (builtins.readFile (src + "/build.toml"))).kernel or { };
```
danieldk, line 113: I don't think we should start parsing here, we have a file somewhere that parses `build.toml` and also extracts a bunch of commonly-used attributes.

nix-builder/lib/extension/tvm-ffi/arch.nix
```diff
@@ -117,6 +124,79 @@ let
...
+  hasCudaOxideKernels = builtins.any (kernel: kernelDsl kernel == "cuda-oxide") (
+    lib.attrValues kernelToml
+  );
+
+  cudaOxideSrc = pkgs.fetchFromGitHub {
+    owner = "drbh";
+    repo = "cuda-oxide";
+    rev = "bc5f33bb1d556671c35e475439bb821d25caa2cb";
+    hash = "sha256-ZIpS0XFDI9yDSXk33OyQ3mGB5gRhE7tldmmex226tLA=";
+  };
```
danieldk, line 149: Should probably be in the overlay, maybe in an attr-set in there.

nix-builder/lib/extension/tvm-ffi/arch.nix
```diff
@@ -117,6 +124,79 @@ let
...
+      allowBuiltinFetchGit = true;
+    };
+    buildInputs = with pkgs; [
+      libffi
+      libxml2
+      zstd
+      zlib
+    ];
+    doCheck = false;
+  };
```
danieldk, line 176: These should probably be factored out somewhere as an `mk.*` function.

nix-builder/lib/extension/tvm-ffi/arch.nix
```diff
@@ -131,6 +211,15 @@ stdenv.mkDerivation (prevAttrs: {
...
   framework = "tvm-ffi";
 
+  cargoDeps =
+    if hasRustKernels then
+      rustPlatform.importCargoLock {
+        lockFile = src + "/Cargo.lock";
+        allowBuiltinFetchGit = true;
+      }
+    else
+      null;
```
danieldk, line 221: I don't think it's good to have this in the environment for C++ builds (even if set to null).

nix-builder/lib/extension/tvm-ffi/arch.nix
```diff
@@ -166,6 +255,11 @@ stdenv.mkDerivation (prevAttrs: {
     chmod -R u+w .
   '';
 
+  preInstallCheck = lib.optionalString (hasRustKernels && cudaSupport) ''
+    addToSearchPath LD_LIBRARY_PATH "${lib.getOutput "stubs" cudaPackages.cuda_cudart}/lib/stubs"
+    export LD_LIBRARY_PATH
+  '';
```
danieldk, line 261: LD_LIBRARY_PATH inside a derivation typically hides another issue.

nix-builder/lib/extension/tvm-ffi/arch.nix
```diff
@@ -175,6 +269,18 @@ stdenv.mkDerivation (prevAttrs: {
     remove-bytecode-hook
     torch-ops-check
   ]
+  ++ lib.optionals hasRustKernels (
+    [ rustPlatform.cargoSetupHook ]
+    ++ (
+      if rust-bin != null && builtins.pathExists (src + "/rust-toolchain.toml") then
+        [ (rust-bin.fromRustupToolchainFile (src + "/rust-toolchain.toml")) ]
```
danieldk, line 276: Do we want to allow custom toolchains? Also, it seems that it can differ from cuda-oxide?

nix-builder/lib/extension/tvm-ffi/arch.nix
danieldk: Reading the rest of the file, maybe we should create a separate `arch-rust.nix`, at least for now?

PR #726, opened 2026-07-23
fix incompatible variants problem on mac.

### LINKED ISSUE #705: Metal kernels built for macOS 26+ are loaded on macOS &lt; 26 (should be treated as incompatible variant)
`kernels` will happily fetch and attempt to load a Metal kernel built for macOS 26+ / Metal 4+ when running on an older macOS (e.g. macOS 15), which then fails at runtime with a variety of confusing errors.

### Repro

```python
import torch

k = get_kernel("kernels-community/relu", version=1)
x = torch.randn(1024, device="mps")
print(k.relu(x))
```

On macOS 15 this yields errors such as:

```
RuntimeError: Function relu_forward_kernel_float is using language version 4.0 which is incompatible with this OS.
```
```
RuntimeError: Failed to get Metal function: steel_attention_float16_bq32_bk32_bd64_wm4_wn1_maskbool_ Error: error: incompatible module, module was built for 'air64_v28-apple-macosx26.0.0' which is not compatible with the target (air64-apple-macosx15.7.0)
```
```
RuntimeError: Kernel not found: bnb_dequantize_blockwise_half_bs_64_qt_2
```

### Background

`kernel-builder` currently targets macOS 26+ / Metal 4+ (see the hardcoded value in [`compile-metal.cmake`](https://github.com/huggingface/kernels/blob/219d1d03e321757bba5a3cad15e8f1720d8c9cc9/kernel-builder/src/pyproject/templates/torch/metal/compile-metal.cmake#L21)), but the [kernel-requirements doc](https://github.com/huggingface/kernels/blob/main/docs/source/kernel-requirements.md#macos) still says macOS 15.0 / Metal 3.0, which is out of date.

### Proposed behavior

Discussed with @danieldk / @sayakpaul:

- Start encoding the macOS/Metal version in the Metal build variant (planned for the next macOS release).
- Until then, only download/resolve a Metal kernel when the user's macOS version is at least 26; otherwise mark it as an **incompatible variant** rather than downloading it.
- Note the intentional divergence between `get_kernel` (which errors when no variant resolves) and `kernelize` (which should *not* fail models — e.g. from transformers — just because there is no matching kernel).
- Update the kernel-requirements docs to reflect the actual macOS/Metal requirement.

---
Requested by Sayak Paul - [Slack thread](https://huggingface.slack.com/archives/C083BS3K2F9/p1781302291265109?thread_ts=1781302291.265109&cid=C083BS3K2F9) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-07-09T04-10-11-914Z_a4678c7e-b22d-4e96-8720-d21705f78ad7.jsonl)

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/705

### DIFF
--- a/docs/source/kernel-requirements.md
+++ b/docs/source/kernel-requirements.md
@@ -277,8 +277,8 @@ These requirements can be checked with the ABI checker (see below).
 
 - Use [ABI3/Limited API](https://docs.python.org/3/c-api/stable.html#stable-application-binary-interface)
   for compatibility with Python 3.9 and later.
-- macOS deployment target 15.0.
-- Metal 3.0 (`-std=metal3.0`).
+- macOS deployment target 26.0.
+- Metal 4.0 (`-std=metal4.0`).
 
 The ABI3 requirement can be checked with the ABI checker (see below).
 
--- a/kernels/src/kernels/variants.py
+++ b/kernels/src/kernels/variants.py
@@ -18,11 +18,18 @@ from kernels.backends import (
     CUDA,
     XPU,
     Backend,
+    Metal,
     ROCm,
     _select_backend,
     parse_backend,
 )
 from kernels.compat import has_torch, has_tvm_ffi
+
+# Metal kernels are currently built with `-std=metal4.0`, which requires
+# macOS 26 or later. Until the macOS/Metal version is encoded in the build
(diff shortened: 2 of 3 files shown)


drbh (approved):
lgtm!

PR #731, opened 2026-07-28
[tools] add utilities for deterministic port to `kernels`

sayakpaul:
Related to https://github.com/huggingface/kernels/issues/729. 

I know we have been talking about having a new DSL. I didn't take that route in this PR. Instead, I decided to add atomic components that can be used in isolation as well as in the DSL. These operations include:

* Determining the latest release tag and then fetching it for a given kernel repository on GitHub.
* Relativizing imports as specified in https://huggingface.co/docs/kernels/main/en/kernel-requirements
* Fixing PyTorch op registration as specified in https://huggingface.co/docs/kernels/main/en/kernel-requirements through ASTs

Have added tests for each atomic operation, too. But didn't intentionally add them to the CI because we still haven't decided on the direction this whole utility would take.

> [!TIP]
> I think for the initial review, it's more important to decide if we want to have this kind of atomic utilities and agree on their signatures rather than their actual implementation details. Best to maybe start with the README.md added in this PR to get a sense of how I have decided to structure it.

TODOs:

- [ ] Documentation. Planning on having either a short section or an entire page dedicated to "porting existing kernel projects to be `kernel-builder` compliant".
- [ ] Wire this up in AGENTS.md of `kernels-community` so that agents actually know which tools to use for deriving deterministic outputs rather than hits and misses.


drbh:
thanks for putting this together! I think the AST utilities here are useful, especially the import relativization and op registration work. however, after looking through this more, I don't think standalone tools driven by an agent are exactly the design we want for the larger porting workflow.

the long term goal is to have a maintainable definition of how an upstream kernel is converted into the kernel-builder format. ideally a port should look something like:

```text
pinned upstream checkout + 
checked-in conversion steps + 
checked-in overlays = 
kernel-builder compatible tree
```

when upstream changes, we should be able to update the pin and rerun the same conversion from a fresh checkout. if upstream changed something the conversion depends on, it should fail at that specific step and require us to update the conversion. this gives us a reviewable history of how the port changes along with upstream.

with the current approach, each individual tool can be deterministic and safe to rerun, but the port as a whole is still stateful. the agent decides what to copy, move, delete, which tools to run, the order to run them in, and how to resolve anything the tools cannot handle. those decisions end up represented in the mutated source tree rather than in something we can replay later.

I think we should start with the higher level pipeline/DSL contract and make transformations like these operations within it. a lot of the AST work in this PR could still be reused there. the important difference is that the main interface should take pinned source inputs, run a checked-in conversion against a fresh tree, and produce the same output for the same inputs.

tldr; I think this PR contains useful primitives, but the thing we need to maintain is the conversion recipe, not the converted checkout. before settling on the standalone tool signatures, I think we should agree on that higher level execution model first.

sayakpaul:
Closing for now.

PR #734, opened 2026-07-29
Crescent Island (cri) support for the XPU/SYCL build

jiqing-feng:
Adds Intel Crescent Island (`cri`) as an AOT target for SYCL kernels.

## Why it isn't just one more device in the list

`cri` needs a newer toolchain: `ocloc` 25.35 doesn't know the device, and
sycl-tla 0.9.1 has no cri kernels. Both are bumped (ocloc → 26.22.38646.4,
sycl-tla → 0.9.2), and AOT for cri is gated on DPC++ 2026.0.

The bigger problem is that a kernel compiled with `__SYCL_TARGET_INTEL_GPU_CRI__`
is an Xe35 unit throughout, and its device code must not be lowered for pvc or
bmg. Since SYCL merges every TU that shares a target triple into one image, and
`ocloc` compiles that image once per device in its `-device` list, such a
translation unit can't share one AOT image across devices. The build now detects
that flag and picks the layout:

- without it — a single `spir64_gen` image serves `pvc,bmg-g21,cri`, which keeps
  the number of device compilations per TU down;
- with it — the generic image is restricted to `cri`, and pvc/bmg get separate
  `intel_gpu_*` alias targets.

Either way the `.so` also carries a `spir64` JIT fallback, so a GPU that isn't
in the AOT list still runs.

## Why the flag handling was rewritten

The SYCL flags used to be two hardcoded strings that dependencies patched with
`string(REPLACE ...)`. That stops working once the target list is dynamic, and
it hid three failure modes that all produce a binary that builds fine and then
misbehaves:

- backend options must be repeated per target, and CMake deduplicates the
  identical `-options` strings, silently dropping every target but the first —
  `SHELL:` prevents this;
- `-spirv-ext` only reaches the image whose triple matches the exact
  `-fsycl-targets` alias, so it has to be emitted once per target;
- the SPIR-V translator *replaces* the extension list instead of merging, so the
  driver defaults are now probed with `icpx -###` and sycl-tla's extras appended.
  A failed probe is a hard error rather than a silent fallback.

Flags are now built from a few variables through a `xpu_compose_sycl_flags()`
macro, so a dependency changes a variable and recomposes.

## Validation

Built `flash-attn2` and `megablocks` from `kernels-community` against this
branch (`torch213-cxx11-xpu20260-x86_64-linux`) and checked numerics on both
BMG hardware and the CRI TBX simulator. Each case runs only the kernel on XPU;
data generation, the fp32 reference, MXFP4/MXFP8 dequantization and error
statistics stay on CPU.

- flash-attn2 — 8/8 pass (fix / varlen × fp16 / bf16 × causal / non-causal),
  max abs error 5.4e-04 (fp16) / 4.0e-03 (bf16).
- megablocks grouped GEMM — 5/5 pass (bf16 ±bias, fp16, mxfp4, mxfp8), max
  relative error 2.6e-03.

Both AOT layouts were also checked in the shipped binaries: flash-attn2, which
has no CRI translation unit, carries one `spir64_gen` image covering all three
devices, while megablocks carries per-device images. On the simulator the
runtime loads the AOT image rather than falling back to JIT.


jiqing-feng:
Hi @danieldk . Would you please review this PR? Thanks!

sayakpaul:
Daniel is away this week. Please expect some delay.

danieldk:
Thanks for the PR, I have it on my list for tomorrow!

danieldk (approved):
Looks good, thanks! Merging when CI is happy.

PR #735, opened 2026-07-29
add comment on trust_remote_code=True

sayakpaul:
It is a bad experience for the users when most of the kernels on hf.co/kernels need `trust_remote_code=True` in the `get_kernel()` call. This PR adds a comment about that in our card template.

### DIFF
--- a/examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
+++ b/examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
@@ -12,8 +12,10 @@ This is the repository card of {{ repo_id }} that has been pushed on the Hub. It
 # make sure `kernels` is installed: `pip install -U kernels`
 from kernels import get_kernel
 
-# If the org / user isn't a trusted publisher, then you should check out the kernel
-# source, and need to then pass `trust_remote_code=True`.
+# If the org / user isn't a trusted publisher, pass `trust_remote_code=True` to the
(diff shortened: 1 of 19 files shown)


examples/kernels/cutlass-gemm-tvm-ffi/CARD.md
```diff
@@ -12,6 +12,8 @@ This is the repository card of {{ repo_id }} that has been pushed on the Hub. It
 # make sure `kernels` is installed: `pip install -U kernels`
 from kernels import get_kernel
 
+# If the org / user isn't a trusted publisher, then you should check out the kernel
+# source, and need to then pass `trust_remote_code=True`.
```
danieldk: What should they check the kernel source for? Maybe something like

```suggestion
# If the org / user isn't a trusted publisher, pass `trust_remote_code=True` to the
# `get_kernel` call. You can find whether this kernel is from a trusted publisher
# by going to the kernel's Hub page and finding the "Trusted publisher" status at
# the top of the page.
```

?

Though it would be nicer if there was a page with trusted publishers, then we could say

```suggestion
# If the org / user isn't a trusted publisher, pass `trust_remote_code=True` to the
# `get_kernel` call. Go to huggingface.co/trusted-publishers for a list of trusted
# publishers.
```

sayakpaul:
Once we have a better way to query the trusted publishers of kernels, I will do another modification.

danieldk (approved):

PR #736, opened 2026-07-29
kernel-builder: stop generating the compat module

danieldk:
The compat module was only needed for ancient versions of `kernels` and tests where `PYTHONPATH` is set manually (which we do not support anymore). The compat module sometimes causes double imports, failing when ops get defined twice and polluting the Python namespace.

So, as discussed, this change removes generation of the compat module.

In line with that the tests in the examples are all updated to use `get_kernel`. Also, all these cheap tests are now marked with `kernels_ci`, so that they run in CI tests.

Also fix two other issues encountered while running the updated tests:

- The `ci-test` command exposed in a kernel flake did not set `LOCAL_KERNELS` (to override `get_kernel` to the local path).
- Tests for tvm-ffi kernels gave warnings because the `kernels_ci` marker was not defined in the generated `pyproject.toml`.


nix-builder/lib/extension/tvm-ffi/arch.nix
```diff
@@ -127,15 +127,18 @@ stdenv.mkDerivation (prevAttrs: {
...
-    kernel-builder create-pyproject \
-      --unique-id ${rev} ${provenanceFlags} .
+  # We run kernel-builder here rather than patchPhase or preConfigure,
+  # so that external users of `src` get the source tree with the files
+  # generated by kernel-builder.
+  src = pkgs.runCommand "source" { } ''
+    mkdir -p $out
+    cp -r --no-preserve=mode ${src}/* $out/
+    ${pkgs.kernel-builder}/bin/kernel-builder create-pyproject \
+      --unique-id ${rev} ${provenanceFlags} $out
```
sayakpaul, line 141: Makes sense. Similar to the `arch.nix` of PyTorch. But, out of curiosity, why did we have it like this in TVM?

nix-builder/lib/checks.nix
```diff
@@ -29,7 +29,7 @@ let
     ];
   };
 in
-assert lib.assertMsg (builtins.all (buildSet: buildSet.torch.version == "2.11.0") kernelBuildSets)
+assert lib.assertMsg (builtins.all (buildSet: buildSet.torch.version == "2.12.0") kernelBuildSets)
```
sayakpaul, line 32: Always pinned to minimum, okay.

nix-builder/tests/Dockerfile.test-kernel
```diff
@@ -14,9 +14,10 @@ ENV DEBIAN_FRONTEND=noninteractive \
     NVIDIA_VISIBLE_DEVICES=all \
     NVIDIA_DRIVER_CAPABILITIES=compute,utility
 
-# Install uv.
+# Install uv and Rust toolchain (Rust is needed to build kernels-data from source).
 RUN dnf install -y --allowerasing curl && \
     curl -LsSf https://astral.sh/uv/install.sh | sh && \
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y && \
```
sayakpaul, line 20: Faced a similar error just today. Good one.

nix-builder/tests/run-tests.sh
```diff
@@ -13,20 +13,20 @@ SILU_MUL_PATH=$(echo silu-and-mul-kernel/torch*)
 RELU_CPU_PATH=$(echo relu-kernel-cpu/torch*)
 CPP20_SYMBOLS_PATH=$(echo cpp20-symbols-kernel/torch*)
 
-PYTHONPATH="$EXTRA_DATA_PATH:$RELU_PATH:$RELU_TORCH_STABLE_ABI_PATH:$RELU_TVM_FFI_PATH:$CUTLASS_PATH:$CUTLASS_TVM_FFI_PATH" \
+LOCAL_KERNELS="kernels-test/extra-data=${EXTRA_DATA_PATH}:kernels-test/relu=${RELU_PATH}:kernels-test/relu-torch-stable-abi=${RELU_TORCH_STABLE_ABI_PATH}:kernels-test/relu-tvm-ffi=${RELU_TVM_FFI_PATH}:kernels-test/cutlass-gemm=${CUTLASS_PATH}:kernels-test/cutlass-gemm-tvm-ffi=${CUTLASS_TVM_FFI_PATH}" \
```
sayakpaul, line 16: Should these changes translate to any guidance in the docs? Or are these fairly CI-specific?

nix-builder/tests/Dockerfile.test-kernel
```diff
@@ -60,6 +61,14 @@ RUN CUDA_MAJOR_MINOR=$(echo ${CUDA_VERSION} | cut -d'.' -f1,2) && \
 # Add additional dependencies.
 RUN uv add "apache-tvm-ffi~=0.1.9" numpy pytest
 
+# Copy kernels-data and kernels source to install from source.
+COPY kernels-data ./kernels-data-src
+COPY kernels ./kernels-src
+
+# Install kernels-data (Rust/maturin package) and kernels from source.
+RUN uv add ./kernels-data-src/bindings/python
```
sayakpaul, line 69: I didn't know that if one has `build-system.requires` configured, it can actually perform the build through `uv add ...`.

sayakpaul (approved):
Just some out-of-curiosity questions. Nothing merge-blocking. Thanks!

nix-builder/lib/extension/tvm-ffi/arch.nix
danieldk: `arch.nix` for Torch was also originally like this. I think we forgot to update it for tvm-ffi.

nix-builder/tests/run-tests.sh
danieldk: We weren't following our own docs 😅:

https://huggingface.co/docs/kernels/main/en/builder/writing-kernels#use-getkernel-in-tests
https://huggingface.co/docs/kernels/main/en/builder/local-dev#testing-kernel-builds-before-publishing

PR #738, opened 2026-07-30
use kernel-builder in actions and jobs

### LINKED ISSUE #717: Make kernel-builder available as a CLI in the job container (use build-and-upload instead of nix run)
The GitHub Actions builder docs currently instruct users to invoke the `kernel-builder` via `nix run ...`. It would be better to expose `kernel-builder` as a proper CLI command so the docs (and the job) can use `kernel-builder build-and-upload` directly. Steps: update the Dockerfile to place the kernel-builder binary on PATH, rebuild/push the container, update the job to use it, and update the docs.

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/717


drbh (approved):
thanks for updating, lgtm!

PR #740, opened 2026-07-30
kernel-builder: fix collision of static variables between kernels

danieldk:
gcc makes template static data members and static local variables in inline functions globally unique. This is, however, a problem for kernels, since two kernels or kernel versions could use a static variable with the same name, resulting in inconsistent state.

Note that this is not resolved by `RTLD_LOCAL` being used when `dlopen`'ing Python extensions. These globally unique symbols are excepted in hiding by `RTLD_LOCAL`.

We resolve this by passing `-fno-gnu-unique` to compilers that support it (compilers that don't support it, most likely do not make these symbols globally unique).

Also add tests that would catch such unwanted sharing in the future.

### DIFF
--- a/.github/workflows/build_kernel_cpu.yaml
+++ b/.github/workflows/build_kernel_cpu.yaml
@@ -1,0 +1,41 @@
+name: "Build and test kernel (CPU)"
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened] # trigger on PRs
+    paths-ignore:
+      - "docs/**"
+      - "*.md"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build kernel
+    runs-on:
+      group: aws-highmemory-32-plus-nix
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25
+        with:
+          extra-conf: |
+            max-jobs = 4
+            cores = 12
+            sandbox-fallback = false
+      - uses: cachix/cachix-action@1eb2ef646ac0255473d23a5907ad7b04ce94065c # v17
+        with:
+          name: huggingface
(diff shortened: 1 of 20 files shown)


.github/workflows/build_kernel_cpu.yaml
```diff
@@ -0,0 +1,41 @@
+name: "Build and test kernel (CPU)"
```
sayakpaul, line 1: Could I get an explanation on why it could not be merged into our existing workflows?

.github/workflows/build_kernel_cpu.yaml
```diff
@@ -0,0 +1,41 @@
...
+          name: huggingface
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
+        env:
+          USER: runner
+      - name: Nix info
+        run: nix-shell -p nix-info --run "nix-info -m"
+      # CPU kernels don't require special hardware, so we build and run
+      # their CI tests (e.g. the symbol-conflicts pytest suite) in one step.
+      - name: Build and test all CPU example kernels
+        run: nix build -L ./examples/kernels#ci-build-cpu
```
sayakpaul, line 41: Should use `kernel-builder ...` here?

kernel-builder/src/pyproject/templates/torch/torch-extension.cmake
```diff
@@ -14,6 +14,15 @@ define_gpu_extension_target(
   USE_SABI 3
   WITH_SOABI)
 
+# Avoid that definitions of template static data members and static local
```
sayakpaul, line 17: We should probably take an example kernel where this was surfaced?

sayakpaul (approved):
Thanks a lot!

.github/workflows/build_kernel_cpu.yaml
danieldk: These are Nix outputs that build multiple kernels. Not something kernel-builder can build.

.github/workflows/build_kernel_cpu.yaml
danieldk: We have separate workflows for CUDA, ROCm, and XPU. CPU was just missing. There are some CPU-only things in the CUDA tests, but I also want to move them into this workflow in the future. The CPU workflow and CI target is much nicer, because we can run tests during the build (which we can't for CUDA, due to the sandbox).

kernel-builder/src/pyproject/templates/torch/torch-extension.cmake
danieldk: Not sure. I think the example kernels that are included in this PR describe the issue in a nice and minimalist way?

PR #741, opened 2026-08-01
Add CUTLASS 4.4.2 dependency and fix 4.5.2 source hash

LiangSu8899:
This does two narrowly scoped builder dependency updates needed by FlashRT FMHA packaging:\n\n- adds `cutlass_4_4` backed by CUTLASS v4.4.2; example-77 FMHA sources use the pre-4.5 `SM100_MMA_F8F6F4_SS` template interface and do not compile against 4.5.2;\n- updates the stale CUTLASS 4.5.2 fixed-output hash reported by Nix.\n\n`cargo test -p kernels-data` passes (29/29). The FlashRT package using `cutlass_4_4` now builds successfully through HF Jobs.

### DIFF
--- a/kernels-data/src/config/deps.rs
+++ b/kernels-data/src/config/deps.rs
@@ -25,6 +25,8 @@ pub enum Dependency {
     Cutlass3_9,
     #[serde(rename = "cutlass_4_0")]
     Cutlass4_0,
+    #[serde(rename = "cutlass_4_4")]
+    Cutlass4_4,
     #[serde(rename = "cutlass_4_5")]
     Cutlass4_5,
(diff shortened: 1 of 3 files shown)


danieldk (approved):

danieldk:
Thanks a lot! I wonder if the 4.5.2 tag was moved/updated.

LiangSu8899:
> Thanks a lot! I wonder if the 4.5.2 tag was moved/updated.

Thanks for the quick response! 
Could be, not sure whether the tag moved or the GitHub archive changed, but the new hash matches the current source and all builder tests pass.

PR #745, opened 2026-08-04
Kernel loading untangling

danieldk:
Kernel dependencies are somewhat hard to implement because we have a bunch not nicely-orthogonal functions. 

This is the first in a series of PR to clean up kernel loading and `utils.py`. I'm doing this over multiple PRs to keep it reviewable. In this PR:

* Remove unnecessary nearly-identical `snapshot_download` calls.
* Remove kernel validation based on lockfile hashes. This is superseded by the metadata digest + signing. Plus doing this on every load has quite some overhead.
* Remove offline argument from `install_kernel_all_variants`. This is basically a function only meant to be used online.
* Remove `variant_locks` argument from `install_kernel` (not needed anymore, since we removed the validation).
* Remove the `revision` argument from `load_kernel`. There is no real use case for it and it complicates the logic and potential refactoring.
* Move all bits related to locking to the new `locking` module.

### DIFF
--- a/kernels/src/kernels/cli/__init__.py
+++ b/kernels/src/kernels/cli/__init__.py
@@ -153,7 +153,6 @@ def download_kernels(args):
             install_kernel_all_variants(
                 kernel_lock.repo_id,
                 revision=kernel_lock.sha,
-                variant_locks=kernel_lock.variants,
             )
         else:
             try:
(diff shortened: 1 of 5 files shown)


drbh (approved):
lgtm!

PR #746, opened 2026-08-04
Untangle kernel loading, part2

danieldk:
* Move hub-related functionality to the new `hf_hub` module.
* Use `install_kernel` in `load_kernel` (goal: make `install_kernel` the default entry point for getting kernel paths).
* Remove `local_files_only` arg from `get_locked_kernel`, since this case is served by `load_kernel`.
* Make `install_kernel` more pure by moving `HF_HUB_OFFLINE` handling out of the function. The behavior should be based on the arguments as much as possible. Implicit behaviors make it hard to reason about functions.


drbh (approved):
lgtm!

PR #747, opened 2026-08-05
Untangle kernel loading, part 3

danieldk:
- Remove unused `KNOWN_BACKENDS` variable.
- Move `get_kernel_variants` to the `variants` module.
- Move import-related functions to `importer` (note that we cannot have `import` because it is a keyboard).
- Move `validate_variant_dependencies` to `deps`.
- Move install functions to `install`.
- Rename `utils` module to `load`.

### DIFF
--- a/kernels/src/kernels/cli/__init__.py
+++ b/kernels/src/kernels/cli/__init__.py
@@ -8,11 +8,11 @@ from kernels.cli.info import print_kernel_info
 from kernels.cli.verify_signature import verify_signature
 from kernels.cli.versions import print_kernel_versions
 from kernels.compat import tomllib
-from kernels.locking import KernelLock, get_kernel_locks
-from kernels.utils import (
+from kernels.install import (
     install_kernel,
     install_kernel_all_variants,
 )
+from kernels.locking import KernelLock, get_kernel_locks
 
 
 def main():
--- a/kernels/src/kernels/importer.py
+++ b/kernels/src/kernels/importer.py
@@ -1,0 +1,122 @@
+import importlib
+import sys
+import warnings
+from dataclasses import dataclass
+from pathlib import Path
+from types import ModuleType
+
+from kernels_data import Metadata
+
+from kernels.hf_hub import RepoInfo
+
+
+@dataclass(frozen=True)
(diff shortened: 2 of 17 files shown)


kernels/src/kernels/cli/__init__.py
```diff
@@ -8,11 +8,11 @@
 from kernels.cli.verify_signature import verify_signature
 from kernels.cli.versions import print_kernel_versions
 from kernels.compat import tomllib
-from kernels.locking import KernelLock, get_kernel_locks
-from kernels.utils import (
+from kernels.install import (
     install_kernel,
     install_kernel_all_variants,
 )
+from kernels.locking import KernelLock, get_kernel_locks
```
sayakpaul, line 15: Love the separation!

drbh (approved):
lgtm!

kernels/src/kernels/importer.py
```diff
@@ -0,0 +1,122 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+from types import ModuleType
+
+from kernels_data import Metadata
+
+from kernels.hf_hub import RepoInfo
+
+
+@dataclass(frozen=True)
```
sayakpaul, line 13: What is the reasoning behind keeping something that is related to loading in "importer"?

kernels/src/kernels/importer.py
```diff
@@ -0,0 +1,122 @@
...
+
+    metadata: Metadata
+    module: ModuleType
+    repo_info: RepoInfo | None
+
+
+_loaded_kernels: dict[Path, LoadedKernel] = {}
+
+
+def get_loaded_kernels() -> list[LoadedKernel]:
```
sayakpaul, line 45: Same as above.

sayakpaul (approved):
Looks really nice. Just some small questions.

kernels/src/kernels/importer.py
danieldk: Because it's used by the import functionality. I think the lines between loading and importing is a bit blurry. But I think importing is the last part of loading, so this could have been called `ImportedKernel` as well.

PR #749, opened 2026-08-07
Add `kernel-depends` option to `general` options in `build.toml`

danieldk:
This change adds kernel-kernel dependency support to `kernels-data`, so that this becomes an option in `build.toml` and gets written to `metadata.json`.  The other bits (e.g. in `kernels`) will be wired up in future PRs.


drbh (approved):
nice!

PR #752, opened 2026-08-10
Request IEEE fp32 division for SYCL kernels

jiqing-feng:
## Summary

Adds `-foffload-fp32-prec-div` and `-foffload-fp32-prec-sqrt` to the SYCL build
flags, matching how PyTorch builds its own XPU kernels in `torch-xpu-ops`.

Without them the compiler defaults to a fast floating-point model in which fp32
division is allowed several ULP of error and `a / b` is lowered to `a * rcp(b)`.
Kernels built this way return results that differ from the eager PyTorch ones in
the last bit, which breaks kernel tests that compare against PyTorch exactly.

These flags affect the offload backend and only take effect when passed to both
the device compile and the device link step, so the per-kernel `sycl-flags` in
`build.toml` cannot substitute for them. Applied to the `torch` and `tvm_ffi`
preambles alike.

## Validation

Verified with `kernels-community/activation` on Intel Arc Pro B60 with
`torch 2.13.0+xpu` and oneAPI 2026.0. Its `silu_and_mul` and `mul_and_silu`
operators go from 42 failures to a fully passing suite (396 cases), with no
change in measured throughput.

### DIFF
--- a/kernel-builder/src/pyproject/templates/torch/preamble.cmake
+++ b/kernel-builder/src/pyproject/templates/torch/preamble.cmake
@@ -205,8 +205,10 @@ elseif(GPU_LANG STREQUAL "SYCL")
   endif()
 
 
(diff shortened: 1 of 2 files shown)


drbh (approved):
lgtm!

PR #754, opened 2026-08-10
feat: add an example on using autotuning.

### LINKED ISSUE #733: Document/example: shipping Triton autotune configs with a kernel
Autotuning typically results in kernel parameters that can be re-used in future JITing, so that the kernel is used with an optimal configuration. In most cases, you do not need anything extra — we have shipped kernels in the past with autotuning files, such as vLLM MoE. In that case, the kernel gets downloaded, Triton looks if there is an autotune configuration available that matches the current GPU and if so, uses that autotune config, if not, uses a generic set of parameters. We shipped the autotune files with the kernel (they are just JSON files).

Proposal: we should have a Triton kernel in the examples with a script that autotunes and shows how to use the autotunes in the kernel.

---
Requested by Sayak Paul - [Slack thread](https://huggingface.slack.com/archives/D0BG02C6HUL/p1785242823225269?thread_ts=1785242823.225269&cid=D0BG02C6HUL) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-07-28T12-47-04-336Z_250ad79e-6754-4881-a33d-72a0264bebfc.jsonl)

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/733


docs/source/builder/writing-kernels.md
```diff
@@ -37,7 +37,9 @@ format of the `build.toml` file, and some additional Python glue that
 `kernel-builder` provides. We will use a [simple ReLU kernel](https://github.com/huggingface/kernels/tree/main/examples/kernels/relu)
 as the running example. After reading this page, you may also want to have
 a look at the more realistic [ReLU kernel with backprop and `torch.compile`](https://github.com/huggingface/kernels/tree/main/examples/kernels/relu-backprop-compile)
-support.
+support. For Triton kernels, see
```
danieldk: I think this is too specific/advanced to have in a section that is a very basic introduction to writing kernels.

docs/source/_toctree.yml
```diff
@@ -47,6 +47,8 @@
 - sections:
     - local: builder/writing-kernels
       title: Write kernels
+    - local: builder/triton-autotune
+      title: Ship Triton autotune configurations
```
danieldk: I think this is too specific to have as the second item in kernel writing, even before basics like setting up a development environment.

I think it makes sense to add a new section called something like: *Learn about best practices* or *Learn about framework-specific techniques* (something along those lines) and make shipping Triton autotune files a subsection. There may be other CuTe/Helion/... subsections in the future.

docs/source/builder/triton-autotune.md
```diff
@@ -0,0 +1,176 @@
+# Ship Triton autotune configurations
+
+Triton kernels typically have parameters — tile sizes, number of warps,
+number of pipeline stages — whose optimal values depend on the GPU and the
```
danieldk:
```suggestion
Triton kernels typically have parameters, such as tile sizes, number of warps and 
number of pipeline stages, whose optimal values depend on the GPU and the
```

I'm very hesitant about em-dashes these days. Since native speakers told me that the Oxford comma is just a matter of taste, leaving it out makes comma-deliniation work.

docs/source/builder/triton-autotune.md
```diff
@@ -0,0 +1,176 @@
...
+it at runtime (e.g. with the
+[`@triton.autotune`](https://triton-lang.org/main/python-api/generated/triton.autotune.html)
+decorator) re-benchmarks every candidate configuration in each new process.
+
+
+However, one can run the tuner for the GPU models they like and store the
+best found configurations as files for using them later. This effectively
+reduces the potentially costly tuning time.
+
+`kernels` support this by packaging these configurations as JSON files
```
danieldk:
```suggestion
`kernels-builder` support this by packaging these configurations as JSON files
```

Since there is no special handling for this in `kernels` itself.

docs/source/builder/triton-autotune.md
```diff
@@ -0,0 +1,176 @@
...
+the [`gemm-triton-autotune`](https://github.com/huggingface/kernels/tree/main/examples/kernels/gemm-triton-autotune)
+example kernel, a Triton GEMM published as
+[`kernels-test/gemm-triton-autotune`](https://huggingface.co/kernels-test/gemm-triton-autotune).
+
+## Shipping data files with a kernel
+
+Configuration files are plain JSON files inside the kernel's Python package,
+in `torch-ext/<package_name>/configs/`. By default, only `py` and `pyi`
+files are picked up from the package directory, so add `json` to the
+[`pyext` option](writing-kernels.md#torch-noarch) in `build.toml`:
```
danieldk:
```suggestion
Since autotune files are plain JSON files, we can store them anywhere inside the
kernels main Python sources in `torch-ext/<kernel_name>`. For this example,
we will use  `torch-ext/<kernel_name>/configs/`. By default, only `py` and `pyi`
files are picked up from the kenel's Python source directory, so add `json` to the
[`pyext` option](writing-kernels.md#torch-noarch) in `build.toml`:
```

docs/source/builder/triton-autotune.md
```diff
@@ -0,0 +1,176 @@
...
+resources of a smaller GPU (e.g. shared memory) is therefore harmless to
+ship.
+
+## Looking up configurations at runtime
+
+At kernel launch, the kernel checks whether a configuration file exists for
+the current device and shape. If it does, the configuration with the nearest
+tuned `M` is used; otherwise the kernel falls back to a conservative default
+and logs a warning. The lookup is cached, so the file is read at most once
+per process:
```
danieldk, line 98: I don't think this is true, since by default `maxsize=128`, so given enough N, K combinations, it will reload. Since the size will be bound by the number of files anyway, I think it's better to use `functools.cache`, which is equivalent to `lru_cache(maxsize=None)`.

docs/source/builder/triton-autotune.md
```diff
@@ -0,0 +1,176 @@
...
+
+```python
+@functools.lru_cache
```
danieldk:
```suggestion
@functools.cache
```

docs/source/builder/triton-autotune.md
```diff
@@ -0,0 +1,176 @@
...
+        path.name,
+        N,
+        K,
+    )
+    return None
+
+
+def get_config(M: int, N: int, K: int) -> Dict[str, int]:
+    tuned = _load_tuned_configs(N, K)
+    if tuned:
```
danieldk:
```suggestion
    if tuned is not None:
```

(truthy values are meh)

examples/kernels/gemm-triton-autotune/torch-ext/gemm_triton_autotune/gemm.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+
+@torch.library.custom_op(add_op_namespace_prefix("gemm"), mutates_args={"out"})
+def _gemm(out: torch.Tensor, a: torch.Tensor, b: torch.Tensor) -> None:
+    M, K = a.shape
+    N = b.shape[1]
+    if M == 0 or N == 0:
+        return
+    launch_gemm_kernel(a, b, out, get_config(M, N, K))
```
danieldk, line 98: Question: how does this interact with `torch.compile`? Since there is a conditional in the main code path (dict lookup).

examples/kernels/flake.nix
```diff
@@ -179,6 +179,11 @@
           path = ./relu-triton;
           drv = sys: out: out.packages.${sys}.redistributable.torch-cuda;
         }
+        {
+          name = "gemm-triton-autotune-kernel";
```
danieldk, line 183: We have an existing test that tests usage of `pyext`. Maybe this new example can replace it? (cuts down test build time)

danieldk (commented):
Really cool! Left a bunch of comments.

docs/source/builder/triton-autotune.md
sayakpaul: Interesting. Some of my native English speaking friends suggested the use of em dashes in academic writing. But I will roll with the suggestion.

docs/source/builder/triton-autotune.md
sayakpaul:
```suggestion
`kernel-builder` support this by packaging these configurations as JSON files
```

docs/source/builder/triton-autotune.md
sayakpaul: Nice I like it. It adds more specificity.

docs/source/builder/triton-autotune.md
sayakpaul: Ah yeah! I follow it in the other libs too (`diffusers`, `peft`, etc.).

docs/source/builder/writing-kernels.md
sayakpaul: Okay, let's not overwhelm our users. Removed.

examples/kernels/flake.nix
sayakpaul: Follow-up?

examples/kernels/gemm-triton-autotune/torch-ext/gemm_triton_autotune/gemm.py
sayakpaul: Compiles cleanly. Just verified:

```py
import torch
import kernels

gemm_kernel = kernels.get_kernel("kernels-test/gemm-triton-autotune", version=1)


def fn(a, b):
    return gemm_kernel.gemm(torch.nn.functional.silu(a), b) * 2

compiled = torch.compile(fn, fullgraph=True, dynamic=True)
b = torch.randn(4096, 4096, device="cuda", dtype=torch.float16)

with torch._dynamo.config.patch(error_on_recompile=True):
    torch.compiler.reset()
    for M in (3, 33, 333, 3333):
        compiled(torch.randn(M, 4096, device="cuda", dtype=torch.float16), b)
    torch.compiler.reset()
```

sayakpaul:
@danieldk thanks for the comments. I believe I have addressed all of them.

examples/kernels/flake.nix
danieldk: Yep

docs/source/_toctree.yml
```diff
@@ -47,6 +47,8 @@
 - sections:
     - local: builder/writing-kernels
       title: Write kernels
+    - local: builder/triton-autotune
```
danieldk: I think you forgot to remove it here.

danieldk (commented):
I think you forgot the docs link in one place.

examples/kernels/flake.nix
sayakpaul: Tracking here: https://github.com/huggingface/kernels/issues/770

docs/source/_toctree.yml
sayakpaul: Good catch. Removed.

danieldk (approved):

PR #761, opened 2026-08-13
fix: pin hf-hub post create pr patch

drbh:
this pr simply bumps hf-hub to include the changes that fix a bug with opening prs on external repos. patched in https://github.com/huggingface/hf-hub/pull/199

note* no tests were added since the hf-hub does not expose the internal urls (tested in the upstream pr) so we cannot check them easily, and a full integration tests requires interacting with the hub with multiple accounts (which was out of the scope of this fix).

### DIFF
--- a/kernel-builder/src/hf.rs
+++ b/kernel-builder/src/hf.rs
@@ -10,9 +10,9 @@ pub fn api() -> Result<hf_hub::HFClientSync> {
 pub fn repo_handle<T: RepoType>(api: &HFClientSync, repo_id: &str) -> HFRepositorySync<T> {
     let parts: Vec<&str> = repo_id.splitn(2, '/').collect();
     if parts.len() == 2 {
-        api.repository::<T>(parts[0], parts[1])
+        api.repository(T::default(), parts[0], parts[1])
     } else {
-        api.repository::<T>("", repo_id)
+        api.repository(T::default(), "", repo_id)
     }
 }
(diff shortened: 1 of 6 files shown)


kernel-builder/src/hf.rs
```diff
@@ -10,9 +10,9 @@ pub fn api() -> Result<hf_hub::HFClientSync> {
 pub fn repo_handle<T: RepoType>(api: &HFClientSync, repo_id: &str) -> HFRepositorySync<T> {
     let parts: Vec<&str> = repo_id.splitn(2, '/').collect();
     if parts.len() == 2 {
-        api.repository::<T>(parts[0], parts[1])
+        api.repository(T::default(), parts[0], parts[1])
```
sayakpaul, line 13: Could I get a primer on why we need `default()`?

sayakpaul (dismissed):
Thanks!

kernel-builder/src/hf.rs
drbh: yea this was added because the hf-hub library updated the `repository` method to expect the `RepoType` to be passed at runtime (in this PR https://github.com/huggingface/hf-hub/pull/164) and since this function is generic over T, T::default() provides the marker type (RepoTypeKernel or RepoTypeModel)

sayakpaul (approved):
Thanks David!

PR #765, opened 2026-08-17
kernels-data: replace `GitHash` by `GitStatus`

danieldk:
`GitHash` backed the Git commit SHA and the dirty status in a struct. But it is a bit of a misnomer, since `dirty` is not part of the hash. The issue that this gives was that we cannot use `GitHash` in other contexts (like lock files).

This change renames `GitHash` to `GitStatus`. It also adds typing to the commit SHA for validation. The newtype is named `Oid` to align with other libraries such as libgit2.

For compatibility with existing metadata files, we still parse the old field name (`sha`) as well. There is currently no release with provenance in the metadata yet, but since kernels are built against `main`, there are many kernels that use this field name already.

### DIFF
--- a/kernel-builder/src/pyproject/common.rs
+++ b/kernel-builder/src/pyproject/common.rs
@@ -1,10 +1,12 @@
 use std::path::PathBuf;
+use std::str::FromStr;
 
 use eyre::Result;
 use itertools::Itertools;
 
 use kernels_data::config::{Backend, Build};
-use kernels_data::metadata::{GitHash, KernelBuilderVersion, Metadata, Provenance};
+use kernels_data::git::{GitStatus, Oid};
+use kernels_data::metadata::{KernelBuilderVersion, Metadata, Provenance};
 
 use crate::pyproject::ops_identifier::KernelIdentifier;
 use crate::pyproject::FileSet;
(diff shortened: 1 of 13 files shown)


drbh (approved):
LGTM!

kernel-builder/src/pyproject/common.rs
```diff
@@ -1,10 +1,12 @@
 use std::path::PathBuf;
+use std::str::FromStr;
 
 use eyre::Result;
 use itertools::Itertools;
 
 use kernels_data::config::{Backend, Build};
-use kernels_data::metadata::{GitHash, KernelBuilderVersion, Metadata, Provenance};
+use kernels_data::git::{GitStatus, Oid};
+use kernels_data::metadata::{KernelBuilderVersion, Metadata, Provenance};
```
sayakpaul, line 9: Cool segregation. How does this handle metadata in the case where metadata also includes `GitHash`?

sayakpaul (approved):
Very nice!

kernel-builder/src/pyproject/common.rs
danieldk: Old serialized `GitHash` will be deserialized into `GitStatus`.

PR #766, opened 2026-08-17
`fetchFromHuggingFace`: add

danieldk:
Add a fetcher to get kernels/models/datasets from the Hub. This will be used in upcoming changes to fetch kernel-kernel dependencies.

### DIFF
--- a/nix-builder/pkgs/fetch-from-huggingface/default.nix
+++ b/nix-builder/pkgs/fetch-from-huggingface/default.nix
@@ -1,0 +1,68 @@
+{
+  lib,
+  cacert,
+  python3,
+  stdenvNoCC,
+}:
+
+let
+  # `hf download` only accepts the `model`, `dataset` and `space` repository
+  # types, so we use `snapshot_download` (which the CLI wraps) directly.
+  python = python3.withPackages (ps: [ ps.huggingface-hub ]);
+in
+
+{
+  # Repository id, e.g. `kernels-community/relu`.
+  repoId,
+
+  # Repository type, e.g. `model`, `dataset` or `kernel`.
+  type ? "model",
+
+  # Tag or commit SHA.
+  revision,
+
+  # SRI hash of the repository snapshot.
+  hash,
+
+  # Derivation name, defaults to `<repoId>-<revision>`.
+  name ? null,
+}:
+
+stdenvNoCC.mkDerivation {
+  name = lib.strings.sanitizeDerivationName (if name != null then name else "${repoId}-${revision}");
+
+  nativeBuildInputs = [ python ];
+
+  dontUnpack = true;
+  dontConfigure = true;
+  dontBuild = true;
+  dontFixup = true;
+
+  installPhase = ''
+    runHook preInstall
+
+    # Downloading into a local directory bypasses the hub cache, but hf-xet
+    # still needs a writable home for its chunk cache.
+    export HOME="$NIX_BUILD_TOP"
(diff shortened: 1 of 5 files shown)


nix-builder/pkgs/fetch-from-huggingface/download.py
```diff
@@ -0,0 +1,18 @@
...
+snapshot_download(
+    repo_id=repo_id,
+    repo_type=repo_type,
+    revision=revision,
+    local_dir=local_dir,
+)
+
+# Downloading leaves per-file metadata (etags, timestamps) behind, which would
+# make the hash of the snapshot unstable.
+shutil.rmtree(Path(local_dir) / ".cache", ignore_errors=True)
```
drbh, line 18: nit/question: is this safe? is there a case where the fetched repo legitimately contains files under `.cache`? if so, this would delete them too. could we remove only .cache/huggingface instead?

**I think it should be safe to remove the whole dir but wanted to double check

drbh (approved):
lgtm just left a question

nix-builder/pkgs/fetch-from-huggingface/download.py
danieldk: Good question. I'm a bit worried about impurities that would arise because dependencies of `huggingface_hub` are also allowed to write to `.cache`, so I think it's safer to yank the whole directory. But we should keep it in mind for the future.

nix-builder/pkgs/fetch-from-huggingface/default.nix
```diff
@@ -0,0 +1,68 @@
...
+
+  # Tag or commit SHA.
+  revision,
+
+  # SRI hash of the repository snapshot.
+  hash,
+
+  # Derivation name, defaults to `<repoId>-<revision>`.
+  name ? null,
+}:
```
sayakpaul, line 29: Feels illegal to write code this way!

nix-builder/pkgs/fetch-from-huggingface/default.nix
```diff
@@ -0,0 +1,68 @@
+{
+  lib,
+  cacert,
+  python3,
+  stdenvNoCC,
+}:
+
+let
```
sayakpaul, line 8: Should we filter only "kernel" type repos?

nix-builder/pkgs/fetch-from-huggingface/default.nix
```diff
@@ -0,0 +1,68 @@
...
+
+    # Downloading into a local directory bypasses the hub cache, but hf-xet
+    # still needs a writable home for its chunk cache.
+    export HOME="$NIX_BUILD_TOP"
+
+    # We use a custom script for now, since hf-cli does not support
+    # downloading the kernel repo type yet.
+    python ${./download.py} \
+      ${lib.escapeShellArg repoId} \
+      ${lib.escapeShellArg type} \
```
sayakpaul, line 52: Shouldn't we only allow "kernel" type repo downloads in this section?

sayakpaul (approved):
Nice. My only comment is maybe we should restrict this to "kernel" type repos for now.

nix-builder/pkgs/fetch-from-huggingface/default.nix
danieldk: I like having a generic fetchers, maybe we can get it upstreamed in nixpkgs at some point 😁 .

PR #767, opened 2026-08-17
fix: avoid skipping dirty arg in gitstatus

drbh:
this pr addresses the case where the `kernel-builder` git status fields are present but cannot be deserialized. previously, serde would silently drop invalid flattened fields and deserialize `git` as `None`, which could hide a `dirty: true` value.

`KernelBuilderVersion` is now deserialized through a repr type that distinguishes missing git fields from invalid ones. missing fields still deserialize as `None`, while incomplete or invalid fields return an error. legacy `sha` metadata remains supported.

a regression test covers an invalid commit with `dirty: true`

### DIFF
--- a/kernels-data/src/metadata.rs
+++ b/kernels-data/src/metadata.rs
@@ -1,11 +1,11 @@
 use std::str::FromStr;
 
 use eyre::Result;
-use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


danieldk (approved):
Great catch!

PR #768, opened 2026-08-18
kernel-builder: move `build.toml` parsing to `kernels-data`

danieldk:
This change moves over `parse_and_validate` to `kernels-data` and exposes it as `Build::open`. We also make this function available through the Python API as well as a small number of methods from `Build`.

The `Build` methods that are made available through the Python API are guided by what is needed for dependency support. In general, we do not want to do any `build.toml` parsing in `kernels`, since it's a build-side file format. However, to be able to build a kernel with dependencies in Nix, we need to lock the dependencies and this is the same locking code as used by `kernels lock`. The only difference is that we take the kernels from `build.toml` and not `pyproject.toml`. For this reason, exposing `build.toml` to the client in a minor way seems justified for this specific purpose.

### DIFF
--- a/kernels-data/bindings/python/kernels_data.pyi
+++ b/kernels-data/bindings/python/kernels_data.pyi
@@ -7,6 +7,8 @@ from typing import Optional, final
 __all__ = [
     "Backend",
     "BackendInfo",
+    "Build",
+    "General",
     "Provenance",
     "DigestAlgorithm",
     "GitStatus",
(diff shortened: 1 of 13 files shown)


kernels-data/bindings/python/kernels_data.pyi
```diff
@@ -324,6 +326,39 @@ class KernelDependency:
...
+
+    @property
+    def backends(self) -> list[Backend]:
+        """Backends the kernel supports."""
+        ...
+
+    def __repr__(self) -> str: ...
+
+@final
+class Build:
```
sayakpaul, line 341: Nice!

kernels-data/bindings/python/tests/test_config.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+
+from kernels_data import Backend, Build
+
+
+def _write_build_toml(path, backends):
```
sayakpaul, line 6: Should we consider having a test that uses a `build.toml` with a kernel as a dependency and checks if it gets parsed as expected?

sayakpaul (dismissed):
Thanks. Just one question regarding adding a test.

kernels-data/bindings/python/tests/test_config.py
danieldk: Great idea, added.

sayakpaul (approved):
Great!

PR #771, opened 2026-08-18
Add locking data structures to `kernels-data`

danieldk:
This change adds locking data structures to `kernels-data` and exposes them through the Python binding:

- `KernelLock`: a lock that stores the Git commit OID.
- `KernelLocks`: a mapping from `KernelDependency` to `KernelLock`.
- `NixKernelLock`: a lock that also stores the SRI hash of the Nix output path of a kernel.
- `NixKernelLocks`: a mapping from `KernelDependency` to `NixKernelLock`.
- `KernelPaths`: a mapping from `KernelDependency` to `Path` for kernels that are available locally. This will be used to point to Nix store paths after realizing `NixKernelLocks`, however it is more generic than just being for Nix - downstream users will also be able to use this to map kernel dependencies to local paths.

These data structures will be used inside `kernels` for lock files and inside `nix-builder` for lock files with output hashes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1576,6 +1576,7 @@ version = "0.17.0-dev0"
 dependencies = [
  "kernels-data",
  "pyo3",
+ "serde_json",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


drbh (approved):
really nice!

Issue #707, opened 2026-07-09
Process exits when architecture is not supported


remi-or:
### Description

This is a pretty ugly issue that arises when a kernel is loaded on a machine with a device for which it has no compatible architecture. Eg, when the FA3 kernel is loaded on a B200, loading happens cleanly (I think because torch and CUDA versions match) but then the process exits because the kernel finds no version compatible with the architecture of the B200. Since the process itself exits, this cannot be caught in python, so we don't even get an error message or a traceback, which may be the worst part.

### Steps to reproduce

```python
"""Minimal repro: `kernels` loads a kernel that cannot run on the current GPU, and the
first call kills the process instead of raising.

On a B200 (compute capability 10.0), `kernels-community/flash-attn3` only ships
sm_80/sm_90a binaries. `kernels` selects build variants by torch/CUDA version only, so
the load succeeds — then the first kernel launch hits flash-attn's `CHECK_CUDA` macro
(`fprintf` + `exit(1)`), which terminates the whole process with no Python traceback:

    $ python repro_fa3_b200.py
    device capability: (10, 0)
    kernel declares archs: ['8.0', '9.0a']
    CUDA error (/build/source/flash-attn/flash_fwd_launch_template.h:193): no kernel image is available for execution on the device
    $ echo $?
    1

Suggested fix by Claude: the build metadata (`metadata.json` -> `backend.archs`,
printed below) already declares the supported architectures, it is just never checked
against the device at load time. Probably will be more complex
"""

import kernels
import torch
from kernels import get_kernel, get_loaded_kernels


print(f"{kernels.__version__ = }")

kernel = get_kernel("kernels-community/flash-attn3", version=1)
print(f"\nLoading is done: {kernel}")

# flush=True matters: the C-level `exit(1)` below does not flush Python's stdout buffer,
# so without it even these lines are lost — the process dies with the CUDA error as its
# only output.
print("device capability:", torch.cuda.get_device_capability(), flush=True)
print("kernel declares archs:", get_loaded_kernels()[-1].metadata.backend.archs, flush=True)

q = torch.randn(1, 8, 4, 64, dtype=torch.bfloat16, device="cuda")
try:
    kernel.flash_attn_func(q, q, q, causal=True)  # process exits here with code 1
except Exception as e:
    print(f"Caught an error: {e}")

print("This is never reached :(", flush=True)
```

yields:
```
kernels.__version__ = '0.16.0'
[...]

PR #774, opened 2026-08-18
feat: arch/cc compatibility check.

sayakpaul:
Closes https://github.com/huggingface/kernels/issues/769 and thereby should also hopefully close https://github.com/huggingface/kernels/issues/707.

### DIFF
--- a/docs/source/basic-usage.md
+++ b/docs/source/basic-usage.md
@@ -65,6 +65,46 @@ for decision in get_kernel_variants("kernels-community/activation", version=1):
         print(f"{name}: rejected ({decision.reason})")
 ```
 
+## Device Architecture Checks
+
+A kernel build declares the device architectures it was built for (e.g. CUDA
+compute capabilities such as `8.0` or `9.0a`, or ROCm archs such as `gfx90a`)
+in its metadata. [`~kernels.get_kernel`] validates these architectures against
+the current device and raises an error when the device is not supported. This
+check can be disabled by passing `check_arch=False` to [`~kernels.get_kernel`].
+Without this check, an incompatible kernel would load fine, but fail at
+launch time — some kernels even terminate the whole process rather than
+raising an exception.
+
+[`~kernels.has_kernel`] performs the same validation and returns `False` when
+the kernel does not support the current device, so you can select a different
+kernel or fall back to another implementation without hard-failing.
+
+The declared architectures are the union over all of a kernel's components, so
+a kernel may support more architectures than it declares. For example, a
+kernel could fall back to a Triton implementation on architectures that its
+CUDA kernels were not built for. For such kernels, the check can be disabled
+with `check_arch=False`:
+
+```python
+from kernels import get_kernel, has_kernel
+
+# Only checks that a compatible build variant exists, the kernel may not
+# support the device architecture.
+has_kernel("kernels-community/flash-attn3", version=1, check_arch=False)
+
+# Skips the device architecture check when loading.
+flash_attn3 = get_kernel("kernels-community/flash-attn3", version=1, check_arch=False)
+```
+
+> [!WARNING]
+> When the architecture check is disabled, loading an incompatible kernel
+> succeeds, but using it can fail or even terminate the process at run time.
+
+For kernels that are used through layers, prefer
+[registering kernels for specific CUDA capabilities](./layers.md#registering-kernels-for-specific-cuda-capabilities)
+instead — `kernels` will then never load an incompatible kernel at all.
+
 ## Inspecting Loaded Kernels
 
 [`~kernels.get_loaded_kernels`] returns a snapshot of every kernel that has been loaded
--- a/kernels/src/kernels/archs.py
+++ b/kernels/src/kernels/archs.py
@@ -1,17 +1,25 @@
-import re
-
 from kernels_data import Backend, Metadata
 
 from kernels.compat import has_torch
 
 
-class UnsupportedArchError(RuntimeError):
-    """The kernel build does not support the architecture of the current device."""
+def _parse_cuda_arch(arch: str) -> tuple[int, int, str] | None:
+    """Parse a CUDA arch into its capability and suffix.
 
+    CUDA archs are compute capabilities like `9.0`, optionally with an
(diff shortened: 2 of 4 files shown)


docs/source/basic-usage.md
```diff
@@ -65,6 +65,46 @@ for decision in get_kernel_variants("kernels-community/activation", version=1):
         print(f"{name}: rejected ({decision.reason})")
 ```
 
+## Device Architecture Checks
```
sayakpaul, line 68: Still debating if this is the right place to include notes about it given it's supposed to be "basic" in nature. Open to suggestions.

kernels/src/kernels/archs.py
```diff
@@ -0,0 +1,98 @@
+import re
+
+from kernels_data import Backend, Metadata
+
+from kernels.compat import has_torch
+
+
+class UnsupportedArchError(RuntimeError):
```
sayakpaul: We could also just keep `RuntimeError` in case architecture and compute capabilities become twingled.

kernels/src/kernels/archs.py
```diff
@@ -0,0 +1,87 @@
+from kernels_data import Backend, Metadata
+
+from kernels.compat import has_torch
+
+
+def _parse_cuda_arch(arch: str) -> tuple[int, int, str] | None:
```
sayakpaul: We could have used Regex here but since they can be fragile (and this is somewhat easy to parse deterministically, I thought this is better.

remi-or:
Awesome, thanks! Re-tried the script I attached in #707 and it adequately crashed.

kernels/src/kernels/archs.py
```diff
@@ -0,0 +1,87 @@
...
+    suffix = ""
+    if arch.endswith(("a", "f")):
+        suffix = arch[-1]
+        arch = arch[:-1]
+
+    arch_major, sep, arch_minor = arch.partition(".")
+    if not sep or not arch_major.isdigit() or not arch_minor.isdigit():
+        return None
+
+    return int(arch_major), int(arch_minor), suffix
```
danieldk: We need to check for `+PTX` as well. E.g. if a supported capability is `9.0+PTX`, then the kernel should work on >= 9.0 as well.

kernels/src/kernels/archs.py
```diff
@@ -0,0 +1,87 @@
...
+    arch_major, arch_minor, suffix = parsed
+    major, minor = capability
+
+    if suffix == "a":
+        # Architecture-specific builds only run on that exact capability.
+        return (major, minor) == (arch_major, arch_minor)
+
+    # Base and family-specific builds run on capabilities of the same
+    # generation with the same or a newer minor version.
+    return major == arch_major and minor >= arch_minor
```
danieldk: Add PTX case.

kernels/src/kernels/archs.py
```diff
@@ -0,0 +1,87 @@
...
+    parsed = _parse_cuda_arch(arch)
+    if parsed is None:
+        return None
+
+    arch_major, arch_minor, suffix = parsed
+    major, minor = capability
+
+    if suffix == "a":
+        # Architecture-specific builds only run on that exact capability.
+        return (major, minor) == (arch_major, arch_minor)
```
danieldk: Handling of the `f` suffix is missing.

kernels/src/kernels/archs.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Base and family-specific builds run on capabilities of the same
+    # generation with the same or a newer minor version.
+    return major == arch_major and minor >= arch_minor
+
+
+def _cuda_archs_support_capability(archs: list[str], capability: tuple[int, int]) -> bool:
```
danieldk: I think we should use something more descriptive, e.g.

```suggestion
def _supports_cuda_capability(archs: list[str], capability: tuple[int, int]) -> bool:
```

kernels/src/kernels/archs.py
```diff
@@ -0,0 +1,87 @@
...
+    supports = [_cuda_arch_supports(arch, capability) for arch in archs]
+    if all(support is None for support in supports):
+        # None of the arch strings are in a known format (e.g. produced by a
+        # newer kernel-builder), so compatibility cannot be determined.
+        return True
+    return any(supports)
+
+
+def _arch_incompatibility(metadata: Metadata) -> str | None:
```
danieldk:
```suggestion
def _check_arch_incompatibility(metadata: Metadata) -> str | None:
```

I would also drop the returns and instead raise if the check fails. Inside `has_kernel` the exception can be caught.

Putting errors in `str` is a bit unpythonic.

kernels/src/kernels/archs.py
danieldk, line 59: The `+PTX` suffix, which we support is not handled:

- `10.0+PTX`: works on any capability >= 10.0.
- `10.0a+PTX`: only works on 9.0 (but embeds PTX, pretty useless, but it is supported).
- `10.0f+PTX`: works only on 10.x where x>=0 (for explicitness). Is also useful, because the runtime can JIT-compile the PTX for optimization if the capability is e.g. 10.3.

kernels/tests/test_archs.py
```diff
@@ -0,0 +1,135 @@
...
+def test_check_skipped_without_device(monkeypatch):
+    monkeypatch.setattr(torch.version, "cuda", "12.8", raising=False)
+    monkeypatch.setattr(torch.cuda, "is_available", lambda: False)
+    assert _arch_incompatibility(make_metadata("cuda", ["8.0"])) is None
+
+
+def test_check_skipped_for_backends_without_archs(fake_cuda_device):
+    # Archs of other backends cannot be checked against the current device.
+    assert _arch_incompatibility(make_metadata("cpu", None)) is None
+    assert _arch_incompatibility(make_metadata("metal", ["applegpu_g13"])) is None
```
danieldk: We should also check that it works for backends that support archs (e.g. CUDA), but don't have them specified (noarch kernels).

danieldk (approved):
Looks great!

@drbh could you also do an extra check, since we are directly rolling this out to stable?

drbh (approved):
lgtm!

PR #775, opened 2026-08-20
Add kernel resolver infrastructure

danieldk:
This change sets up the kernel resolver data structures and functions. These will be used to recursively solve kernel dependencies in subsequent PRs. A revolver is class with a method that returns where the kernel can be found (`LocalKernel` or `RemoteKernel`) given a `KernelDependency`. The current resolvers are:

- `HubResolver`: resolves a kernel from the Hub.
- `HubCacheResolver`: resolves a kernel from the local Hub cache.
- `LockedHubResolver`: resolves a kernel from the Hub using a lock file.
- `LockedHubCacheResolver`: resolves a kernel from the local Hub cache using a lock file.
- `RepoPathsResolver`: resolves kernels from a repo ID -> local path mapping.
- `KernelPathsResolver`: resolves kernels from a kernel dependency -> local path mapping.
- `SequentialResolver`: tries a list of resolvers in order until one resolves.
- `NoopResolver`: noop resolver

### DIFF
--- a/kernels/src/kernels/cli/info.py
+++ b/kernels/src/kernels/cli/info.py
@@ -3,6 +3,7 @@ import sys
 from pathlib import Path
 from typing import Any
 
+from huggingface_hub import constants
 from kernels_data import Metadata
 
 from kernels._versions import _get_available_versions, resolve_version_spec_as_ref
@@ -26,7 +27,10 @@ def print_kernel_info(
     path = Path(kernel)
     if path.is_dir():
         if revision is not None or version is not None:
-            print("--revision and --version cannot be used with a local path", file=sys.stderr)
+            print(
+                "--revision and --version cannot be used with a local path",
+                file=sys.stderr,
+            )
             sys.exit(1)
         info = _local_kernel_info(path)
     else:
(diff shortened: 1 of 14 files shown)


drbh (dismissed):
wooo the new resolvers are great! thanks for adding 🙏

kernels/src/kernels/cli/info.py
```diff
@@ -49,15 +53,18 @@ def _hub_kernel_info(
         sys.exit(1)
 
     if version is not None:
-        revision = resolve_version_spec_as_ref(repo_id, version).name
+        revision = resolve_version_spec_as_ref(repo_id, version, local_files_only=constants.HF_HUB_OFFLINE).name
```
danieldk, line 56: Just a note on these: this was determined locally in the functions, but this kind of global state makes it hard to reason about code, harder to test, and is really incompatible with the new resolvers. So switching to explicit passing here.

drbh:
the changes may be skipping the trusted publisher check? this small test fails to throw on a untrusted publisher.. might be because we need to make a remote call to check the trusted publisher status? may be worth investigating further?

```patch
diff --git a/kernels/tests/test_resolver.py b/kernels/tests/test_resolver.py
index 3ccb3d3..ed65d71 100644
--- a/kernels/tests/test_resolver.py
+++ b/kernels/tests/test_resolver.py
@@ -205,6 +205,32 @@ def test_hub_resolver_blocks_untrusted_org(api):
         )
 
 
+def test_hub_cache_resolver_blocks_untrusted_org(monkeypatch, api):
+    class UntrustedOrg:
+        trustedKernelPublisher = False
+
+    class FakeApi:
+        def get_organization_overview(self, publisher):
+            return UntrustedOrg()
+
+    monkeypatch.setattr("kernels.hf_hub._get_hf_api", lambda: FakeApi())
+    monkeypatch.setattr(
+        "kernels.resolver.resolve_kernel_version",
+        lambda *args, **kwargs: "a" * 40,
+    )
+    monkeypatch.setattr(
+        "kernels.resolver.resolve_hub_cache_kernel",
+        lambda *args, **kwargs: _local_kernel("/tmp/cached"),
+    )
+
+    with pytest.raises(ValueError, match="not from a trusted publisher"):
+        HubCacheResolver(trust_remote_code=False).resolve(
+            api=api,
+            backend="cpu",
+            kernel=_dep("untrusted-org/kernel"),
+        )
+
+
 @pytest.mark.cuda_only
 def test_hub_resolver_trust_remote_code_bypasses_check(api):
     location = HubResolver(trust_remote_code=True).resolve(

```


************UPDATE: this is reasonable since we only need to check trusted publishers are fetch time

drbh (approved):

danieldk:
I have run the tests in various combinations locally. Merging now since we don't seem to get runners. Will fix any fallout 🤞 .

PR #776, opened 2026-08-20
feat: experimental kernel porting tool

drbh:
> [!WARNING]
> This is an experiment/draft. The recipe language, the op set, and the CLI are all subject to change without notice. 

this pr adds `kernel-port`, an experimental tool for porting kernel repos into the kernel-builder layout by running a recipe instead of doing it by hand.

the idea is that porting an existing kernel to the kernel-builder is a set of deterministic rewrites/restructuring. a recipe is a list of those operations, which has the benefit of being able to be checked into the repo and provide a way to deterministically reproduce given a specific upstream commit.

one of the difficulties of maintaining a port is that the upstream repo can drift and currently keeping the port in sync is a manual process.

recipes are a way to codify the porting process, so that if the upstream repo drifts we can simply bump the commit in the recipe and re-run the porting process. if the upstream repo has changed in a way that breaks the porting process, the recipe will fail to apply and we can fix it before continuing.

a port is a `port.kdl` recipe plus an overlay dir of checked in files. same pins + same recipe gives a byte identical tree every time, and every op hard fails on drift rather than silently porting the wrong thing.

recipes are [kdl 2.0](https://kdl.dev) documents, one node per op:

```kdl
recipe version=1

source repo="https://github.com/org/repo" commit="<40 char sha>"
prune keep="csrc/**,pkg/**"
move from="csrc" to="kernel"
move from="pkg" to="torch-ext/pkg"
relativize_imports in="torch-ext/pkg/**" package_root="torch-ext/pkg" changes=6
kernel name="pkg" backend="cuda" src="kernel/**"
manifest name="pkg" backends="cuda" torch_src="torch-ext/*.cpp"
```

there are 15 ops (`source`, `vendor`, `prune`, `delete`, `move`, `overlay`, `replace`, `strip_suffix`, `expect`, `convert_import`, `remap_module`, `relativize_imports`, `ensure_init`, `kernel`, `manifest`). python rewrites go through libcst so comments and formatting are preserved byte for byte. `build.toml` is always generated by the `manifest` op, never overlaid.

the pins are the whole point. `count=`, `files=` and `changes=` are literals that have to match exactly, so a new file upstream cannot be rewritten without someone looking at it:

```
error: recipe line 17: relativize_imports: expected exactly 3 change(s) but made 4 - upstream drifted; review the new rewrites and update changes=
```

you can try an op without a checkout at all, `-e` takes the recipe inline and `--file path=content` builds the input tree in memory

```sh
cargo run -p kernel-port -- -e 'relativize_imports in="pkg/**" package_root="pkg" changes=1' \
    --file 'pkg/__init__.py=from pkg.ops import hello'
```

```
[line   1] relativize_imports  rewrote 1 import(s) in 1 file(s)
M pkg/__init__.py

FILE: pkg/__init__.py
from .ops import hello
```

the readme has a cookbook with one runnable command per op, plus the full arg list and failure modes for each. every example in it was run and pasted, not written by hand.

***NEXT STEPS are to target a JIT and AOT kernel in the kernels-community and experiment using this tool


sayakpaul:
> an overlay dir of checked in files

We might want to elaborate on what overlay means in this context. 

> recipe version=1

Is the manual specification for `version` inspired by how we do it in `build.toml` of kernels?

> the pins are the whole point. count=, files= and changes= are literals that have to match exactly, so a new file upstream cannot be rewritten without someone looking at it:

How are those pins derived?

> you can try an op without a checkout at all, -e takes the recipe inline and --file path=content builds the input tree in memory

Do the users have to specify `changes=` manually like that? Feels cumbersome no?

## More comments

* Yet another language 🥲
* For the code snippet provided in the description, I think we would want to first showcase the high-level structure of the kernel and then it'd be much easier to understand how those ops relate to one another in the context of the builder layout. 
* I didn't find anything that registers the PyTorch ops as per the specifications from https://huggingface.co/docs/kernels/main/en/builder/writing-kernels#registering-torch-operators. Is that intentional? I think this should be an atomic op unless I am missing something.
* Do we have the possibility that `overlay` directory can grow big? If so, wouldn't it be a problem to PR that upstream where we essentially want to just PR with the KDL file along with a lightweight `overlay` folder if needed?
*

drbh:
@sayakpaul 

> We might want to elaborate on what overlay means in this context. 

the docs on the readme may already fill this need https://github.com/huggingface/kernels/blob/b7062eed8821f6a1235680844ad4bbc4711d7a7c/kernel-port/README.md#overlay

copied for reference

> [!NOTE]
> #### `overlay`
> 
> ```kdl
> overlay from="<dir>"
> ```
> 
> Copy a directory of checked-in files over the workspace, overwriting what is there. `from` is relative to the recipe file. This is where the files with no upstream equivalent live: `torch_binding.cpp`, `flake.nix`, a CARD.md. Keep it small - anything derivable from upstream should be an op, not an overlay file, so that upstream drift is detected rather than papered over.

**the readme in the pr has a "cookbook" that explains what each operator in the recipe does and what args are optional, etc..

> Is the manual specification for `version` inspired by how we do it in `build.toml` of kernels?

the version is manually specified when writing the recipe and is only used so if the port tool has any breaking changes we can bump in the future

> How are those pins derived?

these are added by the users (either by manually counting) or more realisitcally through an iterative process. the tool is designed to be run iteratively making changes to the recipe and checking the output. the tool errors if the `count=, files= and changes=` does not match exactly, and reports the count of mutations the op would make. this ensure that every change is accounted for and tracked (and not silently applied when migrating to a newer commit)

example of how it errors

```
error: recipe line 45: expect: expected "torch-ext/einops/**" to match exactly 10 file(s), found 17 - the upstream file set drifted; update the port definition (torch-ext/einops/__init__.py, torch-ext/einops/_backends.py, torch-ext/einops/_torch_specific.py, torch-ext/einops/array_api.py, torch-ext/einops/einops.py, torch-ext/einops/experimental/__init__.py, torch-ext/einops/layers/__init__.py, torch-ext/einops/layers/_einmix.py, torch-ext/einops/layers/flax.py, torch-ext/einops/layers/keras.py, torch-ext/einops/layers/oneflow.py, torch-ext/einops/layers/paddle.py, torch-ext/einops/layers/tensorflow.py, torch-ext/einops/layers/torch.py, torch-ext/einops/packing.py, torch-ext/einops/parsing.py, torch-ext/einops/py.typed)
```

> Do the users have to specify `changes=` manually like that? Feels cumbersome no?

the `changes=` argument is optional in all cases https://github.com/huggingface/kernels/blob/b7062eed8821f6a1235680844ad4bbc4711d7a7c/kernel-port/README.md#rewriting-python-imports, but are intentionally verbose since pins (like changes) are included to make the number of mutations explicit and record. 

in the readme under conventions we include the following (copied for reference) https://github.com/huggingface/kernels/blob/b7062eed8821f6a1235680844ad4bbc4711d7a7c/kernel-port/README.md#cookbook

> [!NOTE]
> - Pins (count=, files=, changes=) are literal integers that must match exactly. They are the point: they turn "upstream changed" from a silent mis-port into a failed run.

so its recommended to include pins, however at the end of the day its fully optional

> Yet another language 🥲

yea true, but at least its really tiny (only our 15 ops) and no control flow/etc so shouldn't be much effort to understand. I think the parts that can be confusing are concepts like pins but sufficient docs and good error messages should make this manageable 

> For the code snippet provided in the description, I think we would want to first showcase the high-level structure of the kernel and then it'd be much easier to understand how those ops relate to one another in the context of the builder layout. 

added examples of a working port for einops for this reason in https://github.com/huggingface/kernels-community/pull/1092 and how to update the pin in https://github.com/huggingface/kernels-community/pull/1093

> I didn't find anything that registers the PyTorch ops as per the specifications from https://huggingface.co/docs/kernels/main/en/builder/writing-kernels#registering-torch-operators. Is that intentional? I think this should be an atomic op unless I am missing something.

yes there are no recipe operations for torch ops specifically - this case is handled by the generic `replace` op, this way the exact string replacement for the ops is stored in the recipe (which makes them a bit longer), but the benefit of a generic op is that it works in all cases.

> Do we have the possibility that `overlay` directory can grow big? If so, wouldn't it be a problem to PR that upstream where we essentially want to just PR with the KDL file along with a lightweight `overlay` folder if needed?

yes 100%, overlays will be big for kernels that are a combination of modified upstream and custom implementations we have in kernels-community. basically `overlay` is a way to inject files that follow kernel-builder conventions (no mutation needed). so in the case we are opening a pr for an external kernel, there is not really a case where we need to inject files (only the flake.nix) but if they dont already have a flake they can just add that to their root and we'll pull it in in the recipe.

tldr; in the pr case we'll just add two files `port.kdl` and `flake.nix`. the overlay is mostly a concept that caters to our situation where we have vendored kernels with non upstream additions.

sayakpaul:
> these are added by the users (either by manually counting) or more realisitcally through an iterative process. the tool is designed to be run iteratively making changes to the recipe and checking the output. the tool errors if the count=, files= and changes= does not match exactly, and reports the count of mutations the op would make. this ensure that every change is accounted for and tracked (and not silently applied when migrating to a newer commit)

Very good idea, IMO. But would we have a simple way to quickly test the correctness of its implementation? Not a blocker but I think we should strive for simplicity here. 

> so its recommended to include pins, however at the end of the day its fully optional

Maybe we need to distinguish between required pins and optional pins? `count` and `files`feel like they should be the required ones. 

> yes there are no recipe operations for torch ops specifically - this case is handled by the generic replace op, this way the exact string replacement for the ops is stored in the recipe (which makes them a bit longer), but the benefit of a generic op is that it works in all cases.

Cool, that works for me!

> basically overlay is a way to inject files that follow kernel-builder conventions (no mutation needed). so in the case we are opening a pr for an external kernel, there is not really a case where we need to inject files (only the flake.nix) but if they dont already have a flake they can just add that to their root and we'll pull it in in the recipe.

Not sure if I fully understood it. Why would it differ for an external repo from `kernels-community`? Is it just because of the top-level flake?

> tldr; in the pr case we'll just add two files port.kdl and flake.nix. the overlay is mostly a concept that caters to our situation where we have vendored kernels with non upstream additions.

Oh okay. I was under the impression that without the `overlay` folder (and with just the KDL recipe), it wouldn't build according to our compliance. So, the scope of `overlay` seemed a bit confusing to me. Possible to provide a bit more elaboration?

nix-builder/overlay.nix
```diff
@@ -10,6 +10,8 @@ final: prev:
   # Local packages/hooks.
   kernel-builder = final.callPackage ./pkgs/kernel-builder { inherit builderProvenance; };
 
+  kernel-port = final.callPackage ./pkgs/kernel-port { };
```
sayakpaul, line 13: Any reasoning behind making it a part of our nix-builder? For future CI?

sayakpaul (commented):
Thanks just left a bunch of comments. I think we are headed in a good direction. Once this takes a bit more shape, we could think about how we wire this in the CI, etc.

kernel-port/src/ops/git.rs
```diff
@@ -0,0 +1,65 @@
+use anyhow::{Context, Result, bail};
```
sayakpaul, line 1: Do we require users to _always_ specify a SHA from which they would like to start the port? It could be beneficial to also let them specify a release tag or something so that one could say we're up-to-date upto this tag (particularly helpful for release tags). 

But not strongly opinionated about this, definitely.

kernel-port/README.md
```diff
@@ -0,0 +1,616 @@
+# kernel-port
```
sayakpaul, line 1: README is fine but we should find a way to make this available in our docs so that it's discoverable (particularly helpful for agents). Cc: @stevhliu here.

kernel-port/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "kernel-port"
+version = "0.1.0"
+edition = "2024"
```
sayakpaul, line 4: 2026?

nix-builder/pkgs/kernel-port/default.nix
```diff
@@ -0,0 +1,65 @@
...
+    outputHashes = {
+      "hf-hub-1.1.0" = "sha256-wClUTCmphrO4QM+IYwYrNxyvDp8qBGAPdP+Wca8TgRA=";
+    };
+  };
+
+  cargoBuildFlags = cargoFlags;
+  cargoTestFlags = cargoFlags;
+
+  meta = {
+    description = "Port third-party kernels to the Hugging Face kernel layout";
```
sayakpaul:
```suggestion
    description = "Port third-party kernels to the Hugging Face Kernels layout";
```

kernel-port/src/python.rs
```diff
@@ -0,0 +1,992 @@
+// Python edits go through libcst, so comments, quoting and layout outside the
```
sayakpaul, line 1: What is the formatting situation here? Do we follow it from upstream? In some of the port PRs on `kernels-community`, we saw a significant amount of formatting-related changes.

kernel-port/README.md
```diff
@@ -0,0 +1,616 @@
...
+
+## What the runner enforces
+
+On top of whatever the recipe says:
+
+- `source` and `vendor` require a full 40-character SHA and a clean checkout, untracked files included.
+- Every op is built before the first one runs, so an argument typo or a bad glob anywhere fails the recipe before anything is mutated.
+- After the last op, every added or modified Python file must still parse.
+- No absolute in-package import may remain under `torch-ext`. The Hub loads a kernel under a build-variant directory name, so only relative intra-package imports resolve at run time.
+- `--out` runs write a `.port-provenance.json`: recipe hash, runner version, pinned sources, output tree hash. All deterministic, so the file reproduces byte-for-byte.
```
sayakpaul: We're doing `port-provenance.json` right?

sayakpaul (commented):
Left some comments. Thanks for the further updates!

Apart from the comments, I think we should log the dirty status in the port provenance.

kernel-port/src/ops/git.rs
drbh: yea the full sha is intentional since tags are mutable and the pins only work if the source cannot move under the recipe. we could allow an optional `tag=` thats recorded in the provenance for readability, with `commit=` staying required. will add if theres interest

kernel-port/README.md
drbh: agreed. plan is to keep the readme as the source for now and mirror it into the docs once the op set settles, since its still experimental and the ops may change

kernel-port/Cargo.toml
drbh: thats the rust edition, 2024 is the latest one, I'm not sure when the next edition is planned, maybe 2027

nix-builder/pkgs/kernel-port/default.nix
drbh: updated with latest changes

kernel-port/src/python.rs
drbh: the python rewrites go through libcst so anything the tool touches keeps upstream formatting and comments byte for byte. the formatting diffs in the port prs come from overlay files and `replace` ops, where the content is whatever the recipe says, not from the rewriter

kernel-port/README.md
drbh: good catch, the file was renamed to `port-provenance.json` (no leading dot) in c99b104 but the readme still had the old name in a few places. fixed thanks

kernel-port/src/python.rs
sayakpaul: Fine for now but essentially, formatting related changes should be very minimal IMO at least for existing kernels. Otherwise, it's just a lot of reviewing time that could have been spent elsewhere.

kernel-port/src/tests.rs
```diff
@@ -0,0 +1,740 @@
+use crate::{ops, python, recipe, workspace::Workspace};
```
sayakpaul, line 1: I didn't go through this fully, but on a high level, I would expect to see the tests covering both AOT and JIT-flavored kernels. If already done, no need for further modifications. If not, then let's do that.

I think tests for atomic ops that we will have in the tool are good but lightweight e2e tests also go a long way (the way we use the RELU example kernel in tests, for example).

sayakpaul (commented):
Thanks a lot for the iterations. Just one comment regarding the tests.

kernel-port/src/tests.rs
drbh: really good point e2e tests for the two cases is really useful! thanks

just updated to include an example for both AOT and JIT kernels. each example has a `upstream`, `expected` dir and a `port.kdl`, then in CI we run the porting tool on `upstream` and check that its identical to the valid kernel format in `expected`

.github/workflows/rust.yaml
```diff
@@ -56,8 +64,12 @@ jobs:
           ( cd kernels && cargo clippy -- -D warnings )
       - name: Clippy (kernel-builder)
         run: ( cd kernel-builder && cargo clippy -- -D warnings )
+      - name: Clippy (kernel-port)
+        run: ( cd kernel-port && cargo clippy -- -D warnings )
 
   test:
+    permissions:
```
sayakpaul, line 71: Running the tests on `kernel-porter` for every rust-related changes doesn't sound good to me. Maybe we should have a separate workflow for the porter and trigger it on changes specific to the porter?

sayakpaul (approved):
One optional comment which could be addressed in a follow-up.

.github/workflows/rust.yaml
drbh: good point, I'll follow up with improved ci in the next porter changes. also lets see how long the tests take to run, if its very small it might be more simple to keep in the rust workflow (will check when merged)

PR #779, opened 2026-08-21
Add kernel dependency tree resolver

danieldk:
This change add basic dependency support:

- `DepTreeNode`: the kernel dependency tree data structure.
- `resolve_kernel_tree`: constructs the dependency tree for a kernel using a solver.
- `get_kernel_dep`: function that can be used by a kernel to get one of its dependencies.

Note: `get_kernel_dep` is not lined in the documentation yet, since not all bits to actually use it have landed.

### DIFF
--- a/kernels/src/kernels/python_deps.py
+++ b/kernels/src/kernels/python_deps.py
@@ -1,0 +1,106 @@
+import importlib.util
+import json
+from dataclasses import dataclass, field
+from pathlib import Path
+
+from huggingface_hub.dataclasses import strict
+from kernels_data import Metadata
+
+from kernels.backends import Backend, _backend
+
+
+@strict
+@dataclass
+class PythonPackage:
+    pkg: str
+    import_name: str | None
+
+    @staticmethod
+    def from_dict(data: dict) -> "PythonPackage":
+        return PythonPackage(
+            pkg=data["pkg"],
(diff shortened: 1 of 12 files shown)


kernels/src/kernels/python_deps.py
```diff
@@ -0,0 +1,106 @@
+import importlib.util
```
danieldk, line 1: This file is the old `deps.py`, no changes.

kernels/src/kernels/resolver.py
```diff
@@ -36,7 +36,7 @@ class LocalKernel:
     metadata: Metadata
     origin: "RemoteKernel | None" = None
 
-    def install(self, *, api: HfApi, all_variants: bool = False) -> Self:
+    def install(self, *, api: HfApi) -> Self:
```
danieldk: `all_variants` was a leftover from an experiment, not needed anymore.

kernels/tests/test_python_deps.py
```diff
@@ -0,0 +1,72 @@
+from importlib.util import find_spec
```
danieldk, line 1: This file is the old `test_deps.py`, GH doesn't visualize the renames well.

drbh (approved):
lgtm!

PR #784, opened 2026-08-25
Add support for Torch 2.14 (2.14-rc6)

danieldk:
Mostly smooth sailing, except ROCm 7.14, which will be added in a subsequent PR.

### DIFF
--- a/nix-builder/pkgs/python-modules/torch/binary/generic.nix
+++ b/nix-builder/pkgs/python-modules/torch/binary/generic.nix
@@ -77,6 +77,7 @@ let
   aotritonVersions = with rocmPackages; {
     "2.12" = aotriton_0_11_2;
     "2.13" = aotriton_0_12;
+    "2.14" = aotriton_0_13;
   };
 
   aotriton =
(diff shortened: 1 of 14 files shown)


nix-builder/pkgs/python-modules/triton-xpu/default.nix
```diff
@@ -18,4 +18,9 @@ in
     hash = "sha256-jMwVkt3aI6sunRQ9orS2nrQIIRcsW9S8TiYjI9x/UXI=";
   };
 
+  triton-xpu_3_8_0 = generic {
+    version = "3.8.0";
+    url = "https://huggingface.co/buckets/danieldk/pytorch-rc/resolve/2.14.0/rc6/triton_xpu-3.8.0-cp313-cp313-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl";
```
sayakpaul, line 23: We should maybe start hosting these under `kernels-community` buckets.

nix-builder/pkgs/python-modules/torch/binary/generic.nix
```diff
@@ -62,6 +62,8 @@
   effectiveStdenv ? if cudaSupport then cudaPackages.backendStdenv else stdenv,
 }:
 let
+  inherit (lib) versionAtLeast;
```
sayakpaul, line 65: Cool!

sayakpaul (approved):

nix-builder/pkgs/python-modules/triton-xpu/default.nix
danieldk: Yep, let's do that for the next RC going forward.

PR #785, opened 2026-08-25
kernels: switch over to the dependency solver

danieldk:
This change switches all the common kernel getter functions (`get_kernel`, etc.), `install`, and locking over to the dependency solver.


drbh (approved):
looks good to me! 

small nit is that we may want to followup with more tests for `locking.py` since the reported coverage is very low - but not a blocker

PR #786, opened 2026-08-25
nix-builder: wire up kernel dependencies

danieldk:
- The new `lock-kernel-deps` command creates a `kernels.lock` file with dependency locks and Nix output hashes. These are used to make fixed-output derivations of kernel dependencies.
- `fetchKernelDeps` is a function that takes a `kernels.lock` file and generates a JSON file that is compatible with the `KernelPaths` data structure from `kernels-data`.
- The kernel builder derivations take a new argument `kernelDeps`, which is used to pass the `KernelPaths` JSON file.
- `get-kernel-check` is updated to read the JSON file and construct a `KernelPathsResolver` that is used by `kernels` to resolve the dependencies from the Nix store.

Aside from that, a new example `kernel-deps` to test building and loading with dependencies.

I have also split the `deps.nix` file into `python-deps.nix` and `cpp-deps.nix` for clarity.

### DIFF
--- a/docs/source/builder/writing-kernels.md
+++ b/docs/source/builder/writing-kernels.md
@@ -2,7 +2,7 @@
 
 > [!TIP]
 > Refer to the [kernel requirements' page](../kernel-requirements.md) to
-> get an idea of what is expected from the kernel structure and content. 
+> get an idea of what is expected from the kernel structure and content.
 
 ## Introduction
 
@@ -320,7 +320,7 @@ The following options can be set for a kernel:
   supported backends are `cpu`, `cuda`, `metal`, `rocm`, and `xpu`.
   **The `cpu` backend is currently experimental and might still change.**
 - `depends` (required): a list of dependencies. The supported dependencies
-  are listed in [`deps.nix`](https://github.com/huggingface/kernels/blob/main/builder/lib/deps.nix).
+  are listed in [`cpp-deps.nix`](https://github.com/huggingface/kernels/blob/main/builder/lib/cpp-deps.nix).
 - `src` (required): a list of source files and headers.
 - `include` (optional): include directories relative to the project root.
   Default: `[]`.
--- a/examples/kernels/kernel-deps/build.toml
+++ b/examples/kernels/kernel-deps/build.toml
@@ -1,0 +1,23 @@
+[general]
+name = "kernel-deps"
+version = 1
+edition = 5
+license = "apache-2.0"
+backends = [
+    "cpu",
+    "cuda",
+    "metal",
+    "rocm",
+    "xpu",
+]
+
+kernel-depends = [
+    { repo-id = "kernels-community/einops", version = 1 },
+]
+
+[general.hub]
+repo-id = "kernels-test/kernel-deps"
+
+[torch-noarch]
+
+[kernel]
--- a/examples/kernels/kernel-deps/build.toml
+++ b/examples/kernels/kernel-deps/build.toml
@@ -12,6 +12,7 @@ backends = [
 ]
 
 kernel-depends = [
+    { repo-id = "kernels-community/activation", version = 1 },
     { repo-id = "kernels-community/einops", version = 1 },
 ]
 
--- a/nix-builder/lib/cpp-deps.nix
+++ b/nix-builder/lib/cpp-deps.nix
@@ -1,0 +1,47 @@
+{
+  lib,
+  pkgs,
+  stdenv,
+  torch,
+}:
+
+let
+  deps = {
+    "cutlass_2_10" = [
+      pkgs.cutlass_2_10
+    ];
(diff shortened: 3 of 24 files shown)


sayakpaul:
Haven't started reviewing yet but a question on

> get-kernel-check is updated to read the JSON file and construct a KernelPathsResolver that is used by kernels to resolve the dependencies from the Nix store.

If `kernels` were to solely lie on the consumer side (no builder component), how is this surfaced?

examples/kernels/kernel-deps/build.toml
```diff
@@ -0,0 +1,23 @@
...
+    "cpu",
+    "cuda",
+    "metal",
+    "rocm",
+    "xpu",
+]
+
+kernel-depends = [
+    { repo-id = "kernels-community/einops", version = 1 },
+]
```
sayakpaul, line 17: Do you think it's worth having one more depends here just to test that we're not limited to just one kernel dependency?

Additionally, is this expected to work seamlessly with `python-depends`? I know we want to get rid of that completely. But till the time we don't, how is that supposed to play with `kernel-depends`?

docs/source/builder/writing-kernels.md
```diff
@@ -2,7 +2,7 @@
 
 > [!TIP]
 > Refer to the [kernel requirements' page](../kernel-requirements.md) to
-> get an idea of what is expected from the kernel structure and content.
```
sayakpaul, line 5: Have we already made modifications to our docs to talk about kernel dependency?

nix-builder/lib/cpp-deps.nix
```diff
@@ -0,0 +1,47 @@
+{
```
sayakpaul, line 1: Very nice segregation!

nix-builder/pkgs/get-kernel-check/get-kernel-check-hook.sh
```diff
@@ -42,7 +47,30 @@ _getKernelCheckHook() {
...
+  RepoPathsResolver(local_kernels={'${out}': Path('${out}')}),
+  KernelPathsResolver(kernel_paths=kernel_paths)
+]
+
+get_kernel_with_resolver(
+  api=_get_hf_api(),
+  backend=None,
+  kernel=kernel,
+  resolver=SequentialResolver(resolvers=resolvers)
+)
```
sayakpaul: Can we keep it in a separate Python file?

nix-builder/pkgs/lock-kernel-deps/lock_kernel_deps.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
+"""Lock a kernel's dependencies and compute the Nix hashes.
+
+Reads the kernel's `build.toml`, resolves its dependencies to commits, and
+outputs a lock file with a `hash` field added to every lock. The hash is the
+SRI hash of a snapshot of the kernel repository, as used by the
+`fetchFromHuggingFace` fixed-output derivation.
+"""
+
+import argparse
```
sayakpaul, line 10: (nit): is there any merit in introducing a directory called `py_utils` here and keeping all the Python utilities inside it?

sayakpaul (commented):
Left some small comments. Looks exceptionally clean!

examples/kernels/kernel-deps/build.toml
danieldk: I can add another.

There should be no interaction with `python-depends` at all.

docs/source/builder/writing-kernels.md
danieldk: No, still intentionally not there, we should play a little with it before making it an 'official' feature.

nix-builder/pkgs/lock-kernel-deps/lock_kernel_deps.py
danieldk: I think support files for Nix functions/derivations like this should be in the directory of the derivation itself. This also keeps things close to upstream nixpkgs.

danieldk:
> Haven't started reviewing yet but a question on
> 
> > get-kernel-check is updated to read the JSON file and construct a KernelPathsResolver that is used by kernels to resolve the dependencies from the Nix store.
> 
> If `kernels` were to solely lie on the consumer side (no builder component), how is this surfaced?

It is not. In the future, when we open up the resolver API, downstream users can use the API to use and combine resolvers directly like we do in get-kernel-check.

sayakpaul (approved):

nix-builder/pkgs/get-kernel-check/get-kernel-check-hook.sh
danieldk: Yep, fixed!

PR #787, opened 2026-08-26
Small Torch 2.14 fixes

danieldk:
- Set sycl-tla version for oneAPI 2026.1.0.
- Switch from loose CUDA architecture intersection to strict.

### DIFF
--- a/kernel-builder/src/pyproject/templates/kernel.cmake
+++ b/kernel-builder/src/pyproject/templates/kernel.cmake
@@ -33,7 +33,7 @@ function(cuda_kernel_component SRC_VAR)
 
     # Determine CUDA architectures
     if(KERNEL_CUDA_CAPABILITIES)
-        cuda_archs_loose_intersection(_KERNEL_ARCHS "${KERNEL_CUDA_CAPABILITIES}" "${CUDA_ARCHS}")
+        cuda_archs_intersection(_KERNEL_ARCHS "${KERNEL_CUDA_CAPABILITIES}" "${CUDA_ARCHS}")
         if(NOT _KERNEL_ARCHS)
             message(FATAL_ERROR "CUDA kernel: ${KERNEL_NAME}, empty set of capabilities after intersection (kernel: ${KERNEL_CUDA_CAPABILITIES}, supported: ${CUDA_ARCHS})")
         endif()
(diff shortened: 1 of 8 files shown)


kernel-builder/src/pyproject/templates/kernel.cmake
```diff
@@ -33,7 +33,7 @@ function(cuda_kernel_component SRC_VAR)
 
     # Determine CUDA architectures
     if(KERNEL_CUDA_CAPABILITIES)
-        cuda_archs_loose_intersection(_KERNEL_ARCHS "${KERNEL_CUDA_CAPABILITIES}" "${CUDA_ARCHS}")
+        cuda_archs_intersection(_KERNEL_ARCHS "${KERNEL_CUDA_CAPABILITIES}" "${CUDA_ARCHS}")
```
sayakpaul, line 36: Would this have any backward breaking behaviour? I am guessing not because we only maintain the last two recent PyTorch versions. But just checking.

kernel-builder/src/pyproject/templates/utils.cmake
```diff
@@ -341,38 +341,46 @@ endmacro()
...
-#  architecture in `SRC_CUDA_ARCHS`.
-# The loose intersection is defined as:
-#   { max{ x \in tgt | x <= y } | y \in src, { x \in tgt | x <= y } != {} }
-#  where `<=` is the version comparison operator.
-# In other words, for each version in `TGT_CUDA_ARCHS` find the highest version
-#  in `SRC_CUDA_ARCHS` that is less or equal to the version in `TGT_CUDA_ARCHS`.
-# We have special handling for x.0a, if x.0a is in `SRC_CUDA_ARCHS` and x.0 is
-#  in `TGT_CUDA_ARCHS` then we should remove x.0a from `SRC_CUDA_ARCHS` and add
-#  x.0a to the result (and remove x.0 from TGT_CUDA_ARCHS).
+#  `<major>.<minor>[letter][+PTX]` compute the intersection with the
```
sayakpaul, line 344: Would be nice to also include the example of megablocks that triggered this.

sayakpaul (approved):
Really nice especially the test!

kernel-builder/src/pyproject/templates/kernel.cmake
danieldk: It should be fine, since in practice we have every CUDA capability listed anyway.

kernel-builder/src/pyproject/templates/utils.cmake
danieldk: Seems too specific for the docs, but it is covered in the tests below.

PR #791, opened 2026-08-26
kernel-builder: upload README to version branch as well

danieldk:


kernel-builder/src/upload.rs
```diff
@@ -777,7 +788,48 @@ mod tests {
     fn test_collect_readme_commit_ops_no_card() {
         let temp_dir = tempfile::tempdir().unwrap();
         let mut operations = vec![];
-        collect_readme_commit_ops(temp_dir.path(), &mut operations);
+        collect_readme_commit_ops(temp_dir.path(), &BTreeSet::new(), &mut operations);
+        assert!(operations.is_empty());
+    }
+
```
sayakpaul: Shouldn't we also test if the card is uploaded to the provided version as well?

sayakpaul (dismissed):
Thanks! Just one comment.

sayakpaul (approved):
Very nice and thanks for iterating!

PR #794, opened 2026-08-28
kernels-data: strongly type version length

### LINKED ISSUE #792: Version round-trip through serde strips trailing zeros, rewriting version bounds in build.toml
`Version` normalizes away trailing zero components when constructed, so a version string does not survive a deserialize → serialize round-trip:

- `kernels-data/src/version.rs`: `impl From<Vec<usize>> for Version` drops trailing zeros ([version.rs#L55](https://github.com/huggingface/kernels/blob/8c59f170a331dba74b7dcf921b40421c2b353e37/kernels-data/src/version.rs#L55)), and `FromStr` goes through that `From` impl.
- `Serialize` / `Display` then emit the normalized form, so `"13.0"` comes back out as `"13"`.

Concrete fallout: an automatic `build.toml` update in kernels-community silently rewrote the CUDA bounds `maxver = "13.0"` → `maxver = "13"` in `flash-attn3` and `vllm-flash-attn3`, changing the meaning of the bound even though these are plain strings that should have been left untouched. Fixup PR: https://github.com/huggingface/kernels-community/pull/1112

Ideas / things to investigate:
- Drop the trailing-zero-skipping normalization entirely, and instead make comparison zero-padding-aware so `13` and `13.0` still compare equal without mutating the stored representation.
- Alternatively, keep the original string (or component count) on `Version` so serialization is lossless, and only normalize for comparison.
- Regardless of the fix, the automated `build.toml` updater should not rewrite version fields it isn't intentionally changing.

Probably needs a bit more investigation into where normalization is actually relied upon before removing it.

---
Requested by Daniël de Kok - [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1787838605066429?thread_ts=1787838605.066429&cid=C090JN2P8NB) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-08-27T13-51-50-744Z_1b6a8dac-9343-4c05-a5ce-7b215f49dc5b.jsonl)

danieldk:
`Version` stored versions in a dynamic-length `Vec` and would trim any trailing zeros (e.g. `13.0` -> `13`). This leads to issues when we re-serialize `build.toml` files where we actually meant `13.0` (e.g. for Nix version comparisons, `13.0` and `13` are different).

We could simply remove the zero-trimming logic, but this also does not seem correct, because a user could e.g. use 13.0.0 for a CUDA version, but this breaks some Nix logic as well and that is not the granularity we want to use for e.g. CUDA and Torch versions.

So instead, add the version length as a type parameter, e.g. `Version<2>`. This ensures that we are always serializing correctly and catches erroneous version use by the user (e.g `13` or `13.0.0` when `N=2`).

For the Python `Version` type, we retain the existing behavior in order to not break kernels for now that do not have a `kernels-data` upper pin. However, we store the version as `Box<[usize]>`, so that we can use the same Python class if we want to embed Rust typed versions in the future (since we can return a slice for `Version<N>`).

Fixes #792.

### DIFF
--- a/kernels-data/bindings/python/tests/test_version.py
+++ b/kernels-data/bindings/python/tests/test_version.py
@@ -1,0 +1,26 @@
+import pytest
+
+from kernels_data import Version
+
+
+def test_version_parse_and_normalize():
+    assert str(Version.from_str("12.8.0")) == "12.8"
+    assert str(Version.from_str("1")) == "1"
+    assert str(Version.from_str("1.2.3")) == "1.2.3"
+
+
+def test_version_ordering_and_hash():
+    v1 = Version.from_str("1.2")
+    v2 = Version.from_str("1.2.0")
+    v3 = Version.from_str("1.3")
+    assert v1 == v2
+    assert v1 < v3
+    assert hash(v1) == hash(v2)
+    assert {v1, v2, v3} == {v1, v3}
+
+
+def test_version_invalid():
+    with pytest.raises(ValueError):
+        Version.from_str("abc")
+    with pytest.raises(ValueError):
+        Version.from_str("")
--- a/kernels-data/bindings/python/src/lib.rs
+++ b/kernels-data/bindings/python/src/lib.rs
@@ -8,44 +8,17 @@ use kernels_data::config::{Backend, KernelDependency, KernelName, KernelVersion}
 use kernels_data::digest::{Digest, DigestAlgorithm, DigestViolation};
 use kernels_data::git::{GitStatus, Oid};
 use kernels_data::metadata::{BackendInfo, KernelBuilderVersion, Metadata, Provenance};
-use kernels_data::version::Version;
 use pyo3::Bound as PyBound;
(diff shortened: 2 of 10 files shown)


kernels-data/bindings/python/tests/test_version.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from kernels_data import Version
+
+
+def test_version_parse_and_normalize():
+    assert str(Version.from_str("12.8.0")) == "12.8"
+    assert str(Version.from_str("1")) == "1"
+    assert str(Version.from_str("1.2.3")) == "1.2.3"
```
sayakpaul, line 9: One for 13.0 as well since it came up quite a bit? 😁

sayakpaul (approved):
Clear

kernels-data/bindings/python/tests/test_version.py
danieldk: 😁 but not in a Python context though.

PR #796, opened 2026-08-28
kernels: support conditions on `use_kernel_forward_from_hub`

danieldk:
Add and optional `condition` argument to `use_kernel_forward_from_hub`.
This condition accepts an instance of the layer and returns whether
kernelization of the layer should proceed. For instance,

```
@use_kernel_forward_from_hub(
    "SwiGLUMLP",
    condition=lambda module: module.config.hidden_act == "silu",
)
class MyMLP(nn.Module):
    ...
```

would only kernelize the layer using a mapping registered for `SwiGLUMLP`
its `hidden_act` configuration is `silu`.

### DIFF
--- a/kernels/tests/test_layer.py
+++ b/kernels/tests/test_layer.py
@@ -423,6 +423,78 @@ def test_layer_fallback_works():
     kernelize(silu_and_mul, device="cuda", mode=Mode.INFERENCE)
 
 
+def test_kernel_condition_skips_kernelization():
+    @use_kernel_forward_from_hub("SiluAndMulNonExisting", condition=lambda module: False)
+    class SiluAndMulConditionSkipped(SiluAndMul):
+        pass
+
+    silu_and_mul = SiluAndMulConditionSkipped()
+
+    kernelize(silu_and_mul, device="cuda", mode=Mode.INFERENCE, use_fallback=False)
+
+    # The forward was not replaced...
+    assert "forward" not in silu_and_mul.__dict__
+
+    # ... and the original implementation is still used.
+    X = torch.randn(32, 64)
+    silu_and_mul(X)
+    assert silu_and_mul.n_calls == 1
+
+
+def test_kernel_condition_holds():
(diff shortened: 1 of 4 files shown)


vasqu (approved):
Careful approval but I sanity checked on https://github.com/huggingface/transformers/pull/48335 (llama) and it worked as expected, e.g. when passing a new config with activation == gelu, we no longer kernelize that module

Ofc, docs are missing but since it's a draft I expect it tbh 🤗

danieldk:
Thanks for testing! I'll add the docs.

kernels/tests/test_layer.py
```diff
@@ -423,6 +423,78 @@ class SiluAndMulWithKernelFallback(SiluAndMul):
...
+    with pytest.raises(ValueError, match="No layer mapping for `SiluAndMulNonExisting`"):
+        kernelize(attempted, device="cuda", mode=Mode.INFERENCE, use_fallback=False)
+
+
+def test_kernel_condition_defaults_to_true():
+    # Without an explicit condition, kernelization is never skipped. Accessing
+    # the condition through an instance must not bind it as a method.
+    silu_and_mul = SiluAndMulWithKernel()
+    assert SiluAndMulWithKernel.kernel_condition(silu_and_mul)
+    assert silu_and_mul.kernel_condition(silu_and_mul)
```
sayakpaul, line 502: There should also be a test to exercise the path:

```py
if cond and not cond(module):
   ...
```

and check if the logging info was as expected.

sayakpaul (commented):
Just single comment regarding testing. Otherwise, looks good.

drbh (dismissed):
lgtm to me too - should be good to merge after the comments above

kernels/tests/test_layer.py
danieldk: This case is tested above in `test_kernel_condition_skips_kernelization`. However, it didn't assert the log, so added a log assertion.

sayakpaul (approved):
Really cool. Particularly the tests are 🤌

PR #798, opened 2026-08-31
feat: implement capability check in main

### LINKED ISSUE #795: Port arch/cc compatibility check from #774 to main
https://github.com/huggingface/kernels/pull/774 (`feat: arch/cc compatibility check.`) was merged into the `release-0.16` branch. We should now port that logic to `main`.

The PR touched:
- `kernels/src/kernels/archs.py` (new, arch/compute-capability matching logic)
- `kernels/src/kernels/utils.py` (use the arch check when selecting a kernel build)
- `kernels/tests/test_archs.py` (new tests)
- `docs/source/basic-usage.md` (docs)

Merge commit: `fe2f16b34343e51e88a244e7b854688528b88697`.

Related: closes-by-proxy #769 and #707 on the release branch — the same fixes should land on `main`.

---
Requested by Daniël de Kok - [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1787908428681879?thread_ts=1787908428.681879&cid=C090JN2P8NB) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-08-28T09-13-59-629Z_16d51e53-273f-47c0-bef0-c1ad8cb455c5.jsonl)

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/795

### DIFF
--- a/kernels/src/kernels/deps.py
+++ b/kernels/src/kernels/deps.py
@@ -70,15 +70,26 @@ class DepTreeNode(Generic[T]):
 
         return _import_from_path(self.location.variant_path, repo_info=repo_info, deps=deps)
 
+    def validate(
+        self: "DepTreeNode[LocalKernel | RemoteKernel]",
+        validator: "Validator",
+    ) -> None:
+        """Validate this kernel and its dependencies with the given validator."""
+
+        validator.validate(tree=self)
+
+        for node in self.deps.values():
+            node.validate(validator)
+
 
 class Validator(Protocol):
-    """Validator for a resolved kernel dependency tree."""
+    """Validator for a node in a resolved kernel dependency tree."""
 
     def validate(self, *, tree: DepTreeNode[LocalKernel | RemoteKernel]) -> None: ...
 
 
 class DependencyValidator:
-    """Validate the Python dependencies of every kernel in a tree."""
+    """Validate the Python dependencies of a kernel tree node."""
 
     def validate(self, *, tree: DepTreeNode[LocalKernel | RemoteKernel]) -> None:
         validate_dependencies(
@@ -87,18 +98,12 @@ class DependencyValidator:
             _backend(),
         )
 
-        for node in tree.deps.values():
-            self.validate(tree=node)
-
 
 class ArchValidator:
-    """Validate architecture compatibility for every kernel in a tree."""
+    """Validate architecture compatibility for a kernel tree node."""
 
     def validate(self, *, tree: DepTreeNode[LocalKernel | RemoteKernel]) -> None:
         _check_arch_incompatibility(tree.location.metadata, tree.location.variant_str)
-
-        for node in tree.deps.values():
-            self.validate(tree=node)
 
 
 @dataclass
--- a/kernels/src/kernels/deps.py
+++ b/kernels/src/kernels/deps.py
@@ -1,7 +1,7 @@
 from contextvars import ContextVar
 from dataclasses import dataclass
 from types import ModuleType
-from typing import Generic, TypeVar
+from typing import Generic, Protocol, TypeVar
 
 from huggingface_hub.hf_api import HfApi
 from kernels_data import KernelDependency, KernelVersion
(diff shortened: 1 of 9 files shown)


kernels/src/kernels/load.py
```diff
@@ -55,6 +55,7 @@ def get_kernel_with_resolver(
     backend: str | None,
     kernel: KernelDependency,
     resolver: Resolver | None,
+    check_arch: bool = False,
```
sayakpaul: I decided to keep the default to False because there are a couple of call sites which we are not testing here in this PR. But I would still prefer to keep to `True` as the default.

kernels/src/kernels/resolver.py
```diff
@@ -36,6 +36,10 @@ class LocalKernel:
     metadata: Metadata
     origin: "RemoteKernel | None" = None
 
+    @property
+    def variant_str(self) -> str:
```
sayakpaul, line 40: These properties are for convenience for better reporting of errors.

kernels/src/kernels/load.py
danieldk: I think we should avoid the same piling of argument that we had in the previous version of loading before the dependencies refactor. I think it's fine to have a `check_arch` argument on 'legacy' functions like `get_kernel`, etc. but it would be nice if we could keep `get_kernel_with_resolver` clean. So I propose to add a protocol `Validator` along the lines of:

```
class Validator(Protocol):
  def validate(self, *, tree: TreeNode[LocalKernel | RemoteKernel]):
    ...
```

No return type because a validator probably raises, then

```
class DependencyValidator:
  def validate(self, *, tree: TreeNode[LocalKernel | RemoteKernel]):
    validate_dependencies(...)

    for node in tree.deps.values():
      self.validate(node)
```

```
class ArchValidator:
  def validate(self, *, tree: TreeNode[LocalKernel | RemoteKernel]):
    _check_arch_incompatibility(tree.location.metadata, tree.location.variant_str)

    for node in tree.deps.values():
      self.validate(node)
```

```
@dataclass
class AllValidator:
    validators: list[Validator]

  def validate(self, *, tree: TreeNode[LocalKernel | RemoteKernel]):
    for validator in validators:
      validator(tree)
```

Then e.g. get_kernel could do something like:

```
return get_kernel_with_resolver(
  ...,
  validators=AllValidator(validator=[DependencyValidator(), ArchValidator] if check_arch else [DependencyValidator()])
)
```

Also dodges the need to have a default behavior (`True` or `False`) for `get_kernel_with_resolver`.

sayakpaul: I will debate that it has more layers of abstraction for what should be more explicit but I will table it for later and not block the PR.

danieldk: Small refinement after letting it sink in a bit. I think the recursion should be tree controlled, so we can have a method on `DepTreeNode`:

```
def validate(self, validator: Validator):
    for node in tree.deps.values():
      validator.validate(node)
```

And then the validator implementations can be simplified like this:

```
class ArchValidator:
  def validate(self, *, tree: TreeNode[LocalKernel | RemoteKernel]):
    _check_arch_incompatibility(tree.location.metadata, tree.location.variant_str)
```

danieldk: > I will debate that it has more layers of abstraction for what should be more explicit but I will table it for later and not block the PR.

I think we end up collecting a lot of incidental arguments for functions and then they are hard to deprecate when we don't need them anymore or when we have to change the type of an argument. With the Resolver/Validator approach we don't have these issues, since it's composable and you can just not pass something anymore if it is not needed in the future.

People who like monoliths can still use `get_kernel` and friends.

danieldk: To add a bit to that, specifically to the present extension. `check_arch` is not an argument that we actually want to have. We should _always_ check if the kernel is compatible with the compute hardware. We need `check_arch` as an argument because we need an escape hatch because the capability computation and representation is currently lacking here in some ways. If we solve the real issue, `check_arch` would become superfluous because it could always be `True`.

kernels/src/kernels/deps.py
```diff
@@ -69,17 +70,46 @@ def load(self) -> ModuleType:
...
+            self.validate(tree=node)
+
+
+class ArchValidator:
+    """Validate architecture compatibility for every kernel in a tree."""
+
+    def validate(self, *, tree: DepTreeNode[LocalKernel | RemoteKernel]) -> None:
+        _check_arch_incompatibility(tree.location.metadata, tree.location.variant_str)
+
+        for node in tree.deps.values():
```
danieldk: Sorry, added a comment later (see the thread) that we probably should de recursion itself in `DepTreeNode`, since it's cleaner and validators don't have to worry how to traverse the tree. Only realized some time later, apologies!

kernels/src/kernels/load.py
sayakpaul: Pushed the updates to change to the validator pattern. 

Making the entrypoint to `kernels` over anything beyond `get_kernel()` is also not healthy, IMO. It doesn't always have to mean having to prefer monoliths; sometimes it's just about preference and going with the simplest alternative available.

kernels/src/kernels/deps.py
sayakpaul: Yeah agreed. This way we can do `tree.validate(ArchValidator())` and not have to do `ArchValidator().validate(...)`.

danieldk (approved):
Super nice!

PR #801, opened 2026-08-31
feat: support and check minver

drbh:
this pr adds `minver` to the metadata so the client library can check if the current version meets the kernel metadata requirements


docs/source/builder/writing-kernels.md
```diff
@@ -208,6 +208,10 @@ The following sections enumerate all supported options for `build.toml`.
 - `source`: Git-compatible URL (passable to `git clone`) of the kernel-builder
   formatted source repository. This repository must contain a `build.toml` and
   `flake.nix` so that it can be pulled and built with the kernel builder.
+- `minver`: the minimum version of the `kernels` Python library required
```
danieldk: I don't think we want this in`build.toml` at all? The kernel builder should determine the minimum version, not the user. They will with a very high probability set the wrong version, e.g.:

* Not realizing that a kernel with dependencies requires 0.17, setting it to something old (or not at all).
* Setting it to _current version_ because that's the only thing they tested on, 'just to be safe', not making the kernel work older kernels versions.

I think the best way to tackle this is to add a method to `Config`:

```
fn required_kernels_version(&self, backend: Backend) -> Version<3> {
  // ...
}
```

That determines based on the features used inside `Config` what to set the version to. The default would be 0.14, but e.g. if there is a non-empty list of kernel dependencies for the given backend, set it to 0.17.

kernels-data/src/metadata.rs
```diff
@@ -93,6 +94,8 @@ pub struct Metadata {
     pub name: KernelName,
     pub id: String,
     pub version: usize,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub minver: Option<Version<3>>,
```
danieldk: I think we should be explicit and use `kernels_minver`.

kernels-data/bindings/python/src/lib.rs
```diff
@@ -393,6 +393,7 @@ struct PyMetadata {
     id: String,
     name: PyKernelName,
     version: usize,
+    minver: Option<String>,
```
danieldk:
```suggestion
    minver: Option<PyVersion>,
```

We still need a conversion from `Version<N>`, but I intentionally made `PyVersion` store a `Box<[usize]>` so that `Version<N>` can be boxed and used inside `PyVersion`.

kernels/src/kernels/importer.py
```diff
@@ -91,6 +94,27 @@ def _warn_if_dirty(metadata: Metadata, variant_str: str) -> None:
...
+    version required by a kernel.
+    """
+    minver = metadata.minver
+    if minver is None:
+        return
+
+    # Avoid an import cycle.
+    from kernels import __version__
+
+    if parse(__version__) < parse(minver):
```
danieldk: For `minver`, we shouldn't rely on `parse` but the fact that we already have a typed version in Rust-land (see comments above).

For the `__version__`, we might also have to strip `dev` tags and such, otherwise we'll get not-matching versions on our development branch:

```
❯ python3 -c 'from packaging.version import parse; print(parse("0.17.0.dev0") < parse("0.17.0"))'                                                                                                                                  
True
```

This would be really annoying, since I don't think we want to write dev versions to metadata.

docs/source/builder/writing-kernels.md
drbh: makes sense, removed it from `build.toml`. the builder now derives it with `Build::required_kernels_version(backend)`, defaulting to `0.14.0` and bumping to `0.17.0` when the backend has kernel dependencies. since `[general]` denies unknown fields, setting `minver` in `build.toml` is now a parse error rather than a silently ignored key

kernels-data/src/metadata.rs
drbh: renamed to `kernels_minver` (`kernels-minver` in `metadata.json`). kept it optional so metadata written before the key existed still parses

kernels-data/bindings/python/src/lib.rs
drbh: updated to `Option<PyVersion>` with a `From<Version<N>>` impl. the trailing zero stripping moved into a shared helper so a converted `Version<3>` and a parsed version compare equal

kernels/src/kernels/importer.py
drbh: the check now compares two `kernels_data.Version`

`parse` is only used on our own `__version__` to take the release segment, so `0.17.0.dev0` satisfies a `0.17.0` requirement (the warning still prints the verbatim `__version__`)

note: if `__version__` is not pep 440 (e.g. a vcs derived version) the check is skipped, since a warning should not be able to break the import

kernels-data/bindings/python/src/version.rs
```diff
@@ -11,6 +12,24 @@ pub(crate) struct PyVersion {
     inner: Box<[usize]>,
 }
 
+impl PyVersion {
+    /// Construct a version from its components, stripping trailing zeros so
```
danieldk: If what we are wrapping is a `Version<3>` and the versions are identical, there will always be the same number of trailing zeros. `Version<N>` is there to strongly type versions.

kernels-data/bindings/python/src/version.rs
```diff
@@ -11,6 +12,24 @@ pub(crate) struct PyVersion {
...
+        Self {
+            inner: parts.into_boxed_slice(),
+        }
+    }
+}
+
+impl<const N: usize> From<Version<N>> for PyVersion {
+    fn from(version: Version<N>) -> Self {
+        PyVersion::normalized(version.to_vec())
+    }
```
danieldk, line 23: Directly put the `Version<N>` in the box. We keep all the guarantees of it.

kernels-data/bindings/python/src/version.rs
```diff
@@ -32,12 +51,7 @@ impl PyVersion {
             })?;
             parts.push(component);
         }
-        // Remove trailing zeros for normalization.
-        let trailing_zeros = parts.iter().rev().take_while(|&&x| x == 0).count();
-        parts.truncate(parts.len() - trailing_zeros);
-        Ok(Self {
-            inner: parts.into_boxed_slice(),
-        })
+        Ok(PyVersion::normalized(parts))
```
danieldk: So, we can't have the same guarantees in Python-land since we get a string. Maybe we can require another argument for the expected length?

Or put the trailing zero logic in Eq/Ord.

kernels-data/bindings/python/src/version.rs
drbh: removed the trailing zero stripping and the `normalized` helper, versions are stored as is now

kernels-data/bindings/python/src/version.rs
drbh: done, `From<Version<N>>` now boxes the components directly through the `[usize]` deref, no copy into a `Vec` first

kernels-data/bindings/python/src/version.rs
drbh: went with the Eq/Ord option to keep `from_str` a single argument. comparisons are component-wise with missing components counting as zero, so `0.17` and `0.17.0` compare as equal (`Hash` skips trailing zeros to stay consistent with `Eq`). storage is untouched, `str()` round trips what was parsed

note: this also covers the pep 440 release segment in `_installed_version`, `parse("0.17").release` is `(0, 17)` and would otherwise compare below a `0.17.0` minver

danieldk (approved):
Nice! 🚀

PR #802, opened 2026-08-31
fix: point get-kernel-check at the built variant

drbh:
this pr fixes the macos ci failures where `get-kernel-check` rejects a kernel that built fine. first hit on the einops port in https://github.com/huggingface/kernels-community/pull/1092

tldr; the check handed `$out` to the resolver and let it pick a variant by detecting the backend at runtime. metal is the only backend detected from live hardware (`torch.backends.mps.is_available()`, everything else is baked into the torch build). the sandboxed macos build cannot see the gpu, so the `torch-metal` variant is never selected and the check fails even though the artifact is fine

repro (no kernel build needed)

```bash
mkdir -p /tmp/fake-kernel/torch-metal && touch /tmp/fake-kernel/torch-metal/metadata.json
kernels/.venv/bin/python -c "from pathlib import Path; import kernels; kernels.get_local_kernel(Path('/tmp/fake-kernel'))"
```

```
ValueError: Cannot parse metadata from `"/tmp/fake-kernel/torch-metal/metadata.json"`: EOF while parsing a value at line 1 column 0
```

on a mac with a visible gpu the variant is selected fine (it gets as far as reading the empty `metadata.json`). same command inside a seatbelt sandbox

```bash
sandbox-exec -p '(version 1)(allow default)(deny iokit-open)' \
    kernels/.venv/bin/python -c "from pathlib import Path; import kernels; kernels.get_local_kernel(Path('/tmp/fake-kernel'))"
```

```
FileNotFoundError: Cannot find a build variant for this system in /tmp/fake-kernel:
```

`sandbox-exec` is the same seatbelt mechanism the nix darwin sandbox uses. denying iokit hides the gpu, `mps.is_available()` returns false, and the same directory no longer resolves

note* this is also why only some metal builds fail in ci. arch metal builds set `__noChroot = metalSupport` since they need the host metal toolchain, so the gpu is still visible during the check. noarch metal builds like einops run fully sandboxed and always fail. the first command also fails on linux since there is no mps at all

the check now globs `$out` for `*/metadata.json` and points the resolver at that exact variant dir instead of auto detecting from the build host. each extension derivation produces exactly one variant, so zero or multiple variants still fail the check

### DIFF
--- a/nix-builder/pkgs/get-kernel-check/get-kernel-check-hook.py
+++ b/nix-builder/pkgs/get-kernel-check/get-kernel-check-hook.py
@@ -19,8 +19,14 @@ with open(kernelDeps) as f:
     kernel_paths = KernelPaths.from_json(f.read())
 
 kernel = KernelDependency(repo_id=out, version=KernelVersion.Version(0))
+
+# The build host may not expose the accelerator being targeted (e.g. Metal is
+# undetectable inside the sandboxed macOS build), so point the resolver at the
+# derivation's single variant instead of relying on backend auto-detection.
+[variant] = (p.parent for p in Path(out).glob("*/metadata.json"))
+
 resolvers = [
-    RepoPathsResolver(local_kernels={out: Path(out)}),
+    RepoPathsResolver(local_kernels={out: variant}),
     KernelPathsResolver(kernel_paths=kernel_paths),
 ]
 
--- a/nix-builder/pkgs/get-kernel-check/get-kernel-check-hook.py
+++ b/nix-builder/pkgs/get-kernel-check/get-kernel-check-hook.py
@@ -18,15 +18,25 @@ if not kernelDeps:
 with open(kernelDeps) as f:
     kernel_paths = KernelPaths.from_json(f.read())
 
+variant = os.getenv("variant")
+if not variant:
+    raise ValueError("`variant` environment variable is not set by Nix derivation")
+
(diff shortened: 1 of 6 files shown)


nix-builder/pkgs/get-kernel-check/get-kernel-check-hook.py
```diff
@@ -19,8 +19,14 @@
     kernel_paths = KernelPaths.from_json(f.read())
 
 kernel = KernelDependency(repo_id=out, version=KernelVersion.Version(0))
+
+# The build host may not expose the accelerator being targeted (e.g. Metal is
+# undetectable inside the sandboxed macOS build), so point the resolver at the
+# derivation's single variant instead of relying on backend auto-detection.
+[variant] = (p.parent for p in Path(out).glob("*/metadata.json"))
```
danieldk: Hmmm, I don't think we should rely on a wildcard. It weakens the test.

At any rate, what I think we should do:

- Add a `variant` argument to the `arch.nix`s and `no-arch.nix`.
- Expose `variant` through the `env`attribute.
- Validate in this hook that `variant` is set and then use it.
- Where the build derivations (arch/no-arch) is called, pass something like `variant = buildSet.variants.kernelVariant kernelConfig;`

Then we are not only testing the correct variant, but would also nicely get CI errors if the build is generating another variant than what is expected (which a wildcard would happily pass).

danieldk (approved):
Awesome!

PR #803, opened 2026-08-31
fix: cleanup failed imports

drbh:
this pr cleans up partially initialized kernel modules after an import fails and adds kernel context to the original exception

previously, a failed `exec_module` call left the module in `sys.modules`, where a later import could return the partially initialized module. `_import_from_path` now removes the entry before re-raising the original exception with the kernel name, variant, and remote revision when available

note: the additional context uses `BaseException.add_note` from PEP 678 on Python 3.11 and newer. Python 3.10 re-raises the original exception without a note

https://peps.python.org/pep-0678/

a regression test covers the failed import cleanup and exception note

### DIFF
--- a/kernels/src/kernels/importer.py
+++ b/kernels/src/kernels/importer.py
@@ -120,8 +120,17 @@ def _import_from_path(
     # Avoid an import cycle.
     from kernels.deps import use_kernel_deps
 
-    with use_kernel_deps(deps):
-        spec.loader.exec_module(module)  # type: ignore
+    try:
+        with use_kernel_deps(deps):
+            spec.loader.exec_module(module)  # type: ignore
+    except Exception as e:
+        # Remove the partially initialized module, so that a retry
+        # imports from scratch.
+        sys.modules.pop(metadata.id, None)
+        if hasattr(e, "add_note"):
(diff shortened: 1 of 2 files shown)


kernels/src/kernels/importer.py
```diff
@@ -120,8 +120,17 @@ def _import_from_path(
...
+    try:
+        with use_kernel_deps(deps):
+            spec.loader.exec_module(module)  # type: ignore
+    except Exception as e:
+        # Remove the partially initialized module, so that a retry
+        # imports from scratch.
+        sys.modules.pop(metadata.id, None)
+        if hasattr(e, "add_note"):
+            origin = f" (from {repo_info.repo_id}, revision: {repo_info.revision})" if repo_info else ""
+            e.add_note(f"while importing kernel '{metadata.name}' variant '{variant_path.name}'{origin}")
```
danieldk: Nit: but I would put the space in the string that gets the substring


```suggestion
            origin = f"({repo_info.repo_id}, revision: {repo_info.revision})" if repo_info else ""
            e.add_note(f"while importing kernel '{metadata.name}', variant '{variant_path.name}' {origin}")
```

danieldk (commented):
Great catch! Small nit.

danieldk (approved):

PR #805, opened 2026-09-02
feat: implement `user_agent` in layers

### LINKED ISSUE #782: Support passing `user_agent` for layer repos / `kernelize`
### Context

Downstream libraries (e.g. `transformers`, via [`integrations/hub_kernels.py`](https://github.com/huggingface/transformers/blob/f440843e1fb6d2b7171cf4cc44125fbdab30ef5b/src/transformers/integrations/hub_kernels.py#L775-L789)) would like to identify which calls to the Hub originate from them, so we can understand what is driving kernel usage.

`get_kernel()` / `install_kernel()` already accept a `user_agent` (str | dict) that is forwarded to `snapshot_download()` (see `kernels/src/kernels/load.py`, `install.py`, `hf_hub.py`). However, the layer-repo path (`LayerRepository` + `kernelize()`, `kernels/src/kernels/layer/`) has no way to pass one — and that's the path `transformers` mostly uses ([layer repo definitions](https://github.com/huggingface/transformers/blob/f440843e1fb6d2b7171cf4cc44125fbdab30ef5b/src/transformers/integrations/hub_kernels.py#L137)), so downloads triggered through `kernelize` are currently unattributable.

### Ask

- Allow wiring a `user_agent` through `LayerRepository` (and/or `kernelize()`) down to the Hub download calls, consistent with `get_kernel()`.
- Keep it strictly optional / opt-in — telemetry has caused pushback before. `HF_HUB_DISABLE_TELEMETRY` already covers the `huggingface_hub` side, so respecting that is enough on our end.

### Timing

Preferably land this *after* the kernel-to-kernel dependency work (kernels-as-dependencies, e.g. `einops`, CuTeDSL) merges, since how these arguments are passed around will change with it.

cc @danieldk

_From a Slack discussion between Sayak Paul, Anton and Daniël de Kok._

---
Requested by Sayak Paul - [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1786371411459309?thread_ts=1786371411.459309&cid=C090JN2P8NB) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-08-24T06-15-24-667Z_a6945d57-6bfd-4eb1-beb7-302296a3aad7.jsonl)

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/782


danieldk (dismissed):

nix-builder/lib/extension/torch/arch.nix
```diff
@@ -244,17 +244,16 @@ stdenv.mkDerivation (prevAttrs: {
...
+  env = {
+    inherit variant;
+  }
+  // lib.optionalAttrs cudaSupport {
+    CUDAToolkit_ROOT = "${lib.getDev cudaPackages.cuda_nvcc}";
+  }
+  // lib.optionalAttrs xpuSupport {
+    MKLROOT = oneapi-torch-dev;
+    SYCL_ROOT = oneapi-torch-dev;
+  };
```
sayakpaul, line 256: To please https://github.com/huggingface/kernels/actions/runs/33708021894/job/100501292549?pr=805

danieldk (approved):

PR #806, opened 2026-09-02
feat: allow passing a list of trusted repo ids.

### LINKED ISSUE #756: Allow `trust_remote_code` to be a per-repo allowlist (e.g. `["org1/kernel1", "org2/kernel2"]`)
### Context

We want to keep the set of `trustedKernelPublisher` orgs on the Hub small, so callers that need a kernel from a non-trusted namespace (e.g. `biohub/esmfold2-trimul`, recently moved out of `kernels-community`) currently have to pass a blanket `trust_remote_code=True`. That's a very crude hammer: it disables the trust check for *every* kernel loaded at that call site, not just the one repo the caller actually vetted.

### Proposal

Let `trust_remote_code` accept a list of repo ids acting as an explicit allowlist:

```python
get_kernel("biohub/esmfold2-trimul", trust_remote_code=["biohub/esmfold2-trimul"])
```

Suggested logic in `_check_trust_remote_code` (`kernels/src/kernels/hf_hub.py`):

```python
if trusted_publisher(repo_id):        # existing Hub query
    return True
elif isinstance(trust_remote_code, list) and repo_id in trust_remote_code:
    return True
elif isinstance(trust_remote_code, bool) and trust_remote_code:  # isinstance check avoids truthy lists
    return True
else:
    return False
```

This gives a useful middle ground between all-or-nothing and full code-signing: "I trust repos X, Y and Z, but nothing else". Downstream libraries (`transformers`, `diffusers`) can then opt into exactly the kernels they need instead of turning trust checks off globally.

### Notes / open questions

- The signature is already `trust_remote_code: bool | list[str] = False`, where `list[str]` was reserved for code-signing identities. Signing will use a different type anyway (`list[Policy]`), so `list[str]` is free for the allowlist meaning. The current "signing identities are not implemented" warning path would be replaced. Worth documenting the type/semantics change explicitly.
- Ideally small enough to cherry-pick onto the current stable branch so `transformers` can use it without waiting for a major release.
- Complementary (not conflicting) with the `diffusers` approach of loading remote kernels by default with a documented env flag to disable — that stays a global switch, this is per-repo granularity.

Discussed in Slack with @danieldk, @Rocketknight1 (Matt) and Pierric.

---
Requested by Sayak Paul - [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1786461294575059?thread_ts=1786461294.575059&cid=C090JN2P8NB) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-08-11T15-24-01-563Z_e36baade-16b6-4676-9c49-864d86eb5a3f.jsonl)

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/756


kernels/src/kernels/layer/func.py
```diff
@@ -94,7 +94,9 @@ def __init__(
 
         self._repo_id = repo_id
         self.func_name = func_name
-        self._trust_remote_code = trust_remote_code
+        self._trust_remote_code = (
+            trust_remote_code.copy() if isinstance(trust_remote_code, list) else trust_remote_code
+        )
```
sayakpaul, line 99: This complexity is for the following reasons:

* We use a check of `if isinstance(trust_remote_code, list): ...`. Otherwise, we could have packed into a `tuple` directly. We need a tuple to be able to `hash()`.
* We want to prevent user-side mutations to an allowlist affecting the status of `trust_remote_code`.

danieldk: I like the defensive coding here!

danieldk (approved):

PR #807, opened 2026-09-02
nix-builder: add support for ROCm 7.14

danieldk:
Torch 2.14 adds support for ROCm 7.14. However, this is not a simple update on our side, since upstream switched to a new build system + repository for ROCm packages (TheRock). The main issue is that all the components are packaged differently, not only do they use different package names, there is no 1:1 mapping at all.

To accommodate these changes, the `rocm-packages` is forked into `rocm-packages` and `rocm-packages-old`. `rocm-packages` is the new package set, whereas `rocm-packages-old` is used for ROCm 7.1/7.2. We will have to keep `rocm-packages-old` around as long as we have Torch + ROCm build variants that use these ROCm 7.1/7.2. Due to the splitting, the diff looks much larger than it really is. The main changes are:

* `rhel2nix.py` points to the new repository, supports having multiple ROCm versions in the same repository, and defines some new merges (because there is a cyclic dependency in the new ROCm).
* A `patchelf`/`autoPatchelfHook` override is added to rocmPackages. This solves an issue where the older `patchelf` in nixpkgs errors out and then writes broken dynamic libraries. We currently pin the same version as TheRock. See a full description of the issue in: https://github.com/ROCm/TheRock/issues/6712
* The generic builder a workaround because the new ROCm package set has setgid set on some directories. Since setgid cannot be set in the build sandbox,`cpio` (used by `rpmextract`) successfully extracts the archive, but emits warning and returns a non-zero exit code. To avoid this issue, we run `cpio` explicitly and permit the non-zero exit code as long as the only emitted error is the failure to set the setgid bit.
* All composed packages in the overlay are updated to use the new package names and paths.
* Torch and the arch kernel derivation are updated to work with the new packages.

Note that to keep things simple, both the old and the new package sets create `rocmPackages_<major>_<minor>` attrsets in nixpkgs. To be able to distinguish the package set in other derivations, we add a top-level `theRock` attribute to the package set, which is only `true` for the new package set. This attribute can be removed once we remove the old package set.

Due to this splitting, the diff looks much larger than it really is, because both package sets


nix-builder/pkgs/rocm-packages/rhel2nix.py
```diff
@@ -21,17 +23,43 @@
 
 REPOMD_NAMESPACES = {"repo": "http://linux.duke.edu/metadata/repo"}
 
+# Suffix marking per-GPU-target subpackages, e.g. amdrocm-blas7.14-gfx942.
+GFX_SUFFIX = re.compile(r"-gfx\d+[a-z0-9]*$")
```
sayakpaul, line 27: I think it's sufficiently reliable but I always have the fear of fragility with regexes.

nix-builder/overlay.nix
```diff
@@ -288,6 +288,10 @@ final: prev:
   inherit (final) callPackage;
   inherit (prev) lib;
 })
+// (import ./pkgs/rocm-packages-old {
+  inherit (final) callPackage;
+  inherit (prev) lib;
+})
```
sayakpaul, line 294: Neato

sayakpaul (approved):

PR #808, opened 2026-09-02
kernels: make kernels version check a validator

danieldk:
This check the kernel version already when just the metadata is loaded (fail early). Also make it raise an exception and not a warning. If the kernel uses a feature that is not supported by the kernels version, it is not going to work, so we should just raise. Avoids downloading the whole kernel and then having it fail with a hard to understand exception.

### DIFF
--- a/kernels/src/kernels/deps.py
+++ b/kernels/src/kernels/deps.py
@@ -4,7 +4,8 @@ from types import ModuleType
 from typing import Generic, Protocol, TypeVar
 
 from huggingface_hub.hf_api import HfApi
-from kernels_data import KernelDependency, KernelVersion
+from kernels_data import KernelDependency, KernelVersion, Version
+from packaging.version import InvalidVersion, parse
 
 from kernels.archs import _check_arch_incompatibility
 from kernels.backends import _backend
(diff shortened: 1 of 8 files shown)


drbh (approved):
lgtm!

kernels/src/kernels/deps.py
```diff
@@ -106,6 +107,57 @@ def validate(self, *, tree: DepTreeNode[LocalKernel | RemoteKernel]) -> None:
         _check_arch_incompatibility(tree.location.metadata, tree.location.variant_str)
 
 
+def _installed_version() -> Version | None:
+    """The installed `kernels` version as a numeric version.
+
+    Pre-release and development suffixes are stripped, since kernel metadata
+    only records release versions. Without stripping, a development version
+    like `0.17.0.dev0` would compare as older than the `0.17.0` it implements.
```
sayakpaul, line 115: But there might still be a possibility where a kernel is implemented with an unreleased version, right? I guess the latter part of the code handles that but I just wanted to flag.

sayakpaul (commented):
Really nice!

PR #809, opened 2026-09-02
Small validator refactorings

danieldk:
- Move validators to the `validation` module.
- Move validator tests to `test_validation`.
- Rename `Validator` to `MetadataValidator`, since we may want to support `KernelValidator`s in the future as well.
-  Let validators only take `Metadata` and the variant string. This makes it possible to change the dependency tree representation without breaking validators.


danieldk:
The nix fmt error is unrelated.

drbh (approved):
lgtm!

drbh:
> The nix fmt error is unrelated.

**nix formatting issue should be resolved by - https://github.com/huggingface/kernels/pull/810

kernels/src/kernels/load.py
```diff
@@ -61,7 +61,7 @@ def get_kernel_with_resolver(
     backend: str | None,
     kernel: KernelDependency,
     resolver: Resolver | None,
-    validator: Validator,
+    metadata_validator: MetadataValidator,
```
sayakpaul, line 64: I think this should be a generic validator still because we might want to add a kernel validator in the future and we will have to add another argument for that, right?

sayakpaul (commented):
My main comment is if we should have separate validators like this. Right now the argument makes it clear that the validator is only for metadata validation. So, are we aiming for a design where we have arguments for different validators that differ in their areas? For example, for a kernel validator we would add a different argument?

kernels/src/kernels/load.py
danieldk: For kernel validators there would be a `kernel_validator: KernelValidator` argument. Otherwise we'd have `MetadataValidator | KernelValidator`, but you would not be able to pass both.

danieldk: Or, we'd need another `Validator` protocol, but it couldn't define any methods, since the arguments would be different between metadata and kernel validators.

sayakpaul (approved):
Fine to merge I think the nix related failures are unrelated. But I am still debating if it makes sense to have separate arguments for different validators or if they should clubbed. I don't have a clear picture of the pros and cons of both yet.

PR #817, opened 2026-09-04
feat: suggest compatible variants when available in newer versions

### LINKED ISSUE #744: Suggest newer kernel version in error when no compatible variant exists for current version
## Problem

When `get_kernel` (or `load_kernel`) can't find a build variant compatible with the current system, the error message only reports that no matching variant exists for the resolved revision:

```
Cannot find a build variant for this system in {repo_id} (revision: {revision})
```

(raised in `kernels/src/kernels/utils.py`)

This is misleading when the user is on an older version branch of the kernel. For example, a user pinned to (or resolving to) an older version may conclude the kernel doesn't support their setup — e.g. "this kernel has no Torch 2.13 build" — when in fact a **newer version branch does have a compatible variant**. There is no hint that upgrading the kernel version would fix the problem.

## Proposal

When variant resolution fails (no matching variant for the current environment):

1. Enumerate the available version branches for the kernel repo (via `_get_available_versions` in `kernels/src/kernels/_versions.py`, which lists the `v<N>` refs — newer versions first).
2. Check whether any newer version has a variant compatible with the current system (`resolve_variant(s)` against that revision).
3. If a newer compatible version exists, extend the error message to suggest it, e.g.:

```
Cannot find a build variant for this system in {repo_id} (revision: v3):

<existing variants trace>

However, version v5 of '{repo_id}' has a build compatible with your system (torch 2.13, cuda 12.8, ...). Consider upgrading to that version.
```

## Considerations

- The check requires resolving variants of newer versions, which may involve extra Hub API calls — acceptable on the error path, but should be best-effort (don't mask the original error if the check itself fails, e.g. offline mode).
- In offline mode (`HF_HUB_OFFLINE`), the cached-version path (`_get_available_versions_from_cache`) could be used, or the check skipped gracefully.

---
Requested by Daniël de Kok - [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1785744399961329?thread_ts=1785744399.961329&cid=C090JN2P8NB) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-08-03T08-07-37-432Z_7bef89d7-1db3-426d-8371-54b66ba493f4.jsonl)

sayakpaul:
Fixes https://github.com/huggingface/kernels/issues/744

### DIFF
--- a/kernels/src/kernels/resolver.py
+++ b/kernels/src/kernels/resolver.py
@@ -4,7 +4,7 @@ from typing import Protocol, runtime_checkable
 
 from huggingface_hub.errors import LocalEntryNotFoundError
 from huggingface_hub.hf_api import HfApi
-from kernels_data import KernelDependency, KernelLocks, KernelPaths, KernelVersion, Metadata
+from kernels_data import KernelDependency, KernelLocks, KernelPaths, Metadata
 
 from kernels._versions import _get_available_versions, resolve_kernel_version
 from kernels.hf_hub import CACHE_DIR, _check_trust_remote_code
@@ -115,7 +115,6 @@ def resolve_hub_kernel(
     api: HfApi,
     backend: str | None,
     revision: str,
-    version: int | None = None,
 ) -> RemoteKernel:
     variants = get_variants(
         api,
@@ -124,11 +123,9 @@ def resolve_hub_kernel(
     )
     variant, trace = resolve_variant(variants, backend)
     if variant is None:
-        suggestion = _newer_compatible_version_suggestion(
+        suggestion = _latest_compatible_version_suggestion(
             api=api,
             repo_id=repo_id,
-            revision=revision,
-            version=version,
             backend=backend,
         )
         raise FileNotFoundError(
@@ -153,12 +150,10 @@ def resolve_hub_kernel(
     return location
 
 
-def _newer_compatible_version_suggestion(
+def _latest_compatible_version_suggestion(
     *,
     api: HfApi,
     repo_id: str,
-    revision: str,
-    version: int | None,
     backend: str | None,
 ) -> str:
     """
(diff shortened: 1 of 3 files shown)


kernels/src/kernels/resolver.py
```diff
@@ -123,8 +124,16 @@ def resolve_hub_kernel(
     )
     variant, trace = resolve_variant(variants, backend)
     if variant is None:
+        suggestion = _newer_compatible_version_suggestion(
```
sayakpaul: Could also call it `_maybe_suggest_newer_compatible_version`

kernels/tests/test_resolver.py
```diff
@@ -4,6 +4,7 @@
 from pathlib import Path
 
 import pytest
+from huggingface_hub.hf_api import GitRefInfo
```
sayakpaul, line 7: I think I have covered most bases here but I have a feeling I am missing 1/2 tests.

kernels/src/kernels/resolver.py
```diff
@@ -144,6 +153,57 @@ def resolve_hub_kernel(
...
+    """
+    try:
+        versions = _get_available_versions(repo_id, local_files_only=False)
+
+        if version is None:
+            version = next(
+                (
+                    candidate
+                    for candidate, ref in versions.items()
+                    if revision in (ref.name, ref.ref, ref.target_commit)
```
danieldk: I don't think this works in the general case. The problem is that the locked revision can come from an earlier commit on a branch.

kernels/src/kernels/resolver.py
```diff
@@ -144,6 +153,57 @@ def resolve_hub_kernel(
...
+                ),
+                None,
+            )
+
+        if version is None:
+            return ""
```
danieldk: I think we can still suggest a version if we don't know what version the current revision comes from.

kernels/src/kernels/resolver.py
```diff
@@ -115,6 +115,7 @@ def resolve_hub_kernel(
     api: HfApi,
     backend: str | None,
     revision: str,
+    version: int | None = None,
```
danieldk: I think just calling this `version` might suggest that it is actually used in resolving for `RemoteKernel`.

kernels/src/kernels/resolver.py
```diff
@@ -144,6 +153,57 @@ def resolve_hub_kernel(
...
+                    for candidate, ref in versions.items()
+                    if revision in (ref.name, ref.ref, ref.target_commit)
+                ),
+                None,
+            )
+
+        if version is None:
+            return ""
+
+        newer_versions = [candidate for candidate in versions if candidate > version]
```
danieldk: This could become quite costly, making the process appear blocked. I wonder if for now we should just check the latest version, since we want people to always be on the latest version and I think in 99% of the cases this is someone installing a new Torch and the kernel then not loading because they are on an old version.

Benefit is that we could also remove the `version` argument from `resolve_hub_kernel`.

danieldk (approved):
Really great UX improvement!

PR #818, opened 2026-09-08
Move `kernels-data` Python binding into `kernels`

danieldk:
This simplifies installs from the development tree. It also avoids having to maintain a separate dependency on kernels-data. The binding bits will always be in sync with kernels.


sayakpaul (dismissed):
Nice. My only concern is that the `kernels` source is now getting a bit harder to navigate. 

Also, is there any benefit of having another directory with `rust`, called `kernels-data` and put the respective files there? This will have a benefit if we ever wanted to add any more such Rust-based components to the `kernels` source.

danieldk:
> Also, is there any benefit of having another directory with `rust`, called `kernels-data` and put the respective files there? This will have a benefit if we ever wanted to add any more such Rust-based components to the `kernels` source.

Since it's internal now, we can always rename things. The module is `_data` now, but we can always rename it `_rust` in the future to be more generic. I think there is a benefit to having all the Rust bits in a single module (smaller compile/link time). To that end, I think I want to rename `kernels-data` (the Rust part) next to make it more generic, since we might also get more non-data stuff there.

PR #822, opened 2026-09-10
kernels-common: add `HFKernelsClientBuilder`

danieldk:
The `hf-hub` crate removed all standard env vars handling. This is a port of the environment handling of `huggingface_hub` to restore support for standard variables. We also add recognition of `KERNELS_CACHE` to align with the Python `kernels` client.


sayakpaul (approved):
Thanks. But I am still not sure why these were removed from the HF Hub Rust lib.

danieldk:
> Thanks. But I am still not sure why these were removed from the HF Hub Rust lib.

Me neither, these defaults were really useful.

PR #827, opened 2026-09-14
kernels: use `Oid` throughout kernels

danieldk:
Start using `Oid` everywhere we need a full Git commit SHA.

This change also refactors the signature verification CLI a bit, moving verification results to the actual verification classes.


drbh (approved):
lgtm!

PR #828, opened 2026-09-15
kernels: verify the signature of a kernel while loading

danieldk:
The basics of kernel signature verification were already in `kernels`, but verification would only be performed when explicitly running the `kernels verify-signature` command.

This change enables kernel signature verification during kernel loading. However, as part of the step-wise roll-out, signature verification will now only warn if the metadata is not signed, not correctly signed, or the data's digest is incorrect.

Signature verification during kernel loading is implemented as follows. First, we add the `KernelValidator` protocol, this is the counterpart to `MetadataValidator` that can only run when the kernel is fully downloaded. `get_kernel_with_resolver` now requires an additional `kernel_validator` argument and runs the given validators on the kernel dependency tree after the kernels are downloaded ('installed').

We add `SignatureValidator` as the first validator to implement the `KernelValidator` protocol and enable it in the relevant `get_.*` functions. This validator calls the existing verification structure. Validation is only performed on remote kernels, since local kernels are used in development and are typically not signed.

There was one more wrinkle in that validating the hash of all kernel files on each load is fairly expensive. So upon the first successful verification, we write a receipt that attests that an earlier verification is done. The receipt is content-addressed by the repo id, the resolved Git commit SHA and build variant, so it does not attest for newly-downloaded revisions or build variants.

The receipts allow us to skip the most expensive parts of the verification. However, the user might use a different policy after the verification that produced the receipt, so we will always test the kernel's signing certificate against the policy. This ensures that the kernel signature will be rejected if the kernel violates the policy used during a kernel load.


kernels-common/src/signing/receipt.rs
```diff
@@ -1,6 +1,34 @@
+use std::fs;
+use std::io::{self, Write as _};
+use std::path::PathBuf;
+
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+use thiserror::Error;
 
 use crate::git::Oid;
+use crate::hf::{UnknownCacheDir, kernels_cache};
```
sayakpaul, line 10: I think we should document kernels cache a bit more, showing how users can configure it, etc. Okay if that is done in a separate PR.

kernels/src/kernels/deps.py
```diff
@@ -68,6 +68,12 @@ def load(self) -> ModuleType:
 
         return _import_from_path(self.location.variant_path, repo_info=repo_info, deps=deps)
 
+    def validate_kernel(self: "DepTreeNode[LocalKernel]", validator: KernelValidator):
+        validator.validate_kernel(kernel=self.location)
+
+        for node in self.deps.values():
+            node.validate_kernel(validator)
```
sayakpaul, line 75: Nice. I like the segregation.

kernels/src/kernels/load.py
```diff
@@ -90,6 +96,7 @@ def get_kernel_with_resolver(
     )
     tree.validate_metadata(metadata_validator)
     tree_only_local = tree.install(api=api)
+    tree_only_local.validate_kernel(kernel_validator)
```
sayakpaul, line 99: Slick!

kernels/src/kernels/validate.py
```diff
@@ -124,3 +139,83 @@ def validate_metadata(self, *, metadata: Metadata, variant: str) -> None:
...
+
+        kernel_str = f"Kernel '{kernel.metadata.name}' variant '{kernel.variant_str}'"
+
+        match result:
+            case VerificationResult.Success():
+                logger.debug(f"{kernel_str}: {result}")
+            case VerificationResult.Failure():
+                logger.warning(f"{kernel_str}: {result}", stacklevel=3)
+            case _ as unreachable:
+                assert_never(unreachable)
```
sayakpaul, line 191: Beautiful!

kernels/tests/test_versions.py
```diff
@@ -93,3 +96,71 @@ def fail(*args, **kwargs):
...
+# a ref when resolving a name. The tests below ensure that this behavior is
+# correct.
+
+
+def _refs_of(cache_dir: Path, repo_id: str = REPO_ID) -> list[str]:
+    refs = cache_dir / repo_folder_name(repo_id=repo_id, repo_type="kernel") / "refs"
+    return sorted(p.name for p in refs.iterdir()) if refs.is_dir() else []
+
+
+@pytest.mark.parametrize("version", [KernelVersion.Revision("v1"), KernelVersion.Version(1)])
```
sayakpaul, line 113: Nice test.

kernels/tests/test_versions.py
```diff
@@ -93,3 +96,71 @@ def fail(*args, **kwargs):
...
+    assert not (tmp_path.parent.parent.parent / "escaped").exists()
+    assert not (tmp_path / "escaped").exists()
+
+
+def test_a_downloaded_revision_resolves_offline(tmp_path, monkeypatch):
+    monkeypatch.setenv("KERNELS_CACHE", str(tmp_path))
+
+    install_kernel("kernels-community/relu", revision="v1", backend="cpu")
+
+    assert install_kernel("kernels-community/relu", revision="v1", backend="cpu", local_files_only=True)
```
sayakpaul, line 158: Should we assert against the path returned by `install_kernel()` rather than a bare assert?

kernels/tests/test_versions.py
```diff
@@ -93,3 +96,71 @@ def fail(*args, **kwargs):
...
+
+
+def test_a_downloaded_version_resolves_offline(tmp_path, monkeypatch):
+    monkeypatch.setenv("KERNELS_CACHE", str(tmp_path))
+
+    install_kernel("kernels-community/relu", version=1, backend="cpu")
+
+    assert install_kernel("kernels-community/relu", version=1, backend="cpu", local_files_only=True)
```
sayakpaul, line 166: Same as above.

sayakpaul (commented):
Most of my comments are just appreciation. 

Some high-level comments:

* How much overhead are we talking about? Depending on that, I wonder if users should have knob to control it (true, by default).
* I think we should document the validation parts a little bit even if we consider the feature to be experimental (doesn't need to be this PR). 
* Is there any merit to parallelize the expensive verification bit? Or the side-effects would be too crazy?

> There was one more wrinkle in that validating the hash of all kernel files on each load is fairly expensive. So upon the first successful verification, we write a receipt that attests that an earlier verification is done. The receipt is content-addressed by the repo id, the resolved Git commit SHA and build variant, so it does not attest for newly-downloaded revisions or build variants.

Should we write that receipt in a "kernels" cache? However, what happens if the version we're loading was updated in between two loads? Since Git commit SHA is part of the content address, the earlier receipt should already be invalidated in that case? Do we do the verification from scratch in that case or do we utilize part of the earlier receipt to determine the new stuff we need to verify and only focus on those?

danieldk:
Thanks for the review, great questions!

> * How much overhead are we talking about? Depending on that, I wonder if users should have knob to control it (true, by default).

- First verification: co a request to `rekor.sigstore.dev` to check if the signing certificate is valid, verify `metadata.json` using the signature, hash all kernel files to check against the digest in `metadata.json`. I think in this step hashing all kernel files is the most expensive, though on modern CPUs with SHA-NI it's pretty fast.
- Subsequent verifications (when there is a receipt): really cheap. Since we do not want to counter cache tampering as an attack vector, we do not have to reverify `metadata.json` or the kernel digest. So it only checks that there is a valid receipt and verifies that the signature is compatible with the given policies.

> * I think we should document the validation parts a little bit even if we consider the feature to be experimental (doesn't need to be this PR).

Definitely!

> * Is there any merit to parallelize the expensive verification bit? Or the side-effects would be too crazy?

Should be possible to do, especially if we switch to `sigstore-rs`, since we could fire off multiple verifications and during verification drop the GIL. But I think the lowest hanging fruit may be parallelizing downloads of kernels.

> Should we write that receipt in a "kernels" cache?

That's where it is written now indeed.

> However, what happens if the version we're loading was updated in between two loads? Since Git commit SHA is part of the content address, the earlier receipt should already be invalidated in that case?

It stays around, but it will not be used anymore (since the content hash doesn't match anymore). I am not sure at what point the receipts should be removed/invalidated. I think by default the older version stays in the cache as well, which may even be necessary because other libraries/apps/kernels may still depend on the older version of the kernel (e.g. because they lock a revision or use an older version). I am not sure when the receipts should/could be garbage collected.

> Do we do the verification from scratch in that case or do we utilize part of the earlier receipt to determine the new stuff we need to verify and only focus on those?

Yes, it is from scratch. We have to, since an attacker could e.g. have replaced a file by another file with the same size and modification time, so we need to check all the contents anyway.

kernels/tests/test_versions.py
danieldk: Great idea! Asserted it to be the same as the non-local files path. Also for the case below.

sayakpaul (approved):
🔥

PR #829, opened 2026-09-16
Fix `egg_info.writers` entry point module path

jiqing-feng:
## Problem

With `kernels` installed, any setuptools source build in the same environment fails, including builds of unrelated packages:

```
File "setuptools/command/egg_info.py", line 304, in run
    writer = ep.load()
ModuleNotFoundError: No module named 'kernels.lockfile'
```

Reproducer:

```bash
pip install kernels==0.17.0
printf 'from setuptools import setup\nsetup(name="smoke", version="0.1")\n' > setup.py
python setup.py egg_info
```

## Cause

`kernels/pyproject.toml` registers a setuptools entry point pointing at a module that is not shipped:

```toml
[project.entry-points."egg_info.writers"]
"kernels.lock" = "kernels.lockfile:write_egg_lockfile"
```

`write_egg_lockfile` lives in `kernels/locking.py`; there is no `kernels/lockfile.py` in the wheel. The entry point was not updated when `lockfile.py` was renamed to `locking.py`.

setuptools loads **all** registered `egg_info.writers` on every `egg_info` run, so a broken writer from any installed distribution takes down unrelated builds. PEP 517 isolated builds are unaffected; `--no-build-isolation` builds reuse the ambient setuptools and always hit it.

Present in 0.17.0 and on `main`; 0.16.2 is the last good release.

## Fix

Point the entry point at `kernels.locking`, and add a regression test that loads it the way setuptools does — nothing covered this path, which is why the rename regressed silently.

## Validation

Built a wheel from this branch and installed it:

- `ep.load()` resolves to the real function.
- The reproducer above succeeds.
- The new test passes, and fails when reverted to `kernels.lockfile`.


jiqing-feng:
Hi @danieldk. Would you please review this PR? Thanks!

danieldk (approved):
Thanks a lot for catching this! Will also backport to 0.17.

PR #831, opened 2026-09-16
nix-builder: update nixpkgs, bumps Python from 3.13 to 3.14

danieldk:


nix-builder/scripts/helpers/prefetch-tpu-wheel.sh
```diff
@@ -8,13 +8,13 @@
 #
 # Usage:
 #   GCLOUD_ACCESS_TOKEN=$(gcloud auth print-access-token) \
-#     scripts/helpers/prefetch-tpu-wheel.sh libtpu 0.0.43 [cp313]
+#     scripts/helpers/prefetch-tpu-wheel.sh libtpu 0.0.43 [cp314]
```
sayakpaul, line 11: Have we verified if this exists?

flake.nix
```diff
@@ -10,10 +10,7 @@
 
   inputs = {
     flake-utils.url = "github:numtide/flake-utils";
-    # Temporarily pin this commit to fix ucx issues:
-    # https://github.com/NixOS/nixpkgs/pull/538738
-    nixpkgs.url = "github:NixOS/nixpkgs/ec1a11210589d294f0ac99d3290a27e6c73dfa1d";
-    #nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable-small";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable-small";
```
sayakpaul, line 13: Can we use pinned commits?

sayakpaul (commented):
Left two minor comments.

flake.nix
danieldk: They are pinned in `flake.lock`.

sayakpaul (approved):

PR #832, opened 2026-09-17
nix-builder: experimental support for CUDA 13.4

danieldk:
Add support for CUDA 13.4 with a Torch 2.15 nightly. Since Torch is still a nightly build, this variant is not included in bundle builds.


drbh (approved):
nice! lgtm!

PR #837, opened 2026-09-21
[docs] add latest learnings from version bumps and test module

sayakpaul:


docs/source/builder/writing-kernels.md
```diff
@@ -545,6 +545,35 @@ $ nix run .#ciTests.torch210-cxx11-cpu-x86_64-linux
 When running the tests on a non-NixOS systems, make sure that
 [the CUDA driver library can be found](https://danieldk.eu/Software/Nix/Nix-CUDA-on-non-NixOS-systems#solutions).
 
+### Misc
```
danieldk: I would name it for what it documents:

```suggestion
### Tests that need access to internal APIs
```

docs/source/builder/writing-kernels.md
```diff
@@ -545,6 +545,35 @@ $ nix run .#ciTests.torch210-cxx11-cpu-x86_64-linux
...
+```py
+...
+
+importlib.import_module(f"{flash_attn4.__name__}.compute_block_sparsity")
+BlockSparsityKernel = flash_attn4.compute_block_sparsity.BlockSparsityKernel
+BlockSparseTensors = flash_attn4.block_sparsity.BlockSparseTensors
+```
+
+This pattern should be avoided. Instead, create a separate private
+module and include the modules needed in the tests inside it. Then
```
danieldk: I think it's best to mandate the name to ensure that this gets somewhat standardized:

```suggestion
This pattern should be avoided. Instead, create a separate private
module named `_private_for testing` and include the modules needed in the tests inside it. Then
```

docs/source/kernel-requirements.md
```diff
@@ -310,6 +310,10 @@ replaced by the new build:
   - Torch stable ABI kernels, as long as the CUDA versions that get built
     overlap with the current build variants.
 
+Note that even for noarch Python-only kernels, we need a version bump
+where arguments are removed from the public API or change the default
+for an argument present in the public API.
```
danieldk: I think there are other cases missing where a bump is also needed. For instance, when a new argument with a default is added that implicitly changes the behavior. Think of something like:

Old:

```
def foo(...) -> States:
```

New:

```
# No bump needed, same behavior when not using the new argument.
def foo(..., return_states: bool=True) -> States|None:

# Bump needed, behavior chanegs.
def foo(..., return_states: bool=False) -> States|None:
```

I think the most important is that people apply the rule of thumb, because there are always cases that we haven't foreseen. For as far as we list concrete cases, I think they should be formulated positively (what does not require a version bump) and not negatively (what requires a version bump), because it is hard to enumerate all negative exhaustively and we want people to err on the side of caution and bump the version.

So I propose the following...

I would change the sentence above:

> A version bump is not needed when every existing build variant is replaced by the new build:

to:

> For the following types of kernels, some API changes are do not require bumping the version:

And then change the new text to:

```suggestion
In the following cases no version bump is required for these types of kernels:

- Adding new functions/methods/classes.
- Adding new arguments with default values to a function/method/constructor, as long as the default values do not change previous behavior.
```

docs/source/kernel-requirements.md
danieldk: We still need to update this text (for suggested wording see my other comment).

sayakpaul: Done

danieldk (approved):

